PR #3, opened 2021-02-06
[binaryparagraph] Fix VS 2015 compile error.

Hoikas:
This fixes compilation on Visual Studio 2015. `constexpr` lambdas were added in Visual Studio 2017 update 15.3. Further, `std::all_of` did not become `constexpr` until C++20.

### DIFF
--- a/src/vcpkg/binaryparagraph.cpp
+++ b/src/vcpkg/binaryparagraph.cpp
@@ -138,7 +138,7 @@ namespace vcpkg
 
     void BinaryParagraph::canonicalize()
     {
-        constexpr auto all_empty = [](const std::vector<std::string>& range) {
+        const auto all_empty = [](const std::vector<std::string>& range) {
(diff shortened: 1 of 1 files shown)


strega-nil:
Are we still guaranteeing that we compile with VS2015? Wasn't part of the point of pulling out the tool that we didn't have to compile with VS2015 anymore?

Hoikas:
Good question. I looked but didn't see any explicit guidance on that in this repository. This change only exists so I could test #4.

strega-nil:
@Hoikas we've agreed that we are explicitly not supporting VS2015 from this point onwards. Thanks for the PR, and @ras0219 will look at #4, but closing as "no moar VS2015 muahahaha"

Hoikas:
I share your sentiment -- 💥 💥 💥  VS 2015 :smile:

PR #4, opened 2021-02-06
[visualstudio] Fix v120 toolset detection issue.

Hoikas:
Previously, the v120 toolset was only detected by finding Visual Studio 2013 with `vswhere.exe`. Unfortunately, `vswhere.exe` is only available if Visual Studio 2017 (or higher) is installed. So, if you are using a configuration that has only ever had Visual Studio 2015 and Visual Studio 2013 installed, then the community triplet `x86-windows-v120` would fail with the error `Could not find Visual Studio instance with v120 toolset`. This can be easily [observed](https://ci.appveyor.com/project/Hoikas/appveyorvcpkg2013/builds/37646642) by attempting to use vcpkg on the "Visual Studio 2015" AppVeyor image. Now that bootstrapping no longer requires Visual Studio >= 2015 available, this also affects machines that only have Visual Studio 2013 installed... eg the "Visual Studio 2013" AppVeyor image.

This pull request add version independent detection of Visual Studio 2013 to [fix](https://ci.appveyor.com/project/Hoikas/appveyorvcpkg2013/builds/37647047) this issue. Because the detection code contained a lot of duplication before I started, this includes a round of de-duplication. I'm unfamiliar with the (what seemed to be) five different variants of string types in use in vcpkg, so please let me know if I goofed with making extra copies or something silly like that :smile:.

### DIFF
--- a/src/vcpkg/visualstudio.cpp
+++ b/src/vcpkg/visualstudio.cpp
@@ -126,45 +126,38 @@ namespace vcpkg::VisualStudio
             }
         }
 
-        // VS2019 instance from environment variable
-        auto maybe_vs160_comntools = System::get_environment_variable("vs160comntools");
-        if (const auto path_as_string = maybe_vs160_comntools.get())
-        {
-            // We want lexically_normal(), but it is not available
-            // Correct root path might be 2 or 3 levels up, depending on if the path has trailing backslash.
-            auto common7_tools = fs::u8path(*path_as_string);
-            if (common7_tools.filename().empty())
-                instances.emplace_back(common7_tools.parent_path().parent_path().parent_path(),
-                                       "16.0",
-                                       VisualStudioInstance::ReleaseType::LEGACY);
-            else
-                instances.emplace_back(
-                    common7_tools.parent_path().parent_path(), "16.0", VisualStudioInstance::ReleaseType::LEGACY);
-        }
-
-        const auto append_if_has_cl_vs140 = [&](fs::path&& path_root) {
+        const auto append_if_has_cl = [&](fs::path&& path_root, CStringView version) {
             const auto cl_exe = path_root / "VC" / "bin" / "cl.exe";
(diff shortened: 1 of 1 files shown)


src/vcpkg/visualstudio.cpp
```diff
@@ -126,45 +126,39 @@ namespace vcpkg::VisualStudio
...
+        const auto maybe_append_legacy_vs = [&](ZStringView env_var, const fs::path& dir, CStringView version) {
+            // VS instance from environment variable
+            append_if_comntools_has_cl(env_var, version);
+            // VS instance from Program Files
+            append_if_has_cl(program_files_32_bit / dir, version);
+        };
 
-        // VS2015 instance from Program Files
-        append_if_has_cl_vs140(program_files_32_bit / "Microsoft Visual Studio 14.0");
+        append_if_comntools_has_cl("vs160comntools", "16.0");
```
ras0219, line 159: This is incorrect -- due to layout changes in VS 2017 (as part of the new installer), the compiler can no longer be located via a simple path navigation. This needs to be changed back to the previous approach (always add it after normalization).

A big difference between the different `VSXYZCOMNTOOLS` vars is that `vs160comntools` is actually not set system-wide as part of installation; it's only available as part of a developer prompt. This means its presence indicates "intent" on the developer's part to make this copy of VS available and it significantly reduces the burden of a false positive.

ras0219 (changes requested):
This looks good to me once the v160 changes are reverted. Thanks for the PR!

src/vcpkg/visualstudio.cpp
Hoikas: That explains why the code looked duplicated to me before - I missed that detail. I hope you don't mind that I tried to keep things generic. It could prove useful if support is ever desired for even older toolsets.

ras0219-msft (approved):

PR #5, opened 2021-02-08
[vcpkg] Support host dependencies

ras0219:
Moved from https://github.com/microsoft/vcpkg/pull/15424

@strega-nil


azure-pipelines/end-to-end-tests.ps1
```diff
@@ -49,9 +49,43 @@ if ($Filter -ne $Null) {
...
+$envvars = $envvars_clear + @("VCPKG_DOWNLOADS")
+
 $AllTests | % {
     Write-Host "[end-to-end-tests.ps1] [$n/$m] Running suite $_"
-    & $_
+
+    $envbackup = @{}
+    foreach ($var in $envvars)
+    {
+        $envbackup[$var] = [System.Environment]::GetEnvironmentVariable($var)
```
strega-nil, line 71:
```suggestion
        $envbackup[$var] = Get-Content Env:/$var
```
I like this better but it's up to you

azure-pipelines/end-to-end-tests.ps1
```diff
@@ -49,9 +49,43 @@ if ($Filter -ne $Null) {
...
+    }
+
+    try
+    {
+        foreach ($var in $envvars_clear)
+    {
```
strega-nil, line 77:
```suggestion
        {
```

azure-pipelines/end-to-end-tests.ps1
```diff
@@ -49,9 +49,43 @@ if ($Filter -ne $Null) {
...
+            [System.Environment]::SetEnvironmentVariable($var, $null)
```
strega-nil, line 78:
```suggestion
            Remove-Item Env:/$var
```

azure-pipelines/end-to-end-tests.ps1
```diff
@@ -49,9 +49,43 @@ if ($Filter -ne $Null) {
...
+        }
+        & $_
+    }
+    finally
+    {
+        foreach ($var in $envvars)
+        {
+            [System.Environment]::SetEnvironmentVariable($var, $envbackup[$var])
```
strega-nil, line 86:
```suggestion
            Set-Content Env:/$var $envbackup[$var])
```

src/vcpkg/vcpkgcmdarguments.cpp
```diff
@@ -925,6 +930,8 @@ namespace vcpkg
 
     constexpr StringLiteral VcpkgCmdArguments::TRIPLET_ENV;
     constexpr StringLiteral VcpkgCmdArguments::TRIPLET_ARG;
+    constexpr StringLiteral VcpkgCmdArguments::HOST_TRIPLET_ENV;
+    constexpr StringLiteral VcpkgCmdArguments::HOST_TRIPLET_ARG;
```
strega-nil, line 934: now that we've removed VS2015 from our support matrix, are these still necessary?

strega-nil (approved):
Approved with some nits

src/vcpkg/vcpkgcmdarguments.cpp
ras0219: I _thought_ these were for GCC. Regardless, I think it's safest to follow the existing pattern and remove them all in a separate PR.

azure-pipelines/end-to-end-tests.ps1
ras0219: That has different semantics -- it throws an error if the var is not defined instead of returning `$null`.

PR #8, opened 2021-02-10
[vcpkg] Require a baseline when using constraints or overrides

vicroms:
Report the following error when a user tries to use versioned dependencies (`"version>="`) or overrides without setting a baseline.
This validation is only needed when the `versions` feature flag is enabled. 

The error message looks like this:

```
PS D:\demo> vcpkg install
D:\demo\vcpkg.json was rejected because it uses "overrides" or versioned dependencies ("version>=") without setting a "builtin-baseline".
```


strega-nil (approved):

ras0219 (approved):

src/vcpkg/sourceparagraph.cpp
```diff
@@ -1068,6 +1068,22 @@ namespace vcpkg
...
+        {
+            bool has_overrides = !core_paragraph->overrides.empty();
+            bool has_versioned_dependencies = std::any_of(
+                core_paragraph->dependencies.begin(), core_paragraph->dependencies.end(), [](const auto& dependency) {
+                    return dependency.constraint.type != Versions::Constraint::Type::None;
+                });
+
+            if (!core_paragraph->builtin_baseline.has_value() && (has_overrides || has_versioned_dependencies))
+            {
+                return Strings::concat(fs::u8string(origin),
```
ras0219-msft, line 1081: My main suggestion for improving the error message would be to provide as specific of a message as possible, so split up the `||`.

PR #10, opened 2021-02-11
[vcpkg] Add experimental $X_VCPKG_ASSET_SOURCES for source caching

ras0219:
Tool changes for https://github.com/microsoft/vcpkg/pull/13639.

See https://github.com/ras0219/vcpkg/blob/dev/roschuma/azblob-src/docs/users/config-environment.md#x_vcpkg_asset_sources for behavior docs.

The `X_VCPKG_ASSET_SOURCES` is designed to be similar to `X_VCPKG_BINARY_SOURCES` with the intent that they can share backends in the future.

### DIFF
--- a/include/vcpkg/binarycaching.private.h
+++ b/include/vcpkg/binarycaching.private.h
@@ -3,7 +3,6 @@
 #include <vcpkg/fwd/packagespec.h>
 #include <vcpkg/fwd/vcpkgpaths.h>
 
-#include <vcpkg/base/parse.h>
 #include <vcpkg/base/strings.h>
 
 #include <vcpkg/dependencies.h>
--- a/include/vcpkg/binarycaching.private.h
+++ b/include/vcpkg/binarycaching.private.h
@@ -3,6 +3,7 @@
 #include <vcpkg/fwd/packagespec.h>
 #include <vcpkg/fwd/vcpkgpaths.h>
 
+#include <vcpkg/base/parse.h>
 #include <vcpkg/base/strings.h>
 
 #include <vcpkg/dependencies.h>
--- a/include/vcpkg/vcpkgcmdarguments.h
+++ b/include/vcpkg/vcpkgcmdarguments.h
@@ -180,6 +180,9 @@ namespace vcpkg
         constexpr static StringLiteral JSON_SWITCH = "x-json";
         Optional<bool> json = nullopt;
 
+        constexpr static StringLiteral READWRITE_MIRROR_URL_TEMPLATE_ENV = "X_VCPKG_ASSET_SOURCES";
+        Optional<std::string> readwrite_mirror_url_template;
+
         // feature flags
         constexpr static StringLiteral FEATURE_FLAGS_ENV = "VCPKG_FEATURE_FLAGS";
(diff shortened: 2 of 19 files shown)


include/vcpkg/binarycaching.private.h
```diff
@@ -3,6 +3,7 @@
 #include <vcpkg/fwd/packagespec.h>
 #include <vcpkg/fwd/vcpkgpaths.h>
 
+#include <vcpkg/base/parse.h>
```
ras0219, line 6:
```suggestion
```

src/vcpkg/base/files.cpp
```diff
@@ -582,6 +582,25 @@ namespace vcpkg::Files
         return result;
     }
 
+    void Filesystem::create_best_link(const fs::path& to, const fs::path& from, std::error_code& ec)
+    {
+        this->create_hard_link(to, from, ec);
```
strega-nil, line 587: why do we want a hard link here? it seems like we should prefer a symlink, at least to me.

include/vcpkg/vcpkgcmdarguments.h
```diff
@@ -180,6 +180,9 @@ namespace vcpkg
         constexpr static StringLiteral JSON_SWITCH = "x-json";
         Optional<bool> json = nullopt;
 
+        constexpr static StringLiteral READWRITE_MIRROR_URL_TEMPLATE_ENV = "X_VCPKG_ASSET_SOURCES";
```
ras0219, line 183:
```suggestion
        constexpr static StringLiteral ASSET_SOURCES_ENV = "X_VCPKG_ASSET_SOURCES";
```

src/vcpkg/base/files.cpp
ras0219: This is actually old code from a previous implementation and can be simply removed -- the reason to prefer a hard link is that it does not become invalidated by move or remove

vicroms (approved):

src/vcpkg/commands.cpp
```diff
@@ -98,6 +100,7 @@ namespace vcpkg::Commands
             {"x-package-info", &info},
             {"x-history", &porthistory},
             {"x-vsinstances", &vsinstances},
+            {"x-download", &xdownload},
```
dan-shaw, line 103: Should we also add this command to `vcpkg help`?

src/vcpkg/base/downloads.cpp
```diff
@@ -418,9 +467,11 @@ namespace vcpkg::Downloads
...
-        fs.remove(download_path, ignore_errors);
-        fs.remove(download_path_part_path, ignore_errors);
+#if defined(_WIN32)
+        download_path_part_path += fs::u8path(Strings::concat(".", _getpid(), ".part"));
+#else
+        download_path_part_path += fs::u8path(Strings::concat(".", getpid(), ".part"));
+#endif
 
         std::string errors;
         for (const std::string& url : urls)
```
dan-shaw, line 477: Maybe not relevant since this is not part of the PR. If a url is not valid, we may get an unhelpful error message. For example, in line 480 (expand below): `split_uri_view(url)` will print `Value is null` and exit if any url is missing a colon.

dan-shaw (approved):

PR #11, opened 2021-02-14
[vcpkg] support Google Cloud Storage as binary cache

coryan:
Adds a new binary cache provider (configured via `x-gcs`) to use
Google Cloud Storage as the backend. All uploads and downloads
are done via the `gsutil` tool.

I expect this PR to need a lot of additional work, mostly I am seeking
feedback.

There is plenty of room for optimization, this uploads and downloads
one object at a time, and does not take advantage of `gsutil` parallelization
either (aka the `-m` option).

Another future improvement would be to structure the cache to support
`gsutil rsync`.

### DIFF
--- a/src/vcpkg/binarycaching.cpp
+++ b/src/vcpkg/binarycaching.cpp
@@ -852,7 +852,7 @@ namespace
                 if (gsutil_upload_file(gcs_object, tmp_archive_path)) ++upload_count;
             }
 
-            System::print2("Uploaded ", upload_count, " binaries to GCS.\n");
+            System::print2("Uploaded binaries to ", upload_count, " GCS remotes.\n");
         }
         void precheck(const VcpkgPaths&,
                       std::unordered_map<const Dependencies::InstallPlanAction*, RestoreResult>& results_map) override
@@ -1308,7 +1308,7 @@ namespace
             }
             else if (segments[0].second == "x-gcs")
             {
-                // Scheme: x-gsutil,<prefix>[,<readwrite>]
+                // Scheme: x-gcs,<prefix>[,<readwrite>]
                 if (segments.size() < 2)
                 {
                     return add_error("expected arguments: binary config 'gcs' requires at least a prefix",
@@ -1605,7 +1605,7 @@ void vcpkg::help_topic_binary_caching(const VcpkgPaths&)
                "Shared Access Signature validation. URL should include the container path.");
     tbl.format("x-gcs,<prefix>[,<rw>]",
                "**Experimental: will change or be removed without warning** Adds a Google Cloud Storage (GCS) source. "
-               "Uses the gsutil CLI for uploads and downloads. prefix should include the gs:// scheme.");
+               "Uses the gsutil CLI for uploads and downloads. Prefix should include the gs:// scheme and be suffixed with a `/`.");
     tbl.format("interactive", "Enables interactive credential management for some source types");
     tbl.blank();
     tbl.text("The `<rw>` optional parameter for certain strings controls whether they will be consulted for "
--- a/src/vcpkg/binarycaching.cpp
+++ b/src/vcpkg/binarycaching.cpp
@@ -746,6 +746,147 @@ namespace
         bool m_interactive;
         bool m_use_nuget_cache;
     };
+
+    bool gsutil_stat(std::string const& url)
+    {
+        System::Command cmd;
+        cmd.string_arg("gsutil").string_arg("-q").string_arg("stat").string_arg(url);
+        const auto res = System::cmd_execute(cmd);
+        return res == 0;
+    }
+
+    bool gsutil_upload_file(std::string const& gcs_object, fs::path const& archive)
+    {
+        System::Command cmd;
+        cmd.string_arg("gsutil").string_arg("-q").string_arg("cp").path_arg(archive).string_arg(gcs_object);
+        const auto out = System::cmd_execute_and_capture_output(cmd);
+        if (out.exit_code == 0) return true;
+        System::print2(
+            System::Color::warning, "gsutil failed to execute with exit code: ", out.exit_code, '\n', out.output);
+        return false;
+    }
+
+    bool gsutil_download_file(std::string const& gcs_object, fs::path const& archive)
+    {
+        System::Command cmd;
+        cmd.string_arg("gsutil").string_arg("-q").string_arg("cp").string_arg(gcs_object).path_arg(archive);
+        const auto out = System::cmd_execute_and_capture_output(cmd);
+        if (out.exit_code == 0) return true;
+        System::print2(
+            System::Color::warning, "gsutil failed to execute with exit code: ", out.exit_code, '\n', out.output);
+        return false;
+    }
+
+    struct GcsBinaryProvider : NullBinaryProvider
+    {
+        GcsBinaryProvider(std::vector<std::string> read_prefixes, std::vector<std::string> write_prefixes)
+            : m_read_prefixes(std::move(read_prefixes)), m_write_prefixes(std::move(write_prefixes))
+        {
+        }
+
+        void prefetch(const VcpkgPaths& paths, std::vector<const Dependencies::InstallPlanAction*>& actions) override
+        {
+            auto& fs = paths.get_filesystem();
+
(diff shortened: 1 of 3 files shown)


coryan:
/cc: @ras0219-msft as we discussed earlier, here is an initial PR.

src/vcpkg/binarycaching.cpp
```diff
@@ -1160,6 +1306,36 @@ namespace
...
+                if (segments.size() < 2)
+                {
+                    return add_error("expected arguments: binary config 'gcs' requires at least a prefix",
+                                     segments[0].first);
+                }
+
+                if (!Strings::starts_with(segments[1].second, "gs://"))
+                {
+                    return add_error(
+                        "invalid argument: binary config 'gcs' requires an gs:// base url as the first argument",
```
vicroms, line 1321:
```suggestion
                        "invalid argument: binary config 'gcs' requires a gs:// base url as the first argument",
```

small typo

src/vcpkg/binarycaching.cpp
```diff
@@ -1160,6 +1306,36 @@ namespace
...
+                        segments[1].first);
+                }
+
+                if (segments.size() > 3)
+                {
+                    return add_error("unexpected arguments: binary config 'gcs' requires 1 or 2 arguments",
+                                     segments[4].first);
```
vicroms, line 1328: maybe do something similar to `handle_segments()` for more helpful error messages

vicroms (commented):
Thanks for the PR! I know it's still a draft but I took a peek and left some comments.

src/vcpkg/binarycaching.cpp
```diff
@@ -746,6 +746,147 @@ namespace
         bool m_interactive;
         bool m_use_nuget_cache;
     };
+
+    bool gsutil_stat(std::string const& url)
```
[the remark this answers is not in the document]
coryan: Thanks, I knew I was going to get those wrong.  Fixed here and elsewhere, but do let me know if I missed any.

src/vcpkg/binarycaching.cpp
```diff
@@ -746,6 +746,147 @@ namespace
...
+    }
+
+    bool gsutil_upload_file(std::string const& gcs_object, fs::path const& archive)
+    {
+        System::Command cmd;
+        cmd.string_arg("gsutil").string_arg("-q").string_arg("cp").path_arg(archive).string_arg(gcs_object);
+        const auto out = System::cmd_execute_and_capture_output(cmd);
+        if (out.exit_code == 0) return true;
+        System::print2(
+            System::Color::warning, "gsutil failed to execute with exit code: ", out.exit_code, '\n', out.output);
```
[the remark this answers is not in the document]
coryan: Fixed.

src/vcpkg/binarycaching.cpp
```diff
@@ -746,6 +746,147 @@ namespace
...
+    }
+
+    bool gsutil_download_file(std::string const& gcs_object, fs::path const& archive)
+    {
+        System::Command cmd;
+        cmd.string_arg("gsutil").string_arg("-q").string_arg("cp").string_arg(gcs_object).path_arg(archive);
+        const auto out = System::cmd_execute_and_capture_output(cmd);
+        if (out.exit_code == 0) return true;
+        System::print2(
+            System::Color::warning, "gsutil failed to execute with exit code: ", out.exit_code, '\n', out.output);
```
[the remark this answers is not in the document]
coryan: Fixed.

src/vcpkg/binarycaching.cpp
```diff
@@ -746,6 +746,147 @@ namespace
...
+        return false;
+    }
+
+    struct GcsBinaryProvider : NullBinaryProvider
+    {
+        GcsBinaryProvider(std::vector<std::string> read_prefixes, std::vector<std::string> write_prefixes)
```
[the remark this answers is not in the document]
coryan: I made the change, for consistency with the rest of the file. Not what I would do 🤷 .

src/vcpkg/binarycaching.cpp
coryan: Fixed, thanks.

src/vcpkg/binarycaching.cpp
coryan: This is consistent with all the other binary configs.  Happy to change it, but do we want this one to be very different?

src/vcpkg/binarycaching.cpp
```diff
@@ -1160,6 +1306,36 @@ namespace
                 handle_readwrite(
                     state->url_templates_to_get, state->azblob_templates_to_put, std::move(p), segments, 3);
             }
+            else if (segments[0].second == "x-gcs")
+            {
+                // Scheme: x-gsutil,<prefix>[,<readwrite>]
```
ras0219, line 1311:
```suggestion
                // Scheme: x-gcs,<prefix>[,<readwrite>]
```
Unless you'd prefer to rename it everywhere to `x-gsutil`

src/vcpkg/binarycaching.cpp
```diff
@@ -1421,6 +1603,9 @@ void vcpkg::help_topic_binary_caching(const VcpkgPaths&)
     tbl.format("x-azblob,<url>,<sas>[,<rw>]",
                "**Experimental: will change or be removed without warning** Adds an Azure Blob Storage source. Uses "
                "Shared Access Signature validation. URL should include the container path.");
+    tbl.format("x-gcs,<prefix>[,<rw>]",
+               "**Experimental: will change or be removed without warning** Adds a Google Cloud Storage (GCS) source. "
+               "Uses the gsutil CLI for uploads and downloads. prefix should include the gs:// scheme.");
```
ras0219, line 1608:
```suggestion
               "Uses the gsutil CLI for uploads and downloads. Prefix should include the gs:// scheme and be suffixed with a `/`.");
```

src/vcpkg/binarycaching.cpp
```diff
@@ -746,6 +746,147 @@ namespace
         bool m_interactive;
         bool m_use_nuget_cache;
     };
+
+    bool gsutil_stat(const std::string& url)
+    {
+        System::Command cmd;
+        cmd.string_arg("gsutil").string_arg("-q").string_arg("stat").string_arg(url);
```
ras0219, line 753: We may want to add `gsutil` to https://github.com/microsoft/vcpkg-tool/blob/main/src/vcpkg/tools.cpp to eventually support automatic download via `vcpkgTools.xml` (https://github.com/microsoft/vcpkg/blob/master/scripts/vcpkgTools.xml).

ras0219 (commented):
Looks great so far 👍

src/vcpkg/binarycaching.cpp
coryan: That does seem like a good idea.  I doubt I will have time to do this until the weekend. We can wait or merge now, your call, but I prefer to merge sooner rather than later.

coryan:
> Looks great so far 👍. I think this is close to mergeable as an initial implementation, 

Cool.

> though supporting parallelism during download would be really great.

`gsutil` likes to puts all the downloaded files in a single destination directory.  It seems to me that `vcpkg` (when using `curl`) takes advantage of the `-o` option to put each file in separate directories?  Is there an acceptable directory to use as staging for the downloads and then do an `fs::????` operation to move them?  Sorry if this should be obvious, the code is fairly clear, but I am still figuring out the lay of the land.

ras0219-msft:
You can use a subdirectory of `buildtrees` that contains an illegal identifier character, such as `_`. So, `paths.buildtrees / "_gsutil"`, for example.

vicroms (approved):

ras0219-msft:
@coryan Let me know if you'd like this to be merged or if you're still working on it.

coryan:
Let's merge as-is, I am working on the parallel download but will take me a while. Thanks for checking on me.

ras0219-msft:
Awesome, thanks!

When we push a binary tool update, I'll ping you to potentially add a docs blurb to binarycaching.md

coryan:
Happy to send patches to the documentation too.  I also created a couple of bugs to implement the suggestions (requests?) that came up during the review. Feel free to assign them to me.

PR #13, opened 2021-02-17
[ABI] ABI calculation should include ports.cmake

strega-nil:


src/vcpkg/build.cpp
```diff
@@ -1043,6 +1043,7 @@ namespace vcpkg::Build
             }
         }
 
+        abi_tag_entries.emplace_back("ports.cmake", vcpkg::Hash::get_file_hash(VCPKG_LINE_INFO, fs, paths.ports_cmake, Hash::Algorithm::Sha1));
```
ras0219, line 1046: Since this should not change during a given invocation of vcpkg, it would be great to cache it inside VcpkgPaths, similar to the existing `VcpkgPaths::get_cmake_script_hashes()`.

ras0219 (approved):

ras0219-msft (approved):

PR #17, opened 2021-02-23
Add vcpkg policy cmake helper port support

strega-nil:
See PR microsoft/vcpkg#16377

### DIFF
--- a/src/vcpkg/postbuildlint.cpp
+++ b/src/vcpkg/postbuildlint.cpp
@@ -243,7 +243,8 @@ namespace vcpkg::PostBuildLint
                                                   const fs::path& package_dir,
                                                   const PackageSpec& spec)
     {
-        const fs::path relative_path = fs::u8path("share") / fs::u8path(spec.name()) / fs::u8path("vcpkg-port-config.cmake");
+        const fs::path relative_path =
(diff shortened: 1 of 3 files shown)


src/vcpkg/postbuildlint.cpp
```diff
@@ -222,6 +238,31 @@ namespace vcpkg::PostBuildLint
...
+
+        if (policies.is_enabled(BuildPolicy::CMAKE_HELPER_PORT))
+        {
+            if (!fs.exists(absolute_path))
+            {
+                System::print2(System::Color::warning,
+                               "The /",
+                               fs::u8string(relative_path),
+                               " "
+                               "file does not exist. This file must exist for CMake helper ports.\n");
```
vicroms, line 258: This line break is strange.

vicroms (approved):

src/vcpkg/postbuildlint.cpp
strega-nil: that is strange...

strega-nil: however, it is the way that clang-format wants to do it...

vicroms (approved):

PR #21, opened 2021-03-02
Add new telemetry points for versioning

vicroms:
Adds telemetry for the following properties:

Name | Description
-- | --
`versioning-error-disabled` | User has declared  `builtin-baseline`, `version>=` or   `overrides` without enabling the versions feature flag
`versioning-error-no-baseline` | User has declared `version>=` or `overrides` but not a baseline
`versioning-error-baseline` | (**WIP**) User has declared an invalid baseline 
`versioning-error-version` | User has declared a constraint on a non-existing version
`manifest_version_constraint` | User has declared `version>=`

### DIFF
--- a/src/vcpkg/portfileprovider.cpp
+++ b/src/vcpkg/portfileprovider.cpp
@@ -2,6 +2,7 @@
 #include <vcpkg/base/system.debug.h>
 
 #include <vcpkg/configuration.h>
+#include <vcpkg/metrics.h>
 #include <vcpkg/paragraphs.h>
 #include <vcpkg/portfileprovider.h>
 #include <vcpkg/registries.h>
@@ -193,6 +194,7 @@ namespace vcpkg::PortFileProvider
                     }
                     else
                     {
+                        Metrics::g_metrics.lock()->track_property("versioning-error-version", "defined");
                         return maybe_path.error();
                     }
                 }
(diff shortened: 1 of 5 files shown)


src/vcpkg/portfileprovider.cpp
```diff
@@ -193,6 +194,7 @@ namespace vcpkg::PortFileProvider
                     }
                     else
                     {
+                        Metrics::g_metrics.lock()->track_property("versioning-error-version", "defined");
```
ras0219, line 197: This is a bit imprecise because we don't actually know whether this is a termination-worthy error, however it may be intractable to move this to every `load_control_file().value_or_exit()`.

ras0219: The more invasive but "correct" approach would be to create a new type like:
```cpp
struct TrackedError {
    std::string message;
    std::string metric;
};
```
which upon calling `Expected<T,TrackedError>::value_or_exit()` would track the property just in time to exit. However, I think that is probably too invasive at this time.

ras0219 (approved):

src/vcpkg/portfileprovider.cpp
vicroms: I may attempt to track in `value_or_exit()` in a separate PR.
For the time being, I'll just merge with the current metrics if there are no objections.

PR #22, opened 2021-03-03
Fix CURL filesystem deadlock with corporate proxy

SoftwareApe:
Prevent CURL hanging up without timeout when connection can't be established. Creating a deadlock on the filesystem when executed behind a proxy.

As requested by @JackBoosY, this is PR https://github.com/microsoft/vcpkg/pull/16499, retargeted to vcpkg-tool. Now with longer timeout.

### DIFF
--- a/src/vcpkg/metrics.cpp
+++ b/src/vcpkg/metrics.cpp
@@ -483,6 +483,8 @@ namespace vcpkg::Metrics
         // TODO: convert to cmd_execute_background or something.
         auto curl = System::Command("curl")
                         .string_arg("https://dc.services.visualstudio.com/v2/track")
+                        .string_arg("--max-time")
+                        .string_arg("3")
                         .string_arg("-H")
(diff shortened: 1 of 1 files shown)


ras0219 (approved):

SoftwareApe:
@ras0219 before merging, @JackBoosY said the file lock should be reset after vcpkg is finished and before the curl call. I think this should be done before merging.

JackBoosY:
That should be what @ras0219-msft said.

ras0219-msft:
I think this is a good change even without resetting the file lock -- though we should do that as well.

LGTM, thanks for the PR!

SoftwareApe:
@ras0219-msft I do agree that the merged change already helps, but due to the timeout it may slow down developers, and now not become easy enough to spot anymore (because the failure is *just* slowness, not outright failure). I hope we can somehow still make the actual fix that the filesystem lock isn't held longer than necessary.

PR #24, opened 2021-03-09
Enable build scripts to know if we are in manifest mode

PazerOP:
Required for https://github.com/microsoft/vcpkg/pull/16613 to work, otherwise it does nothing.

I did not see any other way to determine if manifests were being used without making this code change. If there is a way for https://github.com/microsoft/vcpkg/pull/16613 to work without this code change, please let me know.


ras0219-msft:
Blocked on https://github.com/microsoft/vcpkg/pull/16613#issuecomment-799801106

PR #26, opened 2021-03-09
Add CUDA 11.1 and 11.2 to KEEP_ENV_VARS defaults

upsj:
vcpkg right now can't build CUDA ports with version 11.1 and 11.2, since the associated environment variables aren't being forwarded.

### DIFF
--- a/src/vcpkg/base/system.process.cpp
+++ b/src/vcpkg/base/system.process.cpp
@@ -313,6 +313,8 @@ namespace vcpkg
             L"CUDA_PATH_V10_1",
             L"CUDA_PATH_V10_2",
             L"CUDA_PATH_V11_0",
+            L"CUDA_PATH_V11_1",
+            L"CUDA_PATH_V11_2",
             L"CUDA_TOOLKIT_ROOT_DIR",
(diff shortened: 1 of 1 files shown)


vicroms (approved):

ras0219-msft:
This LGTM, though we should continue to search for a better solution for CUDA in the future that doesn't require modifying the vcpkg binary.

upsj:
@ras0219-msft While this fix came out of my attempt to fix CMake CUDA language support in the vcpkg environment, in the end this didn't fix anything, since as far as I can tell the versioned CUDA_PATH environment variables are only being used by the MSVC CMake generator, which doesn't seem to be used that often across vcpkg, if at all? The underlying cause is that until now, the CMake generators that don't come directly with associated compilers require nvcc to be present in PATH, not just reachable through CUDA_PATH.

This will [probably be fixed in CMake 3.21](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5908), but of course I cannot vouch for how well this works in environments with multiple CUDA versions, users will probably have to specify and KEEP_ENV_VARS CUDACXX manually there for this to work with Ninja.

PR #27, opened 2021-03-10
Remove baseline warning

strega-nil:
we don't want people to use `baseline` necessarily; `builtin-baseline` is also completely valid.

### DIFF
--- a/src/vcpkg/registries.cpp
+++ b/src/vcpkg/registries.cpp
@@ -1158,18 +1158,24 @@ namespace vcpkg
             }
             else
             {
-                System::print2(System::Color::warning,
-            R"(warning: attempting to set builtin baseline in both vcpkg.json and vcpkg-configuration.json
+                System::print2(
+                    System::Color::warning,
+                    R"(warning: attempting to set builtin baseline in both vcpkg.json and vcpkg-configuration.json
     (only one of these should be used; the baseline from vcpkg-configuration.json will be used))");
             }
         }
         else if (auto default_registry = default_registry_.get())
         {
-            System::printf(System::Color::warning, "warning: the default registry has been replaced with a %s registry, but `builtin-baseline` is specified in vcpkg.json. This field will have no effect.", default_registry->kind());
+            System::printf(System::Color::warning,
+                           "warning: the default registry has been replaced with a %s registry, but `builtin-baseline` "
+                           "is specified in vcpkg.json. This field will have no effect.",
+                           default_registry->kind());
         }
         else
         {
-            System::print2(System::Color::warning, "warning: the default registry has been disabled, but `builtin-baseline` is specified in vcpkg.json. This field will have no effect.");
+            System::print2(System::Color::warning,
+                           "warning: the default registry has been disabled, but `builtin-baseline` is specified in "
+                           "vcpkg.json. This field will have no effect.");
         }
     }
 
--- a/src/vcpkg/registries.cpp
+++ b/src/vcpkg/registries.cpp
@@ -1155,14 +1155,20 @@ namespace vcpkg
         if (auto builtin_registry = dynamic_cast<BuiltinRegistry*>(default_registry_.get()))
         {
             default_registry_is_builtin = true;
-            builtin_registry->m_baseline_identifier.assign(baseline.begin(), baseline.end());
+            if (builtin_registry->m_baseline_identifier.empty())
+            {
+                builtin_registry->m_baseline_identifier.assign(baseline.begin(), baseline.end());
+            }
         }
 
         for (auto& reg : registries_)
         {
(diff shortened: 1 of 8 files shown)


src/vcpkg/registries.cpp
```diff
@@ -1143,22 +1143,20 @@ namespace vcpkg
...
-
-        if (!default_registry_is_builtin || registries_.size() != 0)
-        {
-            System::print2(System::Color::warning,
-                           "Warning: when using the registries feature, one should not use `\"builtin-baseline\"` "
-                           "to set the baseline.\n",
-                           "    Instead, use the \"baseline\" field of the registry.\n");
+            if (builtin_registry->m_baseline_identifier.empty())
+            {
+                builtin_registry->m_baseline_identifier.assign(baseline.begin(), baseline.end());
```
ras0219-msft, line 1148: We should warn if `m_baseline_identifier.empty() && !baseline.empty()`

strega-nil: I'm not sure about this; for example:

vcpkg.json:
```json
{
  ...,
  "builtin-baseline": "b1"
```

vcpkg-configuration.json:
```
{
  "registries": [
    {
      "kind": "builtin",
      "baseline": "b2",
      "packages": [ "a", "b", "c" ]
    }
  ]
}
```

ras0219-msft: Right; so in this case we should warn to the user about which of the two baselines they've specified will be used (I assume b2?)

strega-nil: I mean, they'll both be used: `b2` will be used for packages `a`, `b`, and `c`, and `b1` will be used for everything else.

vicroms: I thought that having `"packages": [ "a", "b", "c" ]` would disable all other packages from the built-in registry. 

If not, then trying to install package `d` when there is no `"builtin-baseline": "b1"` and also attempting to use version constraints will fail and we should warn about that.

strega-nil: You can disable the builtin registry, but you must explicitly do so by setting `default-registry` to `null`.

vicroms (approved):

src/vcpkg/registries.cpp
ras0219-msft: What is the syntax for changing the default registry's baseline? I suspect this means we should warn in the case that the user has specified `"builtin-baseline"` AND specifies `"default-registry"`. If that is already the case in this PR, then this comment is resolved.

src/vcpkg/registries.cpp
```diff
@@ -1138,27 +1138,37 @@ namespace vcpkg
...
-        bool default_registry_is_builtin = false;
-        if (auto builtin_registry = dynamic_cast<BuiltinRegistry*>(default_registry_.get()))
-        {
-            default_registry_is_builtin = true;
-            builtin_registry->m_baseline_identifier.assign(baseline.begin(), baseline.end());
-        }
+        static const StringLiteral warning_message =
+            R"(warning: attempting to set builtin baseline in both vcpkg.json and vcpkg-configuration.json
+    (only one of these should be used; the baseline from vcpkg-configuration.json will be used))";
+        bool already_warned = false;
```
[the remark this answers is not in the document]
strega-nil: hmm, probably. Not gonna add it here, I'll maybe make that PR in future.

src/vcpkg/registries.cpp
strega-nil: The syntax for changing the default registry's baseline is:

```json
{
  "default-registry": {
    "kind": "...",
    ...,
    "baseline": "<baseline>"
  }
}
```

This is not warned against in this PR; what should the warning look like, do you think?

ras0219-msft: "Warning: Specifying the "default-registry" overrides the "builtin-baseline" field. Pass <blah> to suppress this warning message."

strega-nil: Do we only want `builtin-baseline` to do something to a registry that is _both_ builtin, and default? Currently it does something to registries that are builtin, even if they're non-default.

ras0219-msft: It should only affect the default builtin baseline in my opinion. I see the following scenarios:
1. No `vcpkg-configuration.json` (no warn)
2. adds other registries but doesn't change the default (no warn)
3. doesn't change the default, but _does_ re-add the builtin registry to get packages with a different baseline (no warn)
4. re-defines the default to builtin, but at a different baseline (warn)

I think the most straightforward UX is _all_ registries in the configuration file _must_ have a baseline specified. Therefore, the only case that `builtin-baseline` could possibly apply would be in the unspecified default case.

ytbigvinny27 (approved):

ytbigvinny27 (approved):

PR #29, opened 2021-03-12
[supports] Add `native` identifier expression and x-check-support command

strega-nil:
`native` means "TARGET_TRIPLET == HOST_TRIPLET

the `x-check-support` command allows one to check if a port supports a platform.

### DIFF
--- a/include/vcpkg/cmakevars.h
+++ b/include/vcpkg/cmakevars.h
@@ -28,10 +28,12 @@ namespace vcpkg::CMakeVars
         virtual void load_dep_info_vars(Span<const PackageSpec> specs) const = 0;
 
         virtual void load_tag_vars(Span<const FullPackageSpec> specs,
-                                   const PortFileProvider::PortFileProvider& port_provider) const = 0;
+                                   const PortFileProvider::PortFileProvider& port_provider,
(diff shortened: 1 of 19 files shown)


include/vcpkg/cmakevars.h
```diff
@@ -28,10 +28,12 @@ namespace vcpkg::CMakeVars
...
 
         virtual void load_tag_vars(Span<const FullPackageSpec> specs,
-                                   const PortFileProvider::PortFileProvider& port_provider) const = 0;
+                                   const PortFileProvider::PortFileProvider& port_provider,
+                                   Triplet host_triplet) const = 0;
 
         void load_tag_vars(const vcpkg::Dependencies::ActionPlan& action_plan,
-                           const PortFileProvider::PortFileProvider& port_provider) const;
+                           const PortFileProvider::PortFileProvider& port_provider,
+                           Triplet host_triplet) const;
```
ras0219-msft, line 36: While there is not a way for this to be specified now, I believe the host triplet can actually be retrieved from each InstallAction and they _very hypothetically_ could be different.

No changes requested, this is just a comment.

ras0219-msft (approved):
Needs documentation changes as well :)

PR #31, opened 2021-03-17
Don't build tls12-download with MINGW

dg0yt:
In MSYS2 MinGW environments on Windows, the build of `tls12-download` fails due to missing symbols. This change disables this component, similar to the situation on non-Windows systems. And this particular configuration of tls12-download is not needed because a binary from a MSVC build is already part of the vcpkg-repository.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -190,7 +190,7 @@ endif()
 # === Target: tls12-download ===
 
 set(TLS12_DOWNLOAD_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/src/tls12-download.c)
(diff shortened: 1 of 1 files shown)


strega-nil:
I'd prefer to do #33. There's no need to build tls12-download anyways.

strega-nil:
(accidentally hit close w/ comment instead of just comment)

dg0yt:
This is okay. (However, #33 does a lot of changes in one commit.)

dg0yt:
This PR is obsolete now. The merge of #33 made the tls12-download build optional and add a fix for MinGW.

PR #33, opened 2021-03-17
don't build tls12-download unless it's needed

strega-nil:
Nobody needs to build it except us, so there's no need to.
Additionally, on mingw, tls12-download fails to build due,
so this also fixes our build on mingw.

Alternative fix to #31

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,9 +5,6 @@ else()
 cmake_minimum_required(VERSION 3.14)
 endif()
 
-project(vcpkg C CXX)
-include(cmake/utilities.cmake)
-
 # ===============
 # === Options ===
 # ===============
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -17,36 +14,60 @@ include(CMakeDependentOption)
...
-file(GLOB VCPKGLIB_BASE_INCLUDES CONFIGURE_DEPENDS include/vcpkg/base/*.h include/vcpkg/base/fwd/*.h)
-
-set(VCPKG_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/src/vcpkg.cpp)
-
-file(GLOB VCPKG_TEST_SOURCES CONFIGURE_DEPENDS src/vcpkg-test/*.cpp)
-file(GLOB VCPKG_TEST_INCLUDES CONFIGURE_DEPENDS include/vcpkg-test/*.h)
-
-file(GLOB VCPKG_FUZZ_SOURCES CONFIGURE_DEPENDS src/vcpkg-fuzz/*.cpp)
+file(GLOB VCPKGLIB_SOURCES CONFIGURE_DEPENDS
+    "${CMAKE_CURRENT_SOURCE_DIR}/src/vcpkg/*.cpp"
```
ras0219-msft, line 43: Why do these need to be qualified with `CMAKE_CURRENT_SOURCE_DIR` (is it clang-format)? If it is optional, I would prefer to omit it.

CMakeLists.txt
```diff
@@ -17,36 +14,60 @@ include(CMakeDependentOption)
 option(BUILD_TESTING "Option for enabling testing" ON)
 option(VCPKG_ALLOW_APPLE_CLANG "Option for allowing apple clang, even versions that we don't know will work" OFF)
 option(VCPKG_DEVELOPMENT_WARNINGS "Option for turning on all warnings" ON)
-option(VCPKG_WARNINGS_AS_ERRORS "Set warnings to be errors" ${VCPKG_DEVELOPMENT_WARNINGS})
+option(VCPKG_WARNINGS_AS_ERRORS "Set warnings to be errors" OFF)
+option(VCPKG_BUILD_TLS12_DOWNLOADER "Enable building the tls12-downloader" OFF)
 option(VCPKG_BUILD_FUZZING "Option for enabling vcpkg-fuzz support" OFF)
 option(VCPKG_EMBED_GIT_SHA "Option for to fill in the Git SHA version; off by default to avoid privacy concerns out of official builds" OFF)
 
+
```
ras0219-msft, line 22:
```suggestion
```

CMakeLists.txt
```diff
@@ -17,36 +14,60 @@ include(CMakeDependentOption)
...
-endif()
 if(DEFINE_DISABLE_METRICS OR VCPKG_DISABLE_METRICS)
     message(DEPRECATION "DEFINE_DISABLE_METRICS / VCPKG_DISABLE_METRICS are now handled by creating a "
         "file vcpkg.disable_metrics next to the binary.")
 endif()
 
+if(VCPKG_BUILD_TLS12_DOWNLOADER AND NOT WIN32)
+    message(FATAL_ERROR "The tls12-downloader binary is only supported on Windows.")
+endif()
+
```
ras0219-msft, line 34:
```suggestion
```
This seems unnecessarily pedantic; it's off-by-default.

ras0219-msft (approved):

CMakeLists.txt
strega-nil: I mostly wanted it to make it similar to the `set` input file commands, it's not necessary.

CMakeLists.txt
strega-nil: sure, wfm.

PR #35, opened 2021-03-20
Use only named packages from extra registries

dg0yt:
This change resolves errors from RegistryEntry::get_path_to_version() when the default registry doesn't have a port for a name which is part of some extra registry but not listed in that registry's "packages" configuration field.
This is consistent with RegistrySet::registry_for_port() which only considers the the named packages from the extra registries.

Use case:
~~~
{
  "default-registry": {
    "kind": "filesystem",
    "path": "/my/tiny/repository"
  },
  "registries": [
    {
      "kind": "git",
      "repository": "https://github.com/microsoft/vcpkg",
      "packages": [ "vcpkg-cmake", "vcpkg-cmake-config" ]
    }
  ]
}
~~~

### DIFF
--- a/src/vcpkg/paragraphs.cpp
+++ b/src/vcpkg/paragraphs.cpp
@@ -426,7 +426,8 @@ namespace vcpkg::Paragraphs
 
         for (const auto& registry : registries.registries())
         {
-            registry.implementation().get_all_port_names(ports, paths);
+            const auto packages = registry.packages();
+            ports.insert(end(ports), begin(packages), end(packages));
         }
         if (auto registry = registries.default_registry())
         {


ras0219-msft:
Thanks for the PR! This change looks generally good to me, however could you go into more detail about what the precise error is and what the expected result was?

strega-nil:
FWIW, this is why https://github.com/microsoft/vcpkg-tool/pull/35/files#diff-d4fe92b1ebc56fd05568a18c509672434c578cf4e147b2d5aad36226c026ff00L440-L447 exists; is there a reason that you're making this change?

dg0yt:
> is there a reason that you're making this change?

I want to combine
- a custom registry of ports ("default-registry")
- selected ports from the main vcpkg repository ("registries"[0]).

> could you go into more detail about  what the precise error is and what the expected result was?

vcpkg commands fail like this:
~~~
$ vcpkg search internal
Error: Couldn't find the versions database file: /tmp/test/versions/3-/3fd.json
~~~

where 3fd.json neither exists in default-registry nor is listed registries[0].packages.

Reproducer (bash, but modeled after azure-pipelines/end-to-end-tests-dir/registries.ps1):
~~~
test -d vcpkg-tool || git clone https://github.com/microsoft/vcpkg-tool.git

mkdir -p ports
cp -R vcpkg-tool/azure-pipelines/e2e_ports/vcpkg-internal-e2e-test-port ports/

mkdir -p versions/v-
cp vcpkg-tool/azure-pipelines/e2e_ports/versions/baseline.json versions/
cat > versions/v-/vcpkg-internal-e2e-test-port.json <<END_PORT
{
  "versions" : {
    "version-string": "1.0.0",
    "path": "$PWD/ports/vcpkg-internal-e2e-test-port"
  }
}
END_PORT

test -d vcpkg || git clone https://github.com/microsoft/vcpkg.git
cat > vcpkg/vcpkg-configuration.json <<END_CONFIG
{
  "default-registry": {
    "kind": "filesystem",
    "path": "$PWD"
  },
  "registries": [
    {
      "kind": "git",
      "repository": "https://github.com/microsoft/vcpkg",
      "packages": [ "vcpkg-cmake", "vcpkg-cmake-config" ]
    }
  ]
}
END_CONFIG

export VCPKG_ROOT=$PWD/vcpkg
export VCPKG_FEATURE_FLAGS=registries

vcpkg --debug search internal
~~~

ras0219-msft (approved):
This LGTM, loading all registry ports shouldn't enumerate ports beyond default + `packages` of each other registry. It should exactly match the valid values of `RegistrySet::registry_for_port()`.

ras0219-msft:
@strega-nil `RegistrySet::registry_for_port()` doesn't guarantee that the port name is valid for the registry, so the `if (!impl)` check doesn't handle @dg0yt's case.

PR #36, opened 2021-03-23
Disable git autocrlf when archiving tree

aee11:
Ensures that file endings are consistent between the ports and the versioned port under `versioning`.

Fixes https://github.com/microsoft/vcpkg/issues/16615#issuecomment-804928402 (verified locally). Before this fix, a patch with LF endings could end up with CRLF endings on Windows causing corrupt patch issues.

### DIFF
--- a/src/vcpkg/vcpkgpaths.cpp
+++ b/src/vcpkg/vcpkgpaths.cpp
@@ -776,6 +776,7 @@ If you wish to silence this error and use classic mode, you can:
         auto tar_cmd_builder = git_cmd_builder(*this, dot_git_dir, dot_git_dir)
                                    .string_arg("archive")
                                    .string_arg(git_tree)
+                                   .string_arg("--worktree-attributes")
                                    .string_arg("-o")
                                    .path_arg(destination_tar);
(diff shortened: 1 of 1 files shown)


src/vcpkg/vcpkgpaths.cpp
```diff
@@ -776,6 +776,7 @@ If you wish to silence this error and use classic mode, you can:
         auto tar_cmd_builder = git_cmd_builder(*this, dot_git_dir, dot_git_dir)
                                    .string_arg("archive")
                                    .string_arg(git_tree)
+                                   .string_arg("--worktree-attributes")
```
ras0219-msft, line 779:
```suggestion
                                   .string_arg("-c").string_arg("core.autocrlf=false")
```
I think it may be better to instead directly set the relevant attribute instead of trying to incidentally find it in a `.gitattributes` file.

aee11: Yes, that makes total sense. I've applied your suggestion.

ras0219-msft (approved):

ras0219-msft:
Thanks for the fix!

PR #43, opened 2021-04-05
[vcpkg] x-add-version now also checks if the manifest file is properly formatted

autoantwort:
Closes microsoft/vcpkg#16880

### DIFF
--- a/azure-pipelines/end-to-end-tests-dir/versions.ps1
+++ b/azure-pipelines/end-to-end-tests-dir/versions.ps1
@@ -26,35 +26,35 @@ Throw-IfNotFailed
 
 $CurrentTest = "x-add-version cat"
 # Do not fail if there's nothing to update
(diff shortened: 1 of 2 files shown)


strega-nil:
I think this is a good idea, although it needs a `--skip-manifest-checks` or something flag.

autoantwort:
Are the tests using the latest master with microsoft/vcpkg#17120 ?

autoantwort:
> Are the tests using the latest master with [microsoft/vcpkg#17120](https://github.com/microsoft/vcpkg/pull/17120) ?

Or should I add `--skip-formatting-check` [here](https://github.com/microsoft/vcpkg-tool/blob/main/azure-pipelines/end-to-end-tests-dir/versions.ps1#L13)

autoantwort:
This is really needed. Especially now when the PR template asks if you have run `x-add-version`

PR #44, opened 2021-04-05
[vcpkg] x-add-version improve speed by calling get_builtin_baseline only once

autoantwort:
Currently get_builtin_baseline is called for every port and 95% of the runtime
is spend inside of get_builtin_baseline when using --all

On my system now `./vcpkg x-add-version --all` runs 40 times faster, 300 ms instead of 12 seconds.


vicroms (approved):

ras0219-msft:
Awesome, thanks for the big improvement!

PR #50, opened 2021-04-09
[vcpkg] allow --version to check the version

autoantwort:
I don't know if my approach is the best one.  
Fixes:
image
:D

### DIFF
--- a/src/vcpkg/vcpkgcmdarguments.cpp
+++ b/src/vcpkg/vcpkgcmdarguments.cpp
@@ -233,6 +233,12 @@ namespace vcpkg
         VcpkgCmdArguments args;
         std::vector<std::string> feature_flags;
 
+        if (arg_first != arg_last && arg_first + 1 == arg_last && *arg_first == "--version")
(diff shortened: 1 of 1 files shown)


strega-nil:
requires:discussion on best way to actually do this

src/vcpkg/vcpkgcmdarguments.cpp
```diff
@@ -260,6 +260,11 @@ namespace vcpkg
                 }
                 continue;
             }
+            if (basic_arg == "--version" && args.command.empty())
```
ras0219-msft, line 263: I'd suggest moving this up to the top as:
```cpp
if (arg_first != arg_last && arg_first + 1 == arg_last && *arg_first == "--version") {
    args.command = "version";
    return args;
}
```

autoantwort: Yeah that sounds better

strega-nil:
Thanks @autoantwort

PR #52, opened 2021-04-11
[vcpkg] add command init-registry

autoantwort:
The next step is to support `x-add-version` in registries.

### DIFF
--- a/src/vcpkg/commands.cpp
+++ b/src/vcpkg/commands.cpp
@@ -18,6 +18,7 @@
 #include <vcpkg/commands.h>
 #include <vcpkg/commands.hash.h>
 #include <vcpkg/commands.info.h>
+#include <vcpkg/commands.init-registry.h>
 #include <vcpkg/commands.integrate.h>
 #include <vcpkg/commands.list.h>
 #include <vcpkg/commands.owns.h>
@@ -41,6 +42,7 @@ namespace vcpkg::Commands
     {
         static const Version::VersionCommand version{};
         static const Contact::ContactCommand contact{};
+        static const InitRegistry::InitRegistryCommand initRegistry{};
 #if defined(_WIN32)
         static const UploadMetrics::UploadMetricsCommand upload_metrics{};
 #endif // defined(_WIN32)
@@ -48,6 +50,8 @@ namespace vcpkg::Commands
         static std::vector<PackageNameAndFunction<const BasicCommand*>> t = {
             {"version", &version},
             {"contact", &contact},
+            {"init-registry", &initRegistry},
+
 #if defined(_WIN32)
             {"x-upload-metrics", &upload_metrics},
 #endif // defined(_WIN32)
--- a/src/vcpkg/commands.cpp
+++ b/src/vcpkg/commands.cpp
@@ -42,7 +42,7 @@ namespace vcpkg::Commands
     {
         static const Version::VersionCommand version{};
         static const Contact::ContactCommand contact{};
-        static const InitRegistry::InitRegistryCommand initRegistry{};
+        static const InitRegistry::InitRegistryCommand init_registry{};
 #if defined(_WIN32)
         static const UploadMetrics::UploadMetricsCommand upload_metrics{};
 #endif // defined(_WIN32)
@@ -50,7 +50,7 @@ namespace vcpkg::Commands
         static std::vector<PackageNameAndFunction<const BasicCommand*>> t = {
             {"version", &version},
             {"contact", &contact},
-            {"x-init-registry", &initRegistry},
+            {"x-init-registry", &init_registry},
 
 #if defined(_WIN32)
             {"x-upload-metrics", &upload_metrics},
--- a/src/vcpkg/commands.cpp
+++ b/src/vcpkg/commands.cpp
@@ -50,7 +50,7 @@ namespace vcpkg::Commands
         static std::vector<PackageNameAndFunction<const BasicCommand*>> t = {
             {"version", &version},
             {"contact", &contact},
-            {"init-registry", &initRegistry},
+            {"x-init-registry", &initRegistry},
 
 #if defined(_WIN32)
             {"x-upload-metrics", &upload_metrics},
--- a/src/vcpkg/commands.init-registry.cpp
+++ b/src/vcpkg/commands.init-registry.cpp
@@ -6,24 +6,20 @@
 
 namespace vcpkg::Commands::InitRegistry
 {
+    static const CommandStructure COMMAND_STRUCTURE = {
+        create_example_string(R"(x-init-registry .)"),
+        1,
+        1,
+        {{}, {}, {}},
+        nullptr,
+    };
+
     void perform_and_exit(const VcpkgCmdArguments& args, Files::Filesystem& fs)
     {
-        if (args.command_arguments.empty())
-        {
-            Checks::exit_with_message(
-                VCPKG_LINE_INFO,
-                "You have to provide a path to a directory where the registry should be created as argument. "
-                "Pass '.' to create the registry in the current directory.");
-        }
-        if (args.command_arguments.size() > 1)
-        {
-            Checks::exit_with_message(VCPKG_LINE_INFO, "Only pass one directory as argument to this command.");
-        }
+        auto parsed_args = args.parse_arguments(COMMAND_STRUCTURE);
+
         auto path = fs::u8path(args.command_arguments.front());
-        if (path.is_relative())
-        {
-            path = fs.current_path(VCPKG_LINE_INFO) / path;
-        }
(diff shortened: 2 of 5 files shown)


autoantwort:
For `x-add-version` I see two possibilities: 
1. Make `x-add-version` a `BasicCommand` and determine the `ports` and `versions` dir inside of `perform_and_exit`, but this is maybe a bad idea because you also want to check `--x-builtin-ports-root` and `--x-builtin-registry-versions-dir` and you maybe also want to change other commands so that they word in registries (`vcpkg edit/create`)
2. Add a new class `VcpkgBasePaths` that contains the `.git`, `ports` and the `versions` folder, but not the rest of `VcpkgPaths` and remove the paths of `VcpkgBasePaths` from `VcpkgPaths`. Add than a new command category `BasePathsCommand`.

I think the second solution is cleaner, but I don't want to implement something and have to change everything later. So which version do you prefer (or maybe you have another idea how to solve this)?

src/vcpkg/commands.cpp
```diff
@@ -41,13 +42,16 @@ namespace vcpkg::Commands
...
         static const Contact::ContactCommand contact{};
+        static const InitRegistry::InitRegistryCommand initRegistry{};
 #if defined(_WIN32)
         static const UploadMetrics::UploadMetricsCommand upload_metrics{};
 #endif // defined(_WIN32)
 
         static std::vector<PackageNameAndFunction<const BasicCommand*>> t = {
             {"version", &version},
             {"contact", &contact},
+            {"init-registry", &initRegistry},
```
strega-nil, line 53:
```suggestion
            {"x-init-registry", &initRegistry},
```

this should be an experimental command.

src/vcpkg/vcpkgcmdarguments.cpp
```diff
@@ -565,6 +565,7 @@ namespace vcpkg
         table.format("vcpkg edit <pkg>",
                      "Open up a port for editing (uses " + format_environment_variable("EDITOR") + ", default 'code')");
         table.format("vcpkg create <pkg> <url> [archivename]", "Create a new package");
+        table.format("vcpkg init-registry", "Initializes a registry in the current directory");
```
strega-nil, line 568:
```suggestion
        table.format("vcpkg x-init-registry", "Initializes a registry in the current directory");
```

see above

src/vcpkg/commands.init-registry.cpp
```diff
@@ -0,0 +1,31 @@
...
+        }
+        if (!fs.exists(baseline))
+        {
+            const auto content = R"({
+  "default": {}
+})";
+            fs.write_contents_and_dirs(baseline, content, VCPKG_LINE_INFO);
+        }
+        System::print2("Sucessfully created registry at ", fs::u8string(fs.current_path(VCPKG_LINE_INFO)), "\n");
+        Checks::exit_success(VCPKG_LINE_INFO);
```
strega-nil, line 24: This should take a path.

strega-nil (changes requested):

src/vcpkg/commands.init-registry.cpp
autoantwort: So that you can specify where the registry should be created?

ras0219-msft: Yeah, I think it's very likely new users will not expect it to "just do it" in the current directory. It would be great if `vcpkg init-registry` would print help, and then users could use `vcpkg init-registry .` to "just do it" in the cwd.

autoantwort: > I think it's very likely new users will not expect it to "just do it" in the current directory

You only know the command when you have read the usage text. And the usage text says `Initializes a registry in the current directory`. I will add a path parameter, but imho `vcpkg init-registry` should continue creating the registry in the cwd, because you have read the usage and decided against passing a path (and then it has the same behavior as `git init` )

ras0219-msft: Unfortunately, users often don't read docs (or worse -- they skim docs and copy paste commands with wild abandon).

autoantwort:
I have now finally implemented your suggestions.

autoantwort:
> For `x-add-version` I see two possibilities:
> 
>    1. Make `x-add-version` a `BasicCommand` and determine the `ports` and `versions` dir inside of `perform_and_exit`, but this is maybe a bad idea because you also want to check `--x-builtin-ports-root` and `--x-builtin-registry-versions-dir` and you maybe also want to change other commands so that they word in registries (`vcpkg edit/create`)
> 
>    2. Add a new class `VcpkgBasePaths` that contains the `.git`, `ports` and the `versions` folder, but not the rest of `VcpkgPaths` and remove the paths of `VcpkgBasePaths` from `VcpkgPaths`. Add than a new command category `BasePathsCommand`.
> 
> 
> I think the second solution is cleaner, but I don't want to implement something and have to change everything later. So which version do you prefer (or maybe you have another idea how to solve this)?

Any opinion about this @strega-nil @ras0219-msft ?

autoantwort:
@strega-nil-ms ping 
It is now finished

src/vcpkg/commands.cpp
```diff
@@ -41,13 +42,16 @@ namespace vcpkg::Commands
     {
         static const Version::VersionCommand version{};
         static const Contact::ContactCommand contact{};
+        static const InitRegistry::InitRegistryCommand initRegistry{};
```
strega-nil-ms, line 45:
```suggestion
        static const InitRegistry::InitRegistryCommand init_registry{};
```

src/vcpkg/commands.cpp
```diff
@@ -41,13 +42,16 @@ namespace vcpkg::Commands
...
 #if defined(_WIN32)
         static const UploadMetrics::UploadMetricsCommand upload_metrics{};
 #endif // defined(_WIN32)
 
         static std::vector<PackageNameAndFunction<const BasicCommand*>> t = {
             {"version", &version},
             {"contact", &contact},
+            {"x-init-registry", &initRegistry},
```
strega-nil-ms, line 53:
```suggestion
            {"x-init-registry", &init_registry},
```

src/vcpkg/commands.init-registry.cpp
```diff
@@ -0,0 +1,48 @@
+#include <vcpkg/base/files.h>
+#include <vcpkg/base/system.print.h>
+
+#include <vcpkg/commands.init-registry.h>
+#include <vcpkg/vcpkgcmdarguments.h>
+
+namespace vcpkg::Commands::InitRegistry
+{
```
strega-nil-ms, line 8:
```suggestion
{
    static const CommandStructure COMMAND_STRUCTURE = {
        create_example_string(R"(x-init-registry .)"),
        1,
        1,
        {{}, {}, {}},
        nullptr,
    };

```

src/vcpkg/commands.init-registry.cpp
```diff
@@ -0,0 +1,48 @@
...
+        {
+            Checks::exit_with_message(
+                VCPKG_LINE_INFO,
+                "You have to provide a path to a directory where the registry should be created as argument. "
+                "Pass '.' to create the registry in the current directory.");
+        }
+        if (args.command_arguments.size() > 1)
+        {
+            Checks::exit_with_message(VCPKG_LINE_INFO, "Only pass one directory as argument to this command.");
+        }
```
strega-nil-ms, line 21:
```suggestion
        auto parsed_args = args.parse_arguments(COMMAND_STRUCTURE);

```

src/vcpkg/commands.init-registry.cpp
```diff
@@ -0,0 +1,48 @@
...
+        auto path = fs::u8path(args.command_arguments.front());
+        if (path.is_relative())
+        {
+            path = fs.current_path(VCPKG_LINE_INFO) / path;
+        }
```
strega-nil-ms, line 26:
```suggestion
        path = fs::combine(paths.original_cwd, path);
```

strega-nil-ms (changes requested):
minor changes

you'll also need to take a `VcpkgPaths&`.

strega-nil-ms (approved):
Thanks so much @autoantwort :)

Neumann-A:
So I used:
`./vcpkg.exe x-init-registry ../my-vcpkg-ports/`
and
`./vcpkg.exe x-add-version --all --overlay-ports=../my-vcpkg-ports/ --x-builtin-registry-versions-dir=../my-vcpkg-ports/versions/ --x-builtin-ports-root=../my-vcpkg-ports/`

What I observed: 
Having the ports directory being generated by `./vcpkg.exe x-init-registry ../my-vcpkg-ports/` is unnecessary.

PR #59, opened 2021-04-15
Attempt to fix crash

frivard-coveo:
Related to https://github.com/microsoft/vcpkg/issues/17269

With this change, I am unable to reproduce the issue.


strega-nil:
Closed in favor of #60

PR #63, opened 2021-04-16
Add SourceLink

frivard-coveo:
This is a possible solution for improvement[ #17325 ](https://github.com/microsoft/vcpkg/issues/17325)

Note that this is a build option. Although it "looks" enabled by default, the full condition is in fact combined with `VCPKG_EMBED_GIT_SHA`, which is OFF by default.
Therefore on official builds it would be ON.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 /cmake-build-*
 /.vs
 /CMakeSettings.json
+/out
 .DS_Store
 # Qt Creator CMake project files
 CMakeLists.txt.user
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -6,6 +6,7 @@
 /cmake-build-*
 /.vs
 /CMakeSettings.json
+/out
```
frivard-coveo, line 9: I added this because in Visual Studio 2019, the default output directory for CMake Projects is "out/build/{buildType}".

strega-nil (changes requested):

ras0219 (approved):

ras0219-msft:
Thanks for the PR!

PR #65, opened 2021-04-20
Add x-block-origin for Asset Caching

ras0219:
Follow up after #10


BillyONeal (commented):
I don't really understand what this is for but I don't see any obvious footguns in the code here.

Except there aren't any tests.

BillyONeal (approved):

PR #67, opened 2021-04-26
Output error code when getting VS ENV failed

JackBoosY:
When using `vcvarsall.bat` to obtain the Visual Studio environment variable fails, output the error message to notice user.

Fixes https://github.com/microsoft/vcpkg/issues/17319


ras0219-msft:
LGTM, thanks!

PR #69, opened 2021-05-03
Implement pseudo-lockfiles

ras0219-msft:
As part of stabilizing Registries, we must avoid network access on otherwise no-op tasks, such as running `vcpkg install` on an already restored directory.

This PR implements a pseudo-lockfile, stored in `$installed/vcpkg/vcpkg-lock.json` which will cache the "HEAD" ref used of each git registry. On future actions against that install tree, vcpkg will attempt to use the stale HEAD until a failure is met (such as needing a newer version). On failure, we will fetch and retry once. If version resolution is successful and any lock entries were invalidated, the file will be flushed to disk.

- [x] Migrate test asset `version-files` from Microsoft/vcpkg to this repo
- [x] E2E Tests


azure-pipelines/e2e_ports/version-files/default-baseline-1/vcpkg.json
```diff
@@ -0,0 +1,8 @@
+{
```
strega-nil, line 1: please format-manifest all these.

src/vcpkg/base/files.cpp
```diff
@@ -487,6 +487,16 @@ namespace vcpkg::Files
             Checks::exit_with_message(linfo, "error writing file: %s: %s", fs::u8string(path), ec.message());
         }
     }
+    void Filesystem::write_rename_contents(const fs::path& path,
+                                           const fs::path& tmpname,
```
strega-nil, line 491: Is there a reason that this takes a tmpname? Should this ever be anything but `"{path}.{pid}.tmp"`?

src/vcpkg/registries.cpp
```diff
@@ -1217,4 +1306,14 @@ namespace vcpkg
     {
         return try_parse_builtin_baseline(paths, "default");
     }
+
+    bool is_git_commit_sha(StringView sv)
+    {
+        static constexpr struct
+        {
+            bool operator()(char ch) { return ('0' <= ch && ch <= '9') || ('a' <= ch && ch <= 'f'); }
+        } is_lcase_ascii_hex;
```
strega-nil, line 1315: any reason not to use lambdas here?

```suggestion
        auto is_lcase_ascii_hex = [](char ch) { return ('0' <= ch && ch <= '9') || ('a' <= ch && ch <= 'f'); };
```

src/vcpkg/registries.cpp
```diff
@@ -491,101 +532,95 @@ namespace
...
-        return res;
+    GitRegistryEntry::GitRegistryEntry(const GitRegistry& reg, StringView name)
+        : port_name(name.to_string()), parent(reg)
+    {
+        // This guarantees that parent.m_paths has been filled before any entries are constructed.
+        // The entries may then rely on (bool)parent.m_paths during their operation
+        Checks::check_exit(VCPKG_LINE_INFO, parent.m_paths);
+#if defined(_MSC_VER)
+        __assume(parent.m_paths);
+#endif
```
strega-nil, line 547: why is this here. This is not performance sensitive code.

src/vcpkg/registries.cpp
```diff
@@ -491,101 +532,95 @@ namespace
...
-        auto res = std::make_unique<GitRegistryEntry>();
-        res->port_name = port_name.to_string();
-        for (auto&& version_entry : version_entries)
-        {
-            res->port_versions.push_back(version_entry.version);
-            res->git_trees.push_back(version_entry.git_tree);
-        }
-        return res;
+    GitRegistryEntry::GitRegistryEntry(const GitRegistry& reg, StringView name)
+        : port_name(name.to_string()), parent(reg)
```
strega-nil, line 540: Note to self: rip this `parent` bit out later.

src/vcpkg/registries.cpp
```diff
@@ -96,11 +100,48 @@ namespace
...
-        DelayedInit<std::string> m_commit; // TODO: eventually this should end up in the vcpkg-lock.json file
-        DelayedInit<fs::path> m_versions_tree;
         std::string m_baseline_identifier;
+        DelayedInit<LockFile::Entry> m_lock_entry;
+        mutable Optional<fs::path> m_stale_versions_tree;
+        DelayedInit<fs::path> m_versions_tree;
         DelayedInit<Baseline> m_baseline;
+
+        // Shared by all child GitRegistryEntry's
+        mutable const VcpkgPaths* m_paths = nullptr;
```
strega-nil, line 144: Note to self: rip this out later. (also please make comments TODO-ing that, since we should keep a record of what we need to do)

src/vcpkg/registries.cpp
```diff
@@ -96,11 +100,48 @@ namespace
             });
         }
 
+        struct VTPResult
```
strega-nil, line 103:
```suggestion
        struct VersionsTreePathResult
```
VTPResult is a really confusing name to me.

src/vcpkg/registries.cpp
```diff
@@ -1100,6 +1167,28 @@ namespace
 
 namespace vcpkg
 {
+    LockFile::Entry LockFile::get_or_fetch(const VcpkgPaths& paths, StringView key)
+    {
+        auto it = lockdata.find(key);
+        if (it == lockdata.end())
+        {
+            auto x = paths.git_fetch_from_remote_registry(key, "HEAD");
```
strega-nil, line 1175: Note to self: we should allow setting a ref in vcpkg-configuration.json

strega-nil (approved):
Approved with comments; please rename `VTPResult` and maybe remove the second parameter in `write_rename_contents`

src/vcpkg/base/files.cpp
ras0219-msft: I'm not confident that's always the correct naming, so I left the api generic.

src/vcpkg/registries.cpp
ras0219-msft: No particular reason, but it's equivalent.

src/vcpkg/registries.cpp
ras0219-msft: Otherwise it fails `/analyze` because the analyzer doesn't know that `Checks::check_exit()` guarantees the condition is true on return.

src/vcpkg/registries.cpp
ras0219-msft: In general, this will not be possible -- shared state is mandated by the actual desired execution (if one GitRegistryEntry forces a `git fetch`, those results should apply to all other Entries). Forcing that shared state to be passed in to every API call results in a completely untestable interface, since the API is type-erased and as such every caller has to pass the superset of all possible shared states (VcpkgPaths) in.

It might be possible to reduce the "scope" of the captured parent pointer, by capturing several pointers to smaller subobjects, but fundamentally there is shared state here between the GitRegistryEntry objects and calls upon one affect the behavior of the others.

cnSchwarzer:
Why `baseline` now a required field in Git Registry? (https://github.com/microsoft/vcpkg/issues/17726)

PR #86, opened 2021-06-04
vcpkg integrate bash: use ~/.bash_profile on macOS

autoantwort:
On mac the `.bash_profile` file is used instead of the `.bashrc` file by the Terminal


strega-nil:
Thanks @autoantwort (although I'm not sure how many people use bash on macOS anymore haha)

autoantwort:
> Thanks @autoantwort (although I'm not sure how many people use bash on macOS anymore haha)

Because of that I have created #87 ;D

PR #90, opened 2021-06-09
Support custom registries in commands

autoantwort:
I have started with `format-manifest` and `x-add-version`. I would add support for more commands, but before that it would be cool if you can say if the current design is okay or if I should change it.


strega-nil-ms:
Thanks @autoantwort ! This is great.

strega-nil-ms:
Reverting due to incompatibility with #88; could you reopen this PR @autoantwort ?

autoantwort:
I already wondered that it was merged so fast without a comment, didn't expected that :D 

Will open a new PR

PR #93, opened 2021-06-14
Support custom registries in commands 

autoantwort:
Same as #90, but it got reverted

### DIFF
--- a/include/vcpkg/vcpkgpaths.h
+++ b/include/vcpkg/vcpkgpaths.h
@@ -141,7 +141,12 @@ namespace vcpkg
         // Git manipulation in the vcpkg directory
         ExpectedS<std::string> get_current_git_sha() const;
         std::string get_current_git_sha_message() const;
-        ExpectedS<fs::path> git_checkout_baseline(StringView commit_sha) const;
+        ExpectedS<std::string> get_current_git_sha(const fs::path& dot_git_dir) const;
+        std::string get_current_git_sha_message(const fs::path& dot_git_dir) const;
+        ExpectedS<fs::path> git_checkout_builtin_baseline(StringView commit_sha) const;
+        ExpectedS<fs::path> git_checkout_baseline(const fs::path& dot_git_dir,
+                                                  const fs::path& versions_dir,
+                                                  StringView commit_sha) const;
         ExpectedS<fs::path> git_checkout_port(StringView port_name,
                                               StringView git_tree,
                                               const fs::path& dot_git_dir) const;
@@ -149,6 +154,8 @@ namespace vcpkg
 
         const Downloads::DownloadManager& get_download_manager() const;
 
+        ExpectedS<std::map<std::string, std::string, std::less<>>> git_get_port_treeish_map(
+            const fs::path& ports_dir) const;
         ExpectedS<std::map<std::string, std::string, std::less<>>> git_get_local_port_treeish_map() const;
 
         // Git manipulation for remote registries
--- a/include/vcpkg/vcpkgpaths.h
+++ b/include/vcpkg/vcpkgpaths.h
@@ -108,6 +108,11 @@ namespace vcpkg
         fs::path builtin_ports;
         fs::path builtin_registry_versions;
 
+        // When we are in a custom registry, these values are different from builtin_ports, otherwise they are the same
+        fs::path current_registry_dot_git_dir;
+        fs::path current_registry_ports;
+        fs::path current_registry_versions;
+
         fs::path tools;
         fs::path buildsystems;
         fs::path buildsystems_msbuild_targets;
--- a/src/vcpkg/vcpkgpaths.cpp
+++ b/src/vcpkg/vcpkgpaths.cpp
@@ -698,7 +698,15 @@ namespace vcpkg
 
     ExpectedS<std::string> VcpkgPaths::get_current_git_sha() const
     {
-        auto cmd = git_cmd_builder(this->root / fs::u8path(".git"), this->root);
+        return get_current_git_sha(this->root / fs::u8path(".git"));
+    }
+    std::string VcpkgPaths::get_current_git_sha_message() const
+    {
+        return get_current_git_sha_message(this->root / fs::u8path(".git"));
+    }
+    ExpectedS<std::string> VcpkgPaths::get_current_git_sha(const fs::path& dot_git_dir) const
+    {
+        auto cmd = git_cmd_builder(dot_git_dir, this->root);
         cmd.string_arg("rev-parse").string_arg("HEAD");
         auto output = System::cmd_execute_and_capture_output(cmd);
         if (output.exit_code != 0)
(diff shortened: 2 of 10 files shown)


src/vcpkg/vcpkgpaths.cpp
```diff
@@ -720,8 +808,9 @@ namespace vcpkg
         if (!fs.exists(destination))
         {
             const fs::path destination_tmp = destination_parent / fs::u8path("baseline.json.tmp");
-            auto treeish = Strings::concat(commit_sha, ":versions/baseline.json");
-            auto maybe_contents = git_show(treeish, this->root / fs::u8path(".git"));
+            auto relative_path = fs.relative(VCPKG_LINE_INFO, versions_dir, dot_git_dir.parent_path());
```
strega-nil-ms, line 811: the versions directory is _always_ at `/versions`.

strega-nil-ms (changes requested):
I'd like to see these changed first, then we can look at the resulting code and see what it looks like

include/vcpkg/vcpkgpaths.h
```diff
@@ -108,6 +108,11 @@ namespace vcpkg
         fs::path builtin_ports;
         fs::path builtin_registry_versions;
 
+        // When we are in a custom registry, these values are different from builtin_ports, otherwise they are the same
+        fs::path current_registry_dot_git_dir;
+        fs::path current_registry_ports;
+        fs::path current_registry_versions;
```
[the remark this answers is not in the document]
autoantwort: Yeah that is true for user registry, but then there are the pipeline tests that set the versions root to `$VersionFilesRoot/versions_incomplete` ...

src/vcpkg/vcpkgpaths.cpp
autoantwort: No, maybe you have:
root
- .git 
- my_registry
   - ports
   - versions

Or do we not support this, so the registry must be at the root of a git repo?

strega-nil-ms: We do not support this.

include/vcpkg/vcpkgpaths.h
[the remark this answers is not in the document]
strega-nil-ms: honestly, I don't care about supporting that for anything but builtin registries.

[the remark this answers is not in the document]
autoantwort: Yeah but x-add-version does not know anything about the buildin registry, it only knows that there is registry at the given folders (`current_registry_...`). So the options are rewriting the tests or add a lot of `if (is_builtin_registr()) ... else ...` into `x-add-version` and probably `format-manifests` (I thing there the ports dir is also set for tests)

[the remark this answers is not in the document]
strega-nil-ms: mhh, best option is rewriting tests I think.

strega-nil-ms:
Depends on #105

azure-pipelines/end-to-end-tests-dir/versions.ps1
```diff
@@ -2,18 +2,34 @@
...
+Throw-IfFailed
+
 # Test verify versions
-mkdir $VersionFilesRoot | Out-Null
-Copy-Item -Recurse "$versionFilesPath/versions_incomplete" $VersionFilesRoot
+Copy-Item -Recurse "$versionFilesPath/versions" "$RegistryRootComplete/versions"
+Copy-Item -Recurse "$versionFilesPath/ports" "$RegistryRootComplete/ports"
+& $GitExe -C "$RegistryRootComplete" add -A 
+Throw-IfFailed
+& $GitExe -C "$RegistryRootComplete" -c user.name='vcpkg-tool' -c user.email='ci@vcpkg.io' commit -m "add files"
```
strega-nil-ms, line 18: is there any reason to do this rather than `git -C ...`?

include/vcpkg/vcpkgpaths.h
```diff
@@ -175,7 +182,12 @@ namespace vcpkg
 
         // the directory of the builtin ports
         // this should be used only for helper commands, not core commands like `install`.
-        path builtin_ports_directory() const { return this->builtin_ports; }
+        path builtin_ports_directory() const { return this->root / u8path("ports"); }
+        path builtin_versions_directory() const { return this->root / u8path("versions"); }
+
+        path current_git_directory() const { return current_registry_root / u8path(".git"); }
+        path current_ports_directory() const { return current_registry_root / u8path("ports"); }
+        path current_versions_directory() const { return current_registry_root / u8path("versions"); }
```
strega-nil-ms, line 190:
```suggestion
        path current_registry_dot_git_dir() const { return current_registry_root / u8path(".git"); }
        path current_registry_ports_dir() const { return current_registry_root / u8path("ports"); }
        path current_registry_versions_dir() const { return current_registry_root / u8path("versions"); }
```

include/vcpkg/vcpkgpaths.h
```diff
@@ -175,7 +182,12 @@ namespace vcpkg
 
         // the directory of the builtin ports
         // this should be used only for helper commands, not core commands like `install`.
-        path builtin_ports_directory() const { return this->builtin_ports; }
+        path builtin_ports_directory() const { return this->root / u8path("ports"); }
+        path builtin_versions_directory() const { return this->root / u8path("versions"); }
```
strega-nil-ms, line 186:
```suggestion
        path builtin_registry_ports_dir() const { return this->root / u8path("ports"); }
        path builtin_registry_versions_dir() const { return this->root / u8path("versions"); }
```

renaming these is good, imo.

src/vcpkg-test/arguments.cpp
```diff
@@ -27,8 +26,7 @@ TEST_CASE ("VcpkgCmdArguments from lowercase argument sequence", "[arguments]")
 
     REQUIRE(*v.vcpkg_root_dir == "C:\\vcpkg");
     REQUIRE(*v.scripts_root_dir == "C:\\scripts");
-    REQUIRE(*v.builtin_ports_root_dir == "C:\\ports");
-    REQUIRE(*v.builtin_registry_versions_dir == "C:\\versions");
+    REQUIRE(*v.registry_root_dir == "C:\\registry");
```
strega-nil-ms, line 29: There should be a test for "what is `*v.registry_root_dir` when `--x-registry-root` isn't passed"

src/vcpkg/vcpkgpaths.cpp
```diff
@@ -445,6 +441,41 @@ namespace vcpkg
...
+                current_registry_root = this->root;
+            }
+        }
+        Debug::print("Using current_registry_root: ", u8string(current_registry_root), '\n');
+        for (auto& dir_name : {".git", "ports", "versions"})
+        {
+            Checks::check_exit(VCPKG_LINE_INFO,
+                               !(current_registry_root / u8path(dir_name)).empty(),
+                               "Error: There is no " + std::string(dir_name) + " dir for current registry at " +
+                                   u8string(current_registry_root));
```
strega-nil-ms, line 477:
```suggestion
            Checks::check_exit(VCPKG_LINE_INFO,
                               !get_filesystem().exists(VCPKG_LINE_INFO, current_registry_root / u8path(dir_name)),
                               "Error: There is no %s directory for current registry at %s",
                               dir_name,
                               u8string(current_registry_root));
```

azure-pipelines/end-to-end-tests-dir/versions.ps1
autoantwort: I thought the `exe` extension is needed for windows, but it seems to be that this is not the case.

theblackunknown:
As a maintainer of a custom registry this PR definitely looks interesting !
Is there anything I can do to help move it forward ?

autoantwort:
First https://github.com/microsoft/vcpkg/issues/18967 must be "resolved" and then it can be implemented

theblackunknown:
I understand, I'll follow the discussion there to keep an eye on this then.

PR #95, opened 2021-06-17
binarycaching: Add NuGet timeout configuration entry

ekilmer:
Related to this issue https://github.com/microsoft/vcpkg/issues/18483

As mentioned in the issue, ideally this timeout value could be configurable, but I'm not sure the best way to approach that in this code base. Help would be greatly appreciated!

Needs the extra second due to some potential issue with how NuGet calculates the timeout:
https://jfrog.com/knowledge-base/how-to-resolve-nuget-push-failures-after-5-minutes-even-though-timeout-value-is-set-as-greater-than-5-minutes-300-seconds/

### DIFF
--- a/src/vcpkg-test/configparser.cpp
+++ b/src/vcpkg-test/configparser.cpp
@@ -104,6 +104,42 @@ TEST_CASE ("BinaryConfigParser nuget source provider", "[binaryconfigparser]")
     }
 }
 
+TEST_CASE ("BinaryConfigParser nuget timeout", "[binaryconfigparser]")
+{
+    {
+        auto parsed = create_binary_provider_from_configs_pure("nugettimeout,3601", {});
+        REQUIRE(parsed.has_value());
+    }
+    {
+        auto parsed = create_binary_provider_from_configs_pure("nugettimeout", {});
+        REQUIRE(!parsed.has_value());
+    }
+    {
+        auto parsed = create_binary_provider_from_configs_pure("nugettimeout,", {});
+        REQUIRE(!parsed.has_value());
+    }
+    {
+        auto parsed = create_binary_provider_from_configs_pure("nugettimeout,nonsense", {});
+        REQUIRE(!parsed.has_value());
+    }
+    {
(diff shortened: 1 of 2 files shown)


ras0219-msft:
Thanks for the PR!

I think we will eventually want this setting to come through the binary caching configuration string. To preserve our ability to fuse multiple NuGet calls together, I think this means it will need to come from a separate entry:
```
nuget,<url1>;nuget,<url2>;nugettimeout,3601
```

ekilmer:
> we will eventually want this setting to come through the binary caching configuration string

Just to confirm: You are referencing the [valid source strings](https://github.com/microsoft/vcpkg/blob/08214e18eeb146542944067c35d77faca8962be5/docs/users/binarycaching.md) section and `VCPKG_BINARY_SOURCES` environment variable/`--binarysource` CLI option in the documentation on binary caching, and adding a new keyword `nugettimeout`?

If so, that makes sense to me! Thank you.

ekilmer:
@ras0219-msft I've tried implementing a solution to what you described. I'd appreciate a review when you get a chance. Thank you!

ras0219-msft (approved):
This looks really good to me, thanks for making this PR!

Could you also add the new documentation to `docs/users/binarycaching.md`?

ekilmer:
> Could you also add the new documentation to `docs/users/binarycaching.md`?

I'd be happy to! Just one question, how does that work? It looks like that file lives in the main [vcpkg repo](https://github.com/microsoft/vcpkg/blob/master/docs/users/binarycaching.md), and a PR to update the doc should also bump the vcpkg-tool to a commit that has this PR merged. But a bump to the vcpkg-tool version look like it requires a release (so that Windows users can download a pre-built executable). 

Am I missing something?

--------------

It also looks like there is now a conflict with `binarycaching.cpp` file. I will try to fix that today at some point.

ras0219-msft:
Ah, you're absolutely right. Sorry for my confusion! Still getting used to the split repos I suppose :)

Thanks again!

PR #100, opened 2021-06-25
vcpkg search: Users should use git pull to get the latest results

autoantwort:
See:
- microsoft/vcpkg#18648
- microsoft/vcpkg#18576 
- microsoft/vcpkg#18280
- microsoft/vcpkg#18165
- ...

The users don't know that the list is probably outdated and must be updated with `git pull`. So explain that.

### DIFF
--- a/src/vcpkg/commands.search.cpp
+++ b/src/vcpkg/commands.search.cpp
@@ -181,6 +181,7 @@ namespace vcpkg::Commands::Search
         if (!enable_json)
         {
             System::print2(
+                "The search result is maybe outdated. Execute `git pull` to get the latest results.\n"
                 "\nIf your library is not listed, please open an issue at and/or consider making a pull request:\n"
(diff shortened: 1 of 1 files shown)


strega-nil-ms:
Yeah this is probably good.

strega-nil-ms (approved):

src/vcpkg/commands.search.cpp
```diff
@@ -181,6 +181,7 @@ namespace vcpkg::Commands::Search
         if (!enable_json)
         {
             System::print2(
+                "The search result is maybe outdated. Execute `git pull` to get the latest results.\n"
```
strega-nil-ms, line 184:
```suggestion
                "The search result may be outdated. Run `git pull` to get the latest results.\n"
```

strega-nil-ms: also, this should probably only be printed when one isn't using registries?

PR #103, opened 2021-07-01
Add serialization of configuration and RegistryImplementation

mathisloge:
The serialization of the vcpkg-configuration.json and RegistryImplementation is missing. 

I don't know if this is in the scope of vcpkglib but I don't see any other way to generate the vcpkg-configuration.json file.

it would be good if someone could check this and guide me your coding guidelines / style a bit.

### DIFF
--- a/include/vcpkg/configuration.h
+++ b/include/vcpkg/configuration.h
@@ -21,4 +21,5 @@ namespace vcpkg
     };
 
     std::unique_ptr<Json::IDeserializer<Configuration>> make_configuration_deserializer(const path& config_directory);
+    Json::Object serialize_configuration(const Configuration& config);
 }
--- a/include/vcpkg/registries.h
+++ b/include/vcpkg/registries.h
@@ -67,6 +67,8 @@ namespace vcpkg
 
         virtual Optional<VersionT> get_baseline_version(const VcpkgPaths& paths, StringView port_name) const = 0;
 
+        virtual Json::Object serialize() const;
+
         virtual ~RegistryImplementation() = default;
     };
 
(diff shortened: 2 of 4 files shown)


mathisloge:
And can we move the subclasses of `RegistryImplementation` into the header to enable instantiation of the different RegistryImplementations ?

strega-nil-ms:
This looks great, but is there a reason you're adding it? Do you need it for something?

mathisloge:
yeah. 
I'm developing a small tool to better manage manifests with multiple registries and all the fancy stuff.
Basically I need a way to generate the vcpkg-configuration.json and I also need access to the RegistryImplementation specialisations.

strega-nil-ms:
huh, so you're trying to use vcpkglib as an actual library?
neat!

mathisloge:
yes. I don't care so much about a stable API. Doing it only with commands might be a bit hard and sometimes not really feasable or involves some manual vcpkg -> strings -> my own data flow which is more work than to adapt to a changing api. 


(just changed visibility to public. will make it private later again. (only be public if it reaches some point of functionality) 
https://github.com/mathisloge/vcpkg-gui

mathisloge:
just as a thought experiment: 

would it be possible to for example split some implementation detail e.g. registries.h/.cpp into a structure like this: 
include/registries.h (public interface)
include/impl/registries/gitregistry.h (also public, but in a special impl folder that makes it clear that these are implementation details that are nevertheless publicly available.)
include/impl/registries/filesystemregistry.h
etc.

strega-nil-ms:
It's like, possible, but we don't consider `vcpkglib` as a publicly consumable library. We don't want to, at this point, support it. The best thing to do is to communicate with `vcpkg` as an executable, imo.

mathisloge:
okay.

PR #104, opened 2021-07-02
Initial attempt at showing usage information in manifest mode

mrexodia:
I switched to using manifest mode with vcpkg, but I'm missing the functionality of showing usage information after running vcpkg. I tried to add this back in, but I couldn't manage to get the usage information for the packages already installed. It doesn't seem to be possible to get the `BinaryParagraph` when that package isn't installed, could you help me with this?

Also is this the right place to ask for feedback or should I move to the vcpkg issue tracker?


strega-nil-ms:
This should probably use #82 instead?

mrexodia:
Yep, agreed! I'll wait until that is merged then.

PR #105, opened 2021-07-02
init-registry: registries can only be created at the root of a git repository

autoantwort:
As a response to [this](https://github.com/microsoft/vcpkg-tool/pull/93#discussion_r658918215)


strega-nil-ms:
Please format the code, then approved :)

strega-nil-ms (approved):

autoantwort:
> Please format the code, then approved :)

The code is properly formatted, but I have to escape a `path` by `u8string` when I want to print the path on windows :)

strega-nil-ms (approved):

PR #107, opened 2021-07-07
Fixed issue with output from CMake 3.21.0

MatthewPowley:
When calling CMake to obtain triplet information a regression change from CMake 3.20.5 to 3.21.0-rc1
causes the tool to exit due to an exception.

This is due to the response containing a mixture of '\n' and  '\r\n' line endings.

After splitting the output on line feeds ('\n') each line is then trimmed of whitespace to remove
any carriage returns ('\r') or trailing whitespace. This could be fixed in CMake but this
change makes it robust.

This commit in CMake caused the fault: https://github.com/Kitware/CMake/commit/0a0a0f8a744e2bfa35cdbc90db6e4e23adadd59b
It is unlikely to change as they have added a new helper file for writing the terminal output.

### DIFF
--- a/src/vcpkg/cmakevars.cpp
+++ b/src/vcpkg/cmakevars.cpp
@@ -183,7 +183,8 @@ endmacro()
         const auto ec_data = cmd_execute_and_capture_output(cmd_launch_cmake);
         Checks::check_exit(VCPKG_LINE_INFO, ec_data.exit_code == 0, ec_data.output);
 
-        const std::vector<std::string> lines = Strings::split(ec_data.output, '\n');
+        std::vector<std::string> lines = Strings::split(ec_data.output, '\n');
+        Strings::trim_all_and_remove_whitespace_strings(&lines);
 
         const auto end = lines.cend();
 
--- a/src/vcpkg/cmakevars.cpp
+++ b/src/vcpkg/cmakevars.cpp
@@ -180,20 +180,27 @@ endmacro()
         static constexpr CStringView BLOCK_END_GUID = "e1e74b5c-18cb-4474-a6bd-5c1c8bc81f3f";
 
         const auto cmd_launch_cmake = vcpkg::make_cmake_cmd(paths, script_path, {});
-        const auto ec_data = cmd_execute_and_capture_output(cmd_launch_cmake);
-        Checks::check_exit(VCPKG_LINE_INFO, ec_data.exit_code == 0, ec_data.output);
 
-        std::vector<std::string> lines = Strings::split(ec_data.output, '\n');
-        Strings::trim_all_and_remove_whitespace_strings(&lines);
+        std::vector<std::string> lines;
+        auto const exit_code = cmd_execute_and_stream_lines(
+            cmd_launch_cmake, [&](StringView sv) { lines.emplace_back(sv.begin(), sv.end()); });
+
+        Checks::check_exit(VCPKG_LINE_INFO, exit_code == 0, exit_code == 0 ? "" : Strings::join("\n", lines));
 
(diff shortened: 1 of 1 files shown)


MatthewPowley:
Fixes these issues reported in Vcpkg.
https://github.com/microsoft/vcpkg/issues/18718
https://github.com/microsoft/vcpkg/issues/18804
https://github.com/microsoft/vcpkg/issues/18806

src/vcpkg/cmakevars.cpp
```diff
@@ -183,7 +183,8 @@ endmacro()
         const auto ec_data = cmd_execute_and_capture_output(cmd_launch_cmake);
         Checks::check_exit(VCPKG_LINE_INFO, ec_data.exit_code == 0, ec_data.output);
 
-        const std::vector<std::string> lines = Strings::split(ec_data.output, '\n');
+        std::vector<std::string> lines = Strings::split(ec_data.output, '\n');
+        Strings::trim_all_and_remove_whitespace_strings(&lines);
```
ras0219-msft, line 187: This seems a bit aggressive compared to instead just trimming a single trailing `\r`.

```suggestion
        for (auto&& line : lines) {
            if (!line.empty() && line.back() == '\r') line.pop_back();
        }
```

ras0219-msft (commented):
We call CMake from several different locations; we need to audit each of these for potential `\r\n` mishandling.

MatthewPowley:
I have had a look through the places where CMake is called and apart from this instance they either use 'cmd_execute_and_stream_lines', which your pull request #108 handles, or they are written directly to the console.

I agree the above is aggressive to remove trailing '\r' characters. (I made the change quickly as I was proving out a large system build using Visual Studio 2022 and I came across this blocking issue).

The code could be consistent with other CMake calls in the system by also using 'cmd_execute_and_stream_lines'.

```
        std::vector<std::string> lines;
        auto const exit_code = cmd_execute_and_stream_lines(
            cmd_launch_cmake, [&](StringView sv) { lines.emplace_back(sv.begin(), sv.end()); });

        Checks::check_exit(VCPKG_LINE_INFO, exit_code == 0, Strings::join("\n", lines));

        const auto end = lines.cend();

        auto port_start = std::find(lines.cbegin(), end, PORT_START_GUID);
        ...
```
The only 'nasty' bit I don't like is having to join the lines again in case of an error.

What is your preference?

ras0219-msft:
Yeah I like your suggested unification 👍 . The idea of the lines join is fine, though unfortunately the code as written will execute the join on all cases (including non-error), so it'd be better to do:
```cpp
if (exit_code != 0)
{
    Checks::exit_with_message(VCPKG_LINE_INFO, Strings::join("\n", lines));
}
```

strega-nil-ms:
Fixed by #107, I believe?

ras0219-msft (approved):

strega-nil-ms:
Oh, cool, thanks @MatthewPowley :)

cenit:
a release with this PR is very urgent to be deployed in vcpkg because cmake 3.21 is now released and many CI tests using :latest cmake version are now failing :)

bradking:
@MatthewPowley thanks for trying the 3.21 release candidates and tracking this down. 

> This is due to the response containing a mixture of '\n' and '\r\n' line endings.

I think it is good to tolerate both newline styles in vcpkg.

> It is unlikely to change as they have added a new helper file for writing the terminal output.

The newline change was not intentional.  I've opened [CMake Issue 22444](https://gitlab.kitware.com/cmake/cmake/-/issues/22444) to discuss this on the CMake side.

PR #111, opened 2021-07-10
integrate project: add option to set the nuget package name

autoantwort:
Currently the name is based on the path of the current vcpkg root, but if you want to commit
the project integration, you have the problem that the package name is different on every
developer machine. So add an option to set the package name.

### DIFF
--- a/src/vcpkg/commands.integrate.cpp
+++ b/src/vcpkg/commands.integrate.cpp
@@ -351,7 +351,7 @@ CMake projects should use: "-DCMAKE_TOOLCHAIN_FILE=%s"
     }
 
 #if defined(WIN32)
-    static void integrate_project(const VcpkgPaths& paths)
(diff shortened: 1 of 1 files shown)


ras0219-msft:
The package contents hardcode the full path to the vcpkg instance on the machine, so `vcpkg integrate project` is not able to be used cross-machine regardless of the package id. For users that want behavior like `vcpkg integrate project` cross machine, they should instead use a submodule and directly pull in the props/targets.

autoantwort:
Hm I see. We actually use vcpkg as a submodule. Maybe vcpkg can simply provide a property sheet that can be added to a project. For me that actually sounds cleaner that the nuget way.

ras0219-msft:
We do!

`scripts\buildsystems\msbuild\vcpkg.props` and `scripts\buildsystems\msbuild\vcpkg.targets`

PR #127, opened 2021-07-23
Show error instead of success if the package to remove is already not installed

rouisibra:
Just a small change, The output should be in red.

### DIFF
--- a/src/vcpkg/remove.cpp
+++ b/src/vcpkg/remove.cpp
@@ -163,7 +163,7 @@ namespace vcpkg::Remove
         switch (action.plan_type)
         {
             case RemovePlanType::NOT_INSTALLED:
-                vcpkg::printf(Color::success, "Package %s is not installed\n", display_name);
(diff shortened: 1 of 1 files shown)


ras0219-msft:
Thanks for the PR!

We purposely show a green success here because we view the vcpkg commands as "desired state" and not as "state change". For example, `vcpkg install x` where `x` is already installed is also a "success" because after the command completes, `x` is available. Thus, `vcpkg remove x` where `x` is not installed is a success because after the command completes, `x` is no longer installed.

rouisibra:
> Thanks for the PR!
> 
> We purposely show a green success here because we view the vcpkg commands as "desired state" and not as "state change". For example, `vcpkg install x` where `x` is already installed is also a "success" because after the command completes, `x` is available. Thus, `vcpkg remove x` where `x` is not installed is a success because after the command completes, `x` is no longer installed.

I get it now, Thank you for explaining.

PR #128, opened 2021-07-23
fix check for ignore_lock_failures

strega-nil-ms:
cc @ras0219

### DIFF
--- a/src/vcpkg/vcpkgpaths.cpp
+++ b/src/vcpkg/vcpkgpaths.cpp
@@ -345,7 +345,9 @@ namespace vcpkg
 
             if (ec)
             {
-                if (ec == std::errc::device_or_resource_busy || args.ignore_lock_failures.value_or(false))
+                bool is_already_locked = ec == std::errc::device_or_resource_busy;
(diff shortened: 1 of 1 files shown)


src/vcpkg/vcpkgpaths.cpp
```diff
@@ -345,7 +345,9 @@ namespace vcpkg
 
             if (ec)
             {
-                if (ec == std::errc::device_or_resource_busy || args.ignore_lock_failures.value_or(false))
+                bool is_already_locked = ec == std::errc::device_or_resource_busy;
+                bool allow_errors = args.ignore_lock_failures.value_or(false);
+                if (is_already_locked || !allow_errors)
```
ras0219-msft, line 350: So the intention here is that `ignore_lock_failures` only applies to errors other than `std::errc::device_or_resource_busy`? And there's no way to completely suppress all failures to lock?

strega-nil: right - `std::errc::device_or_resource_busy` means "already locked". We do not want to suppress "already locked", because that would be a terrible idea.

ras0219-msft (approved):

PR #133, opened 2021-07-25
Display usage information after installing in manifest mode

ras0219-msft:
This PR adds tracking of top-level dependencies versus transitive dependencies through the versioned install planner. Then, after completing the "set-installed" algorithm, we print usage information for all top-level package specs.

Before:
```
PS C:\src\vcpkg\mtest> vcpkg install

Total elapsed time: 3.2 us
```
After:
```
PS C:\src\vcpkg\mtest> vcpkg install

Total elapsed time: 3.2 us

The package cpprestsdk:x86-windows provides CMake targets:

    find_package(cpprestsdk CONFIG REQUIRED)
    target_link_libraries(main PRIVATE cpprestsdk::cpprest cpprestsdk::cpprestsdk_zlib_internal cpprestsdk::cpprestsdk_brotli_internal)
```

Note that in the example, there are three installed packages (`cpprestsdk`, `zlib`, `brotli`) but usage information is only shown for the top level dependency `cpprestsdk`.

### DIFF
--- a/include/vcpkg/install.h
+++ b/include/vcpkg/install.h
@@ -102,6 +102,7 @@ namespace vcpkg::Install
     };
 
     CMakeUsageInfo get_cmake_usage(const BinaryParagraph& bpgh, const VcpkgPaths& paths);
+    void print_usage_information(const BinaryParagraph& bpgh, const VcpkgPaths& paths);
 
     extern const CommandStructure COMMAND_STRUCTURE;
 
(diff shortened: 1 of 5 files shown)


ras0219-msft:
This PR is related to https://github.com/microsoft/vcpkg-tool/pull/82, however it solves the problem of "where's the install tree" by incorporating the display of usage information into the install command.

PR #139, opened 2021-07-28
change the vcpkglib library to an object library

strega-nil-ms:
this is in line with that vcpkglib is _not_ a thing we want to ship

See the comments on PR #132

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,7 +114,7 @@ endif()
 
 # === Target: vcpkglib ===
 
-add_library(vcpkglib
+add_library(vcpkglib OBJECT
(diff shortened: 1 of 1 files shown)


Neumann-A:
this is not enough you also need to use `$<TARGET_OBJECTS:objlib>` in the targets consuming it. And instead of linking the object library you add it to the sources.

ras0219-msft:
We've been informed that `target_link_libraries()` is now supported for `OBJECT` libraries since a cmake 3.12: https://cmake.org/cmake/help/latest/command/target_link_libraries.html#linking-object-libraries

ras0219-msft (approved):

PR #140, opened 2021-07-28
Split instance locking into 4 output locks

ras0219-msft:
This PR contains one primary change and four drive-bys:
1. We now lock `packages`, `buildtrees`, `installed`, and `downloads` instead of a single instance lock. This enables "read-only" vcpkg instances by redirecting all output folders.
2. I've added an `--x-lock-instance` workaround in case this is insufficient; users can pass this to reenable the previous behavior (even in classic mode!) while waiting for a tool update that fixes the bug.
3. Push the cmake version required to 3.15 to support `CMAKE_MSVC_RUNTIME_LIBRARY` on the configure cli
4. Fix propagation of disable metrics to actually propagate the bool instead of assuming true
5. Add explicit guards against performing file locks inside a recursive vcpkg invocation.

I've tested this locally by doing:
```
$ vcpkg install --debug >log1 & vcpkg install --debug >log2
```

which results in log2 containing the expected extra line:
> [DEBUG] Waiting to take filesystem lock on /workspaces/vcpkg/downloads/.lock...

Note that since we're now using multiple locks, lock ordering is important. Future tool updates must either preserve the lock order or implement a "safe multiple locking" algorithm.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.15)
 
 # ===============
 # === Options ===
--- a/include/vcpkg/vcpkgcmdarguments.h
+++ b/include/vcpkg/vcpkgcmdarguments.h
@@ -178,6 +178,9 @@ namespace vcpkg
         constexpr static StringLiteral IGNORE_LOCK_FAILURES_ENV = "X_VCPKG_IGNORE_LOCK_FAILURES";
         Optional<bool> ignore_lock_failures = nullopt;
 
+        constexpr static StringLiteral LOCK_INSTANCE_SWITCH = "x-lock-instance";
+        Optional<bool> lock_instance = nullopt;
+
         constexpr static StringLiteral JSON_SWITCH = "x-json";
         Optional<bool> json = nullopt;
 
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.15)
```
strega-nil-ms, line 1: ?

strega-nil-ms (commented):
I still disagree strongly with this change...

CMakeLists.txt
ras0219-msft: > Push the cmake version required to 3.15 to support CMAKE_MSVC_RUNTIME_LIBRARY on the configure cli

strega-nil-ms: fair enough I guess.

strega-nil-ms (approved):

ras0219-msft:
Unfortunately, this implementation has a critical flaw that the four folders must exist for the locks to be taken. The correct improvement (which avoids creating the folders for read-only commands like "help") requires splitting out parts of VcpkgPaths to be deferred to the commands that require them.

That refactoring is complex because:
1. Big Ball of Mud Syndrome
2. Lock ordering (we can't acquire the resources in a random order because we would deadlock with a concurrent vcpkg that acquires them in a different order) - I plan to solve this by simply having a hierarchy of locks and gratuitously taking all locks before the desired one:
  i. Downloads (eventually to be removed)
  ii. Installed
  iii. Buildtrees
  iv. Packages

ras0219-msft:
Closing to be re-developed after teasing apart VcpkgPaths further.

PR #149, opened 2021-08-06
[commands.format-manifest.cpp] Update tags for port and convert manifest of all but ones with open PRs (from GitHub API)

SamuelMarks:
https://github.com/microsoft/vcpkg/issues/19378

(WiP)

### DIFF
--- a/src/vcpkg/commands.format-manifest.cpp
+++ b/src/vcpkg/commands.format-manifest.cpp
@@ -169,11 +169,75 @@ Please open an issue at https://github.com/microsoft/vcpkg, with the following o
             }
         }
     }
+
+    /* https://docs.github.com/en/rest/reference/pulls#list-pull-requests + .title + substr to get just with brackets */
+    std::vector<std::string> ports_with_open_prs(const char *repo="microsoft/vcpkg")
+    {
+        std::vector<std::string> ports;
+
+        /* TODO: Equivalent to:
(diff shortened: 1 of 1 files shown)


autoantwort:
I think this can be closed

SamuelMarks:
@autoantwort It's a draft, and is waiting on [#19436](https://github.com/microsoft/vcpkg/issues/19436)… I have updated my [libacquire](https://github.com/offscale/libacquire) project and am maintaining it pretty often. So if the lack of HTTPS client in vcpkg/vcpkg-tool is because of an outside dependency, then use libacquire to bring in your OS dependency.

autoantwort:
I thought you want a HTTPS client because of microsoft/vcpkg#19378, but since microsoft/vcpkg#19378 is done now, the need for a HTTPS client is not given anymore.

autoantwort:
Also the code of this pr has nothing to do with a https client ... Its all about preserving ports in format-manifest from being converted to vcpkg.json, but there is no port anymore that uses the CONTROL file format. So this PR is imho obsolete

SamuelMarks:
@autoantwort Yes this PR is obsolete, because of lack of activity on the issue requesting an HTTP client so that this could all be done natively (and not reverting to PowerShell). But the fact that the PR to change all the CONTROL files to json came more than a month after I requested—and offered—an HTTP client makes me a little annoyed.

PR #151, opened 2021-08-10
Fix versioning features of transitive manifests

ras0219-msft:
This PR fixes two outstanding bugs with using versioning features in transitive manifests:

1. We previously required a `builtin-baseline` as part of transitive manifests if they provide `version>=`. This was a bug; we only need `builtin-baseline` for the top-level manifest because it is a non-transitive property.
2. Manifest mode users that aren't using versioning or registries encountered errors when trying to use transitive dependencies from the built-in registry that have version constraints. This was not desired, because manifest mode without versioning should act the same as classic mode. In this PR, I've implemented that behavior by appending the builtin ports directory as an overlay. This is likely incorrect with respect to classic mode registry configuration (adding a vcpkg-configuration.json into the vcpkg instance), however that mode has no testing coverage currently anyway.

## Example setup:
**ports/overlay/vcpkg.json**
```json
{
    "name": "overlay",
    "version": "0",
    "dependencies": [
        {
            "name": "overlay-zlib",
            "version>=": "99999"
        }
    ]
}
```
**ports/overlay-zlib/vcpkg.json**
```json
{
    "name": "overlay-zlib",
    "version": "1"
}
```

### Example error 1:
```
$ cd /workspaces/vcpkg-tool/azure-pipelines/e2e_ports/version-files/transitive-constraints
$ /workspaces/vcpkg/vcpkg install --dry-run --x-builtin-ports-root=ports --x-install-root=vb
Parsing manifest failed: /workspaces/vcpkg-tool/azure-pipelines/e2e_ports/version-files/transitive-constraints/ports/overlay was rejected because it uses "version>=" without setting a "builtin-baseline".
```
### Example error 2 (after fixing the missing baseline):
```
$ /workspaces/vcpkg/vcpkg install --dry-run --x-builtin-ports-root=ports --x-install-root=vb
Error: no version entry for overlay-zlib at version 99999.
We are currently using the version in the ports tree (1).
```
The above example happens whenever the `version>=` disagrees with the current ports tree version.

### Expected output:
```
$ vcpkg install --dry-run --x-builtin-ports-root=ports --x-install-root=vb
Detecting compiler hash for triplet x64-linux...
    overlay[core]:x64-linux -> 0
  * overlay-zlib[core]:x64-linux -> 1
Additional packages (*) will be modified to complete this operation.
```

### DIFF
--- a/azure-pipelines/e2e_ports/version-files/transitive-constraints/ports/overlay-zlib/vcpkg.json
+++ b/azure-pipelines/e2e_ports/version-files/transitive-constraints/ports/overlay-zlib/vcpkg.json
@@ -1,0 +1,4 @@
+{
+    "name": "overlay-zlib",
+    "version": "1"
+}
--- a/azure-pipelines/e2e_ports/version-files/transitive-constraints/ports/overlay-zlib/vcpkg.json
+++ b/azure-pipelines/e2e_ports/version-files/transitive-constraints/ports/overlay-zlib/vcpkg.json
@@ -1,4 +1,4 @@
 {
-    "name": "overlay-zlib",
-    "version": "1"
+  "name": "overlay-zlib",
+  "version": "1"
 }
--- a/azure-pipelines/e2e_ports/version-files/transitive-constraints/ports/overlay/vcpkg.json
+++ b/azure-pipelines/e2e_ports/version-files/transitive-constraints/ports/overlay/vcpkg.json
@@ -1,0 +1,10 @@
+{
+    "name": "overlay",
+    "version": "0",
+    "dependencies": [
+        {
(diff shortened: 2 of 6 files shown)


azure-pipelines/e2e_ports/version-files/transitive-constraints/ports/overlay-zlib/vcpkg.json
```diff
@@ -0,0 +1,4 @@
+{
+    "name": "overlay-zlib",
+    "version": "1"
+}
```
strega-nil-ms, line 4: Could you format these files with `format-manifest`?

strega-nil-ms (approved):

PR #161, opened 2021-08-14
 Implement parallel cache decompression

ras0219-msft:
This PR primarily introduces parallel decompression of binary cache archives, however it also addresses a few co-located issues:
1. Reduce verbosity of binary cache restore by summarizing total number of packages
2. Remove `... done` messages for `Installing <x>` and `Building <x>`
3. Remove `Building <x>` message if restored from cache
4. `vcpkg build` no longer attempts to do a cache restore (by virtue of lifting the restore out of `build_package()`)

I chose to implement a simple task queue instead of using parallel STL to avoid introducing additional C++17 dependencies in case we need to port down to older compilers/systems. Pending perf results for Windows, I might add a win32-only variant that utilizes the threadpool via `std::async()`.

Data:
1. Linux container
    - 32/64 core desktop machine (AMD 3970X) running Linux in Docker Desktop with Windows 10 as the host OS
    - Installing cpprestsdk (64 packages, including significant sections of boost)
    - Time to unzip serially (old): 1864 ms (1864178 us)
    - Time to unzip in parallel (new): 743 ms
    - **Result: 2.5x speedup**

2. Windows Laptop
    - 4/8 core laptop (Intel i7-1065G7) running Win10
    - Tested under a "developer workstation load" (vscode running, MS Teams running, outlook running, etc).
    - Installing cpprestsdk, boost-system, and tinyxml (12 packages)
    - Time to unzip serially (old): 3.5 s
    - Time to unzip in parallel (new): 2.634 s
    - **Result: 1.33x speedup**

Additional notes: the serial times are very conservative; they were calculated by adding together the reported times by `cmd_execute_and_stream_data()` instead of measuring the end-to-end time with a timer. The parallel unzip times are measured end-to-end, including filesystem existence checks for cache hit/miss. 

Ex1:
```
[DEBUG] popen(unzip -qq /workspaces/.cache/vcpkg/ff/ffaec50a99e95d77ac0e9f7523d07bb18bb7b832f10d4187d69de4e1c630828c.zip -d/workspaces/vcpkg/packages/boost-iterator_x64-linux 2>&1)
[DEBUG] popen(unzip -qq /workspaces/.cache/vcpkg/88/885a7a3deb03f7f47677d9c29fcf265d9c90f300ba36f26cdc218b711802a48c.zip -d/workspaces/vcpkg/packages/boost-function_x64-linux 2>&1)
...
[DEBUG] popen(unzip -qq /workspaces/.cache/vcpkg/bf/bf9e2b42389152522f8ce53b49cb7962742e4bd25ce6f5527191a8817caac0bc.zip -d/workspaces/vcpkg/packages/boost-exception_x64-linux 2>&1)
[DEBUG] cmd_execute_and_stream_data() returned 0 after     3829 us
[DEBUG] cmd_execute_and_stream_data() returned 0 after     8670 us
[DEBUG] cmd_execute_and_stream_data() returned 0 after    11916 us
...
[DEBUG] cmd_execute_and_stream_data() returned 0 after    13104 us
[DEBUG] cmd_execute_and_stream_data() returned 0 after    11321 us
[DEBUG] cmd_execute_and_stream_data() returned 0 after     9049 us
[DEBUG] cmd_execute_and_stream_data() returned 0 after     9419 us
...
Restored 36 packages from /workspaces/.cache/vcpkg in 268.7 ms. Use --debug to see more details.
```

Ex2:
```
Starting package 3/64: boost-vcpkg-helpers:x64-linux
Installing package boost-vcpkg-helpers[core]:x64-linux...
Elapsed time for package boost-vcpkg-helpers:x64-linux: 870.5 us
Starting package 4/64: boost-predef:x64-linux
Installing package boost-predef[core]:x64-linux...
Elapsed time for package boost-predef:x64-linux: 6.828 ms
Starting package 5/64: boost-config:x64-linux
Installing package boost-config[core]:x64-linux...
Elapsed time for package boost-config:x64-linux: 10.91 ms
Starting package 6/64: boost-winapi:x64-linux
Installing package boost-winapi[core]:x64-linux...
Elapsed time for package boost-winapi:x64-linux: 5.67 ms
```

Ex3:
```
Starting package 1/32: boost-numeric-conversion:x64-linux
Building package boost-numeric-conversion[core]:x64-linux...
-- Downloading https://github.com/boostorg/numeric_conversion/archive/boost-1.76.0.tar.gz -> boostorg-numeric_conversion-boost-1.76.0.tar.gz...
-- Extracting source /workspaces/vcpkg/downloads/boostorg-numeric_conversion-boost-1.76.0.tar.gz
-- Using source at /workspaces/vcpkg/buildtrees/boost-numeric-conversion/src/ost-1.76.0-7c72fef13f.clean
-- Packaging headers
-- Packaging headers done
-- Using /workspaces/vcpkg/downloads/boost_LICENSE_1_0.txt
-- Installing: /workspaces/vcpkg/packages/boost-numeric-conversion_x64-linux/share/boost-numeric-conversion/usage
-- Installing: /workspaces/vcpkg/packages/boost-numeric-conversion_x64-linux/share/boost-numeric-conversion/copyright
Stored binary cache: /workspaces/.cache/vcpkg/1c/1cecde4ed646d6acead889c64462aa79965866c7b9850ad9e4a5e03246af0dc0.zip
Installing package boost-numeric-conversion[core]:x64-linux...
Elapsed time for package boost-numeric-conversion:x64-linux: 922.3 ms
```

+@coryan for a glance over the changes to the gcs binary provider


src/vcpkg/base/system.cpp
```diff
@@ -325,7 +325,22 @@ namespace vcpkg
...
+    {
+        static int concurrency = [] {
+            auto user_defined_concurrency = get_environment_variable("VCPKG_MAX_CONCURRENCY");
+            if (user_defined_concurrency)
+            {
+                return std::stoi(user_defined_concurrency.value_or_exit(VCPKG_LINE_INFO));
+            }
+            else
+            {
+                return static_cast<int>(std::thread::hardware_concurrency()) + 1;
```
coryan, line 338: maybe `return max(1, static_cast<int>(std::thread::hardware_concurrency());` why add another thread?

src/vcpkg/binarycaching.cpp
```diff
@@ -950,25 +1017,40 @@ namespace
...
                 print2("Attempting to fetch ", url_paths.size(), " packages from GCS.\n");
+                std::vector<Command> jobs;
+                std::vector<size_t> idxs;
                 for (size_t idx = 0; idx < url_paths.size(); ++idx)
                 {
                     auto&& action = actions[url_indices[idx]];
                     auto&& url_path = url_paths[idx];
                     if (!gsutil_download_file(url_path.first, url_path.second)) continue;
-                    if (decompress_archive(paths, paths.package_dir(action.spec), url_path.second).exit_code != 0)
+                    jobs.push_back(decompress_archive_cmd(paths, paths.package_dir(action.spec), url_path.second));
```
coryan, line 1027: This is all good.

Separate / future discussion: have you folks considered using something like `expected<T>` (or `std::variant<T, Error>`) to simplify these loops and eliminate the need for the "original index" stuff?

coryan (commented):
The GCS changes LGTM.

src/vcpkg/base/system.cpp
```diff
@@ -325,7 +325,22 @@ namespace vcpkg
...
     }
 
-    int get_num_logical_cores() { return std::thread::hardware_concurrency(); }
+    int get_concurrency()
+    {
+        static int concurrency = [] {
+            auto user_defined_concurrency = get_environment_variable("VCPKG_MAX_CONCURRENCY");
+            if (user_defined_concurrency)
+            {
+                return std::stoi(user_defined_concurrency.value_or_exit(VCPKG_LINE_INFO));
```
strega-nil-ms, line 334: what if `user_defined_concurrency` is not a positive integer?

src/vcpkg/base/system.cpp
ras0219-msft: Fwiw this is not a behavior change; this is the behavior we had before.

I believe this comes from the apocryphal idea that the compiler threads will sometimes be cpu-bound and sometimes be io-bound and thus you want to oversubscribe to ensure you saturate the cpu (so 1 thread at a time can be waiting for i/o).

strega-nil-ms (approved):

src/vcpkg/base/system.cpp
```diff
@@ -131,11 +131,11 @@ namespace vcpkg
         Checks::check_exit(VCPKG_LINE_INFO, sz2 + 1 == sz);
         ret.pop_back();
         return Strings::to_utf8(ret.c_str());
-#else // ^^^ defined(_WIN32) / !defined(_WIN32) vvv
+#else
```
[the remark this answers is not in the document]
ras0219-msft: It's encoding sensitive; so we'd need to use the `_wgetenv` extension anyway. If we're going to the trouble of `#ifdef`'ing it specifically for Windows, it seems simpler to just call the OS API directly and not worry about how the CRT state might differ.

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/getenv-wgetenv?view=msvc-160

src/vcpkg/binarycaching.cpp
ras0219-msft: We do have `ExpectedT<T,Error>` in the codebase -- what pattern are you suggesting?

The reason for this "Struct of Arrays" vs "Array of Structs" approach is to enable the parallelization function (`cmd_execute_x_parallel`) to be completely unaware of the outer job indexing, associated data, etc.

src/vcpkg/base/system.cpp
ras0219-msft: Then `get_concurrency()` will return a negative number; that's why it returns `int`. Agreed this is not a useful value, but this is not a code change from today -- the only use of `get_num_logical_cores()` was in `get_concurrency()` elsewhere which had this exact same semantic.

PR #163, opened 2021-08-18
Add post build check to verify that the libs on macOS have the right architecture

autoantwort:
Fixes #19623


Neumann-A:
hmm can this be generalized to include apple and normal linux with objdump instead?
```
neumann@heineken:~$ objdump -f /usr/lib/libvmtools.so.0

/usr/lib/libvmtools.so.0:     file format elf64-x86-64
architecture: i386:x86-64, flags 0x00000150:
HAS_SYMS, DYNAMIC, D_PAGED
start address 0x000000000001f9e0
```

autoantwort:
I don't know objdump very good, but it seems that it does not support universal binaries:
```
➜  vcpkg git:(update-fontconfig) objdump -f /Applications/Firefox.app/Contents/MacOS/libosclientcerts.dylib
➜  vcpkg git:(update-fontconfig) lipo -archs /Applications/Firefox.app/Contents/MacOS/libosclientcerts.dylib
x86_64 arm64
```
it also returns sometimes errors where lipo just works:
```
➜  vcpkg git:(update-fontconfig) objdump -f  /Users/leanderSchulten/git_projekte/Lichtsteuerung/vcpkg/packages/gettext_x64-osx/debug/lib/libintl.a 

/Users/leanderSchulten/git_projekte/Lichtsteuerung/vcpkg/packages/gettext_x64-osx/debug/lib/libintl.a(bindtextdom.o):   file format mach-o arm64

/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/objdump: error: 'bindtextdom.o': Invalid/Unsupported object file format
➜  vcpkg git:(update-fontconfig) echo $?                                                                                                          
1
```

autoantwort:
Here an example that shows that it works:
<details><summary>Log</summary>

```
$ vcpkg install --triplet=x64-osx --host-triplet=x64-osx --binarysource=clear
... 
Starting package 2/3: gettext:x64-osx
Building package gettext[core,tools]:x64-osx...
-- Downloading https://ftp.gnu.org/pub/gnu/gettext/gettext-0.21.tar.gz;https://www.mirrorservice.org/sites/ftp.gnu.org/gnu/gettext/gettext-0.21.tar.gz -> gettext-0.21.tar.gz...
-- Cleaning sources at /Users/leanderSchulten/git_projekte/Lichtsteuerung/vcpkg/buildtrees/gettext/src/0.21-d3cbdc875d.clean. Use --editable to skip cleaning for the packages you specify.
-- Extracting source /Users/leanderSchulten/git_projekte/Lichtsteuerung/vcpkg/downloads/gettext-0.21.tar.gz
-- Applying patch 0002-Fix-uwp-build.patch
-- Applying patch 0003-Fix-win-unicode-paths.patch
-- Applying patch rel_path.patch
-- Applying patch android.patch
-- Using source at /Users/leanderSchulten/git_projekte/Lichtsteuerung/vcpkg/buildtrees/gettext/src/0.21-d3cbdc875d.clean
-- Getting CMake variables for x64-osx-rel
-- Configuring x64-osx-rel
-- Building x64-osx-rel
-- Installing x64-osx-rel
-- Configuring x64-osx-dbg
-- Building x64-osx-dbg
-- Installing x64-osx-dbg
The following files were built for an incorrect architecture:

    /Users/leanderSchulten/git_projekte/Lichtsteuerung/vcpkg/packages/gettext_x64-osx/debug/lib/libintl.a
Expected x64, but was: arm64


Found 1 error(s). Please correct the portfile:
    /Users/leanderSchulten/git_projekte/Lichtsteuerung/vcpkg/ports/gettext/portfile.cmake
Error: Building package gettext:x64-osx failed with: POST_BUILD_CHECKS_FAILED
Please ensure you're using the latest portfiles with `./vcpkg update`, then
submit an issue at https://github.com/Microsoft/vcpkg/issues including:
  Package: gettext:x64-osx
  Vcpkg version: 2021-08-12-unknownhash-debug

Additionally, attach any relevant sections from the log files above.

```
</details>

strega-nil-ms (approved):
This looks really good; we should also set up testing for arm64-osx I think.

src/vcpkg/postbuildlint.cpp
```diff
@@ -578,6 +581,26 @@ namespace vcpkg::PostBuildLint
...
+#elif defined(__APPLE__)
+        std::vector<FileAndArch> binaries_with_invalid_architecture;
+        const auto requested_arch = expected_architecture == "x64" ? "x86_64" : expected_architecture;
+        for (const Path& file : files)
+        {
+            auto cmd_line = Command("lipo").string_arg("-archs").path_arg(file);
+            ExitCodeAndOutput ec_data = cmd_execute_and_capture_output(cmd_line);
+            Checks::check_exit(
+                VCPKG_LINE_INFO, ec_data.exit_code == 0, "Running command:\n   %s\n failed", cmd_line.command_line());
+            if (ec_data.output.find(requested_arch) == std::string::npos)
```
ras0219-msft, line 598: Will this find `x86` inside `x86_64`?

src/vcpkg/postbuildlint.cpp
```diff
@@ -578,6 +581,26 @@ namespace vcpkg::PostBuildLint
             }
         }
 
+        if (!binaries_with_invalid_architecture.empty())
+        {
+            print_invalid_architecture_files(expected_architecture, binaries_with_invalid_architecture);
+            return LintStatus::ERROR_DETECTED;
+        }
+#elif defined(__APPLE__)
+        std::vector<FileAndArch> binaries_with_invalid_architecture;
```
ras0219-msft, line 590: I think this would be better implemented as a separate function unless it can actually share some of the implementation.

src/vcpkg/postbuildlint.cpp
```diff
@@ -891,9 +914,9 @@ namespace vcpkg::PostBuildLint
         const auto release_bin_dir = package_dir / "bin";
 
         std::vector<Path> debug_libs = fs.get_regular_files_recursive(debug_lib_dir, IgnoreErrors{});
-        Util::erase_remove_if(debug_libs, NotExtensionCaseInsensitive{".lib"});
+        Util::erase_remove_if(debug_libs, NotExtensionsCaseInsensitive{{".lib", ".so", ".a", ".dylib"}});
```
ras0219-msft, line 917: This will break cross-compilation from Windows to Linux (e.g. Android).

I think the cleanest approach will be to split it per target (not host) instead of attempting some merged path.

ras0219-msft (changes requested):

src/vcpkg/postbuildlint.cpp
autoantwort: Yeah the function name was very matching and it is doing the, but now for mac. If you want I can change it, but I would prefer the current solution.

src/vcpkg/postbuildlint.cpp
autoantwort: There is no `x86` on macOS. They removed the 32 bit support on macOS and ios a few years ago.

autoantwort:
What is the state of this PR?

src/vcpkg/postbuildlint.cpp
strega-nil-ms: it was like 10.7 where they removed x86_32 machine support, and 10.14 is the last version of macOS that supported 32-bit applications. However, I'd still appreciate doing a split on spaces and checking if any of the architectures equal, not just doing a find.

src/vcpkg/postbuildlint.cpp
```diff
@@ -890,10 +903,15 @@ namespace vcpkg::PostBuildLint
         const auto debug_bin_dir = package_dir / "debug/bin";
         const auto release_bin_dir = package_dir / "bin";
 
+#ifdef __APPLE__
+        NotExtensionsCaseInsensitive libFilter{{".so", ".a", ".dylib"}};
+#else
+        NotExtensionCaseInsensitive libFilter{".lib"};
+#endif
```
strega-nil-ms, line 910:
```suggestion
#ifdef __APPLE__
        NotExtensionsCaseInsensitive lib_filter{{".so", ".a", ".dylib"}};
#else
        NotExtensionCaseInsensitive lib_filter{".lib"};
#endif
```

strega-nil-ms: This should also probably check if the target is osx or windows, not if the host is osx.

strega-nil-ms:
It looks like this has an infinite loop; can you take a look?

autoantwort:
> It looks like this has an infinite loop; can you take a look?

For me building any package on windows fails with 
```
Starting package 2/2: zlib:x86-windows
Building package zlib[core]:x86-windows...
-- Using C:/Users/LeanderSchulten/git_projects/vcpkg/downloads/zlib1211.tar.gz
-- Cleaning sources at C:/Users/LeanderSchulten/git_projects/vcpkg/buildtrees/zlib/src/1.2.11-e9a093319e.clean. Use --editable to skip cleaning for the packages you specify.
-- Extracting source C:/Users/LeanderSchulten/git_projects/vcpkg/downloads/zlib1211.tar.gz
-- Applying patch cmake_dont_build_more_than_needed.patch
-- Applying patch 0001-Prevent-invalid-inclusions-when-HAVE_-is-set-to-0.patch
-- Applying patch add_debug_postfix_on_mingw.patch
-- Applying patch 0002-android-build-mingw.patch
-- Using source at C:/Users/LeanderSchulten/git_projects/vcpkg/buildtrees/zlib/src/1.2.11-e9a093319e.clean
-- Configuring x86-windows
-- Building x86-windows-dbg
-- Building x86-windows-rel
-- Installing: C:/Users/LeanderSchulten/git_projects/vcpkg/packages/zlib_x86-windows/share/zlib/vcpkg-cmake-wrapper.cmake
-- Using msys root at C:/Users/LeanderSchulten/git_projects/vcpkg/downloads/tools/msys2/aa5af7b2aa7e90e8
-- Fixing pkgconfig file: C:/Users/LeanderSchulten/git_projects/vcpkg/packages/zlib_x86-windows/lib/pkgconfig/zlib.pc
-- Fixing pkgconfig file: C:/Users/LeanderSchulten/git_projects/vcpkg/packages/zlib_x86-windows/debug/lib/pkgconfig/zlib.pc
-- Installing: C:/Users/LeanderSchulten/git_projects/vcpkg/packages/zlib_x86-windows/share/zlib/copyright
CMake Warning at scripts/cmake/vcpkg_copy_pdbs.cmake:70 (message):
  Could not find a matching pdb file for:



Call Stack (most recent call first):
  ports/zlib/portfile.cmake:54 (vcpkg_copy_pdbs)
  scripts/ports.cmake:141 (include)


Import libs were not present in C:\Users\LeanderSchulten\git_projects\vcpkg\packages\zlib_x86-windows\debug/lib
If this is intended, add the following line in the portfile:
    SET(VCPKG_POLICY_DLLS_WITHOUT_LIBS enabled)
Import libs were not present in C:\Users\LeanderSchulten\git_projects\vcpkg\packages\zlib_x86-windows\lib
If this is intended, add the following line in the portfile:
    SET(VCPKG_POLICY_DLLS_WITHOUT_LIBS enabled)
Found 2 error(s). Please correct the portfile:
    C:\Users\LeanderSchulten\git_projects\vcpkg\ports\zlib\portfile.cmake
Error: Building package zlib:x86-windows failed with: POST_BUILD_CHECKS_FAILED
Please ensure you're using the latest portfiles with `.\vcpkg update`, then
submit an issue at https://github.com/Microsoft/vcpkg/issues including:
  Package: zlib[core]:x86-windows -> 1.2.11#11
  Vcpkg version: 2021-09-10-unknownhash-debug

Additionally, attach any relevant sections from the log files above.
```

src/vcpkg/postbuildlint.cpp
```diff
@@ -577,16 +576,30 @@ namespace vcpkg::PostBuildLint
             {
                 binaries_with_invalid_architecture.push_back({file, actual_architecture});
             }
-        }
+        };
+#elif defined(__APPLE__)
+        const auto requested_arch = expected_architecture == "x64" ? "x86_64" : expected_architecture;
+        const auto checkLib = [&](const Path& file) {
```
strega-nil-ms, line 582:
```suggestion
        const auto check_lib = [&](const Path& file) {
```

autoantwort:
Any news?

src/vcpkg/postbuildlint.cpp
```diff
@@ -590,10 +590,10 @@ namespace vcpkg::PostBuildLint
             }
         };
 #else
-        const auto checkLib = [&](const Path&) {};
+        const auto chec_lib = [&](const Path&) {};
```
strega-nil-ms:
```suggestion
        const auto check_lib = [&](const Path&) {};
```

include/vcpkg/base/util.h
```diff
@@ -142,6 +142,13 @@ namespace vcpkg::Util
         return std::find(begin(cont), end(cont), v);
     }
 
+    template<class Container, class V>
+    bool contains(const Container& cont, const V& v)
+    {
+        using std::end;
+        return Util::find(cont, v) != end(cont);
+    }
+
```
strega-nil-ms, line 151:
```suggestion
    template<class Range, class T>
    bool contains(const Range& r, const T& el)
    {
        using std::end;
        return Util::find(r, el) != end(r);
    }

```

strega-nil-ms (approved):

strega-nil-ms:
I do want @ras0219-msft's review on this.

autoantwort:
Ping @ras0219-msft for review :)

src/vcpkg/postbuildlint.cpp
```diff
@@ -577,16 +576,30 @@ namespace vcpkg::PostBuildLint
...
             }
-        }
+        };
+#elif defined(__APPLE__)
+        const auto requested_arch = expected_architecture == "x64" ? "x86_64" : expected_architecture;
+        const auto check_lib = [&](const Path& file) {
+            auto cmd_line = Command("lipo").string_arg("-archs").path_arg(file);
+            ExitCodeAndOutput ec_data = cmd_execute_and_capture_output(cmd_line);
+            Checks::check_exit(
+                VCPKG_LINE_INFO, ec_data.exit_code == 0, "Running command:\n   %s\n failed", cmd_line.command_line());
```
ras0219-msft, line 586:
```suggestion
                VCPKG_LINE_INFO, ec_data.exit_code == 0, "Error: unable to determine the architectures of binary file %s. Running lipo failed with status code %d\n    %s", file, ec_data.exit_code, cmd_line.command_line());
```
```suggestion
                VCPKG_LINE_INFO, ec_data.exit_code == 0, "Running command:\n   %s\n failed", cmd_line.command_line());
```

src/vcpkg/postbuildlint.cpp
```diff
@@ -1012,10 +1025,20 @@ namespace vcpkg::PostBuildLint
         const auto debug_bin_dir = package_dir / "debug" / "bin";
         const auto release_bin_dir = package_dir / "bin";
 
+        const auto lib_filter = [&pre_build_info]() {
+            if (pre_build_info.cmake_system_name.empty() || pre_build_info.cmake_system_name == "Windows" ||
+                pre_build_info.cmake_system_name == "WindowsStore")
+            {
+                if (pre_build_info.cmake_system_name == "MinGW") return NotExtensionsCaseInsensitive{{".lib", ".a"}};
```
ras0219-msft, line 1032: This condition is never true -- system name here is exactly empty, "Windows", or "WindowsStore". Was this intended to be an `else if()`?

ras0219-msft (commented):
Approved after my comments are addressed. Thanks for the PR and happy holidays!

src/vcpkg/postbuildlint.cpp
```diff
@@ -554,11 +555,9 @@ namespace vcpkg::PostBuildLint
                                              const std::vector<Path>& files,
                                              const Filesystem& fs)
     {
-#if defined(_WIN32)
         std::vector<FileAndArch> binaries_with_invalid_architecture;
-
-        for (const Path& file : files)
-        {
+#if defined(_WIN32)
```
Osyotr, line 559: IIUC, this check will be skipped if vcpkg is running on Linux/Mac host but the library is being built for windows (MinGW or something).

src/vcpkg/postbuildlint.cpp
```diff
@@ -577,16 +576,35 @@ namespace vcpkg::PostBuildLint
...
                 binaries_with_invalid_architecture.push_back({file, actual_architecture});
             }
-        }
+        };
+#elif defined(__APPLE__)
+        const auto requested_arch = expected_architecture == "x64" ? "x86_64" : expected_architecture;
+        const auto check_lib = [&](const Path& file) {
+            auto cmd_line = Command("lipo").string_arg("-archs").path_arg(file);
+            ExitCodeAndOutput ec_data = cmd_execute_and_capture_output(cmd_line);
+            Checks::check_exit(VCPKG_LINE_INFO,
```
ras0219-msft, line 585: Since we have downgraded all checks to optional, it no longer makes sense to hard crash if we fail to run `lipo` or `dumpbin`.

src/vcpkg/postbuildlint.cpp
ras0219-msft: +1 to this, this should really be based on the target triplet, not the OS that vcpkg-tool was built on.

We even are already doing this below w.r.t. the `lib_filter`

ras0219-msft (approved):
I think there are still a few deficiencies, but I don't think this makes our current situation worse except perhaps cross-compiling from MacOS to Windows with MinGW, which according to my reading will now crash when lipo fails to understand the PE binaries.

strega-nil-ms (approved):

dg0yt:
> except perhaps cross-compiling from MacOS to Windows with MinGW

... or mingw in general?  #23450

PR #165, opened 2021-08-20
[vcpkg] rolling back versions should be allowed

strega-nil-ms:
This removes the check for "the version in the thing is the latest".


vicroms (approved):
LGTM

PR #173, opened 2021-08-25
Document the dependencies graph code

autoantwort:
The dependency code is imo hard to understand for new developers. 2000 lines of code without a comment. Nowhere general concepts are explained. Then there is a mixture of classic mode and manifest mode code. That nearly all functions of the class `VersionedPackageGraph` have the name `add_constraint` also does not really help 😅 (without a debugger really hard to understand). 

I renamed some functions to make it easier to understand the code. The whole thing startet because I wanted to implement "default-features": false as default for host dependencies.

One of the questions currently: What is the purpose of the `FeatureSpec` (`PackageSpec` + one feature)? It seems that it is used to create the installation plan


src/vcpkg/dependencies.cpp
```diff
@@ -1250,36 +1250,41 @@ namespace vcpkg::Dependencies
                 std::vector<std::string> features;
             };
 
-            // This object contains the current version within a given column.
-            // Each "string" scheme text is treated as a separate column
-            // "relaxed" versions all share the same column
+            // This object contains the current version within a given version sheme (except for the "string" scheme,
```
strega-nil-ms, line 1253:
```suggestion
            // This object contains the current version within a given version scheme (except for the "string" scheme,
```

src/vcpkg/dependencies.cpp
```diff
@@ -1250,36 +1250,41 @@ namespace vcpkg::Dependencies
...
                 bool is_less_than(const Versions::Version& new_ver) const;
             };
 
             struct PackageNode
             {
+                // Mapping from version to the newest version in the cooresponding version sheme
                 std::map<Versions::Version, VersionSchemeInfo*, VersionTMapLess> vermap;
+                // We don't know how to compare "string" versions, so we save all versions seperatly
                 std::map<std::string, VersionSchemeInfo> exacts;
+                // Within each version sheme("version", semver, date) we only track the newest version
```
strega-nil-ms, line 1274:
```suggestion
                // for each version type besides string (relaxed, semver, date), we only track the latest version we know about
```

src/vcpkg/dependencies.cpp
```diff
@@ -1250,36 +1250,41 @@ namespace vcpkg::Dependencies
...
                 std::map<std::string, std::vector<FeatureSpec>> deps;
 
                 bool is_less_than(const Versions::Version& new_ver) const;
             };
 
             struct PackageNode
             {
+                // Mapping from version to the newest version in the cooresponding version sheme
                 std::map<Versions::Version, VersionSchemeInfo*, VersionTMapLess> vermap;
+                // We don't know how to compare "string" versions, so we save all versions seperatly
```
strega-nil-ms, line 1272:
```suggestion
                // We don't know how to compare "version-string" versions, so keep all the versions separately
```

src/vcpkg/dependencies.cpp
```diff
@@ -1250,36 +1250,41 @@ namespace vcpkg::Dependencies
...
                 std::vector<std::string> origins;
+                // mapping from feature name -> dependencies of this feature
                 std::map<std::string, std::vector<FeatureSpec>> deps;
 
                 bool is_less_than(const Versions::Version& new_ver) const;
             };
 
             struct PackageNode
             {
+                // Mapping from version to the newest version in the cooresponding version sheme
```
strega-nil-ms, line 1270:
```suggestion
                // Mapping from version to the newest version in the corresponding version scheme
                // For example, given the versions:
                //   - "version-string": "1.0.0"
                //   - "version": "1.0.1"
                //   - "version": "1.0.2"
                // you'd have a map:
                //   {
                //     "1.0.0": { "version-string": "1.0.0" },
                //     "1.0.1": { "version": "1.0.2" },
                //     "1.0.2": { "version": "1.0.2" }
                //  }
```

(assuming my understanding of this is correct)

src/vcpkg/dependencies.cpp
```diff
@@ -1250,36 +1250,41 @@ namespace vcpkg::Dependencies
...
                 Optional<std::unique_ptr<VersionSchemeInfo>> relaxed;
                 Optional<std::unique_ptr<VersionSchemeInfo>> semver;
                 Optional<std::unique_ptr<VersionSchemeInfo>> date;
-                std::set<std::string> features;
+                std::set<std::string> requested_features;
                 bool default_features = true;
                 bool user_requested = false;
 
                 VersionSchemeInfo* get_node(const Versions::Version& ver);
+                // Created/Updates(is newer than existing) the version entry for the given sheme
```
strega-nil-ms, line 1283:
```suggestion
                // Adds the version to the version resolver:
                //   - for string version schemes, just adds the newer version to the set
                //   - for non-string version schemes:
                //     - if the scheme doesn't exist in the set, adds the version to the set
                //     - if the scheme already exists in the set, and the version is newer than the existing entry, replaces the current entry for the scheme
```

src/vcpkg/dependencies.cpp
```diff
@@ -1250,36 +1250,41 @@ namespace vcpkg::Dependencies
...
                 VersionSchemeInfo& emplace_node(Versions::Scheme scheme, const Versions::Version& ver);
 
                 PackageNode() = default;
+                PackageNode(bool default_features) : default_features(default_features){};
```
strega-nil-ms, line 1287:
```suggestion
                explicit PackageNode(bool default_features) : default_features(default_features){};
```
This should also probably be an `enum class` `DefaultFeatures::{Yes, No}`

strega-nil-ms (changes requested):
@ras0219-msft needs to review this.

src/vcpkg/dependencies.cpp
```diff
@@ -1307,24 +1312,22 @@ namespace vcpkg::Dependencies
                 }
             };
 
-            std::vector<DepSpec> m_roots;
+            std::vector<DepSpec> m_roots; // the roots of the dependency graph (given in the manifest file)
```
ras0219-msft, line 1315:
```suggestion
            // the roots of the dependency graph (given in the manifest file)
            std::vector<DepSpec> m_roots;
```

src/vcpkg/dependencies.cpp
```diff
@@ -1307,24 +1312,22 @@ namespace vcpkg::Dependencies
...
+            // mapping from portname -> version. "overrides" field in manifest file
             std::map<std::string, Versions::Version> m_overrides;
-            std::map<PackageSpec, PackageNode> m_graph;
+            std::map<PackageSpec, PackageNode> m_graph; // mapping from (portname, triplet) -> what to install
```
ras0219-msft, line 1318:
```suggestion
            // mapping from { package specifier -> node containing resolution information for that package }
            std::map<PackageSpec, PackageNode> m_graph;
```

src/vcpkg/dependencies.cpp
```diff
@@ -1307,24 +1312,22 @@ namespace vcpkg::Dependencies
...
-            void add_constraint(std::pair<const PackageSpec, PackageNode>& ref,
-                                const Versions::Version& ver,
-                                const std::string& origin);
-            void add_constraint(std::pair<const PackageSpec, PackageNode>& ref,
-                                const std::string& feature,
-                                const std::string& origin);
-
-            void add_constraint_default_features(std::pair<const PackageSpec, PackageNode>& ref,
-                                                 const std::string& origin);
+            void add_dependency_from(std::pair<const PackageSpec, PackageNode>& ref,
```
ras0219-msft, line 1322: The reason these are all named "add_constraint" is because they are just that: constraints. These functions do not passively "make this version known to the algorithm"; they have specific semantic effects upon the results.

I'd be fine with seeing them renamed separately, but their names should reflect their semantics. As a suggestion:
- require_dependency
- require_version
- require_feature

or

- add_package_constraint
- add_version_constraint
- add_feature_constraint

src/vcpkg/dependencies.cpp
```diff
@@ -1553,84 +1540,87 @@ namespace vcpkg::Dependencies
...
-                                                   const Versions::Version& version,
-                                                   const std::string& origin)
+        void VersionedPackageGraph::add_port_at_version(std::pair<const PackageSpec, PackageNode>& graph_entry,
+                                                        const Versions::Version& version,
+                                                        const std::string& origin)
         {
             ExpectedS<const vcpkg::SourceControlFileLocation&> maybe_scfl;
 
-            auto maybe_overlay = m_o_provider.get_control_file(ref.first.name());
+            // if this is a overload port, ignore the given version and use the version from the overlay
```
ras0219-msft, line 1563:
```suggestion
            // if this port is an overlay port, ignore the given version and use the version from the overlay
```

src/vcpkg/dependencies.cpp
```diff
@@ -1553,84 +1540,87 @@ namespace vcpkg::Dependencies
...
+            auto maybe_overlay = m_o_provider.get_control_file(graph_entry.first.name());
             if (auto p_overlay = maybe_overlay.get())
             {
                 auto overlay_version = to_version(*p_overlay->source_control_file);
-                if (version != overlay_version)
+                if (version != overlay_version) // prevent endless recursion
```
ras0219-msft, line 1568:
```suggestion
                // If the original request did not match the overlay version, restart this function to operate on the overlay version
                if (version != overlay_version)
```

src/vcpkg/dependencies.cpp
```diff
@@ -1715,43 +1705,33 @@ namespace vcpkg::Dependencies
             Util::sort_unique_erase(specs);
             m_var_provider.load_dep_info_vars(specs);
             const auto& vars = m_var_provider.get_dep_info_vars(toplevel).value_or_exit(VCPKG_LINE_INFO);
-            std::vector<const Dependency*> active_deps;
```
ras0219-msft, line 1708: This is semantically changing the algorithm and should not be done during a documentation pass.

We perform this sequence to ensure the `default_features` of each top level package is set to false _before_ recursing into the individual packages. Otherwise, a case like:

```
A -> B, C[core]
B -> C
```
could install the default features of `C`.

autoantwort:
@strega-nil-ms @ras0219-msft Thank you for your help, I think I now understood how the algorithm worked :)

autoantwort:
Any news?

ras0219-msft:
This LGTM, thanks for the code quality improvements!

PR #174, opened 2021-08-26
[asset caching] Add cli parameter to pass asset sources

autoantwort:
Closing https://discord.com/channels/400588936151433218/687365466422902841/877303084940918814

### DIFF
--- a/src/vcpkg-test/binarycaching.cpp
+++ b/src/vcpkg-test/binarycaching.cpp
@@ -229,6 +229,11 @@ TEST_CASE ("generate_nuspec", "[generate_nuspec]")
 {
     auto& fsWrapper = get_real_filesystem();
     VcpkgCmdArguments args = VcpkgCmdArguments::create_from_arg_sequence(nullptr, nullptr);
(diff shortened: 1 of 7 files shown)


ras0219-msft (approved):
While reviewing, I removed the dependence upon getenv/setenv in the tests to better ensure repeatability and unit test isolation.

Thanks for the PR!

src/vcpkg-test/binarycaching.cpp
```diff
@@ -229,7 +229,6 @@ TEST_CASE ("generate_nuspec", "[generate_nuspec]")
 {
     auto& fsWrapper = get_real_filesystem();
     VcpkgCmdArguments args = VcpkgCmdArguments::create_from_arg_sequence(nullptr, nullptr);
-    args.imbue_from_environment();
```
autoantwort, line 232: I can imagine this was needed for the VCPKG_ROOT variable. Now it fails because it couldn't found the vcpkg root.

ras0219-msft: Ah, foo. Thanks for the catch.

ras0219-msft:
LGTM, thanks!

PR #176, opened 2021-08-28
Rename SourceControlFileLocation to SourceControlFileAndLocation

autoantwort:
As mentioned in #173 the current name is imo confusing


strega-nil-ms (approved):
I like this a lot, it confused me greatly when I was first writing code in vcpkg.

autoantwort:
So this can be merged?

strega-nil-ms:
I want one more approval

ras0219-msft (approved):

strega-nil-ms:
Thanks @autoantwort :)

PR #184, opened 2021-09-05
 support-expressions: Add support expressions to features. Error when supports expression evaluates to false in classic and manifest mode.

autoantwort:
Fixes https://github.com/microsoft/vcpkg/issues/17651

### DIFF
--- a/include/vcpkg/dependencies.h
+++ b/include/vcpkg/dependencies.h
@@ -25,6 +25,12 @@ namespace vcpkg
 
 namespace vcpkg::Dependencies
 {
+    enum class SupportExpressionAction : bool
+    {
+        Warn,
+        Error,
+    };
+
     enum class RequestType
     {
         UNKNOWN,
@@ -118,6 +124,7 @@ namespace vcpkg::Dependencies
         std::vector<RemovePlanAction> remove_actions;
         std::vector<InstallPlanAction> already_installed;
         std::vector<InstallPlanAction> install_actions;
+        std::vector<std::string> warnings;
     };
 
     enum class ExportPlanType
@@ -159,10 +166,14 @@ namespace vcpkg::Dependencies
     struct CreateInstallPlanOptions
     {
         CreateInstallPlanOptions(Graphs::Randomizer* r, Triplet t) : randomizer(r), host_triplet(t) { }
-        CreateInstallPlanOptions(Triplet t) : host_triplet(t) { }
+        CreateInstallPlanOptions(Triplet t, SupportExpressionAction action = SupportExpressionAction::Error)
+            : host_triplet(t), support_expression_action(action)
+        {
+        }
 
         Graphs::Randomizer* randomizer = nullptr;
         Triplet host_triplet;
+        SupportExpressionAction support_expression_action;
     };
 
     std::vector<RemovePlanAction> create_remove_plan(const std::vector<PackageSpec>& specs,
@@ -201,7 +212,8 @@ namespace vcpkg::Dependencies
                                                         const std::vector<Dependency>& deps,
                                                         const std::vector<DependencyOverride>& overrides,
                                                         const PackageSpec& toplevel,
-                                                        Triplet host_triplet);
+                                                        Triplet host_triplet,
+                                                        SupportExpressionAction support_expression_action);
 
     void print_plan(const ActionPlan& action_plan, const bool is_recursive = true, const Path& builtin_ports_dir = {});
 }
--- a/include/vcpkg/dependencies.h
+++ b/include/vcpkg/dependencies.h
@@ -25,7 +25,7 @@ namespace vcpkg
 
 namespace vcpkg::Dependencies
 {
-    enum class SupportExpressionAction : bool
+    enum class UnsupportedPortAction : bool
     {
         Warn,
         Error,
@@ -166,14 +166,14 @@ namespace vcpkg::Dependencies
     struct CreateInstallPlanOptions
     {
         CreateInstallPlanOptions(Graphs::Randomizer* r, Triplet t) : randomizer(r), host_triplet(t) { }
-        CreateInstallPlanOptions(Triplet t, SupportExpressionAction action = SupportExpressionAction::Error)
-            : host_triplet(t), support_expression_action(action)
+        CreateInstallPlanOptions(Triplet t, UnsupportedPortAction action = UnsupportedPortAction::Error)
+            : host_triplet(t), unsupported_port_action(action)
         {
         }
 
         Graphs::Randomizer* randomizer = nullptr;
         Triplet host_triplet;
-        SupportExpressionAction support_expression_action;
+        UnsupportedPortAction unsupported_port_action;
     };
 
     std::vector<RemovePlanAction> create_remove_plan(const std::vector<PackageSpec>& specs,
@@ -213,7 +213,7 @@ namespace vcpkg::Dependencies
                                                         const std::vector<DependencyOverride>& overrides,
                                                         const PackageSpec& toplevel,
                                                         Triplet host_triplet,
-                                                        SupportExpressionAction support_expression_action);
+                                                        UnsupportedPortAction unsupported_port_action);
 
     void print_plan(const ActionPlan& action_plan, const bool is_recursive = true, const Path& builtin_ports_dir = {});
 }
--- a/src/vcpkg/commands.upgrade.cpp
+++ b/src/vcpkg/commands.upgrade.cpp
@@ -22,10 +22,13 @@ namespace vcpkg::Commands::Upgrade
 
     static constexpr StringLiteral OPTION_NO_DRY_RUN = "no-dry-run";
     static constexpr StringLiteral OPTION_KEEP_GOING = "keep-going";
+    static constexpr StringLiteral OPTION_ONLY_WARN_UNSUPPORTED = "only-warn-unsupported";
 
-    static constexpr std::array<CommandSwitch, 2> INSTALL_SWITCHES = {{
+    static constexpr std::array<CommandSwitch, 3> INSTALL_SWITCHES = {{
         {OPTION_NO_DRY_RUN, "Actually upgrade"},
         {OPTION_KEEP_GOING, "Continue installing packages on failure"},
+        {OPTION_ONLY_WARN_UNSUPPORTED,
+         "Only issue a warning if a port or feature is unsupported regarding to support expressions"},
     }};
 
     const CommandStructure COMMAND_STRUCTURE = {
@@ -52,6 +55,9 @@ namespace vcpkg::Commands::Upgrade
 
         const bool no_dry_run = Util::Sets::contains(options.switches, OPTION_NO_DRY_RUN);
         const KeepGoing keep_going = to_keep_going(Util::Sets::contains(options.switches, OPTION_KEEP_GOING));
+        const auto support_expression_action = Util::Sets::contains(options.switches, OPTION_ONLY_WARN_UNSUPPORTED)
+                                                   ? Dependencies::SupportExpressionAction::Warn
+                                                   : Dependencies::SupportExpressionAction::Error;
 
         BinaryCache binary_cache{args};
         StatusParagraphs status_db = database_load_check(paths);
@@ -88,7 +94,7 @@ namespace vcpkg::Commands::Upgrade
                 var_provider,
                 Util::fmap(outdated_packages, [](const Update::OutdatedPackage& package) { return package.spec; }),
                 status_db,
-                {host_triplet});
+                {host_triplet, support_expression_action});
         }
         else
         {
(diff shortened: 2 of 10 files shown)


strega-nil-ms (changes requested):
So, I _love_ this change; this would make my experience using vcpkg so much better.

However, I think there needs to be an escape hatch; some kind of "please ignore supports expressions for this install". Additionally, I think it's very likely that we should start by printing warnings instead of errors, since otherwise we will cause breakage on people installing over-constrained packages.

So far, though, this looks amazing. Thank you so much for your work here @autoantwort!

autoantwort:
> Additionally, I think it's very likely that we should start by printing warnings instead of errors, since otherwise we will cause breakage on people installing over-constrained packages.

I have checked some random ports, and the supports expression was almost always back upped by a `vcpkg_fail_port_install(...)` call in the portfile.cmake. So instead of only printing a warning I would prefer to fix the support expressions of the few ports where it is only used to disable the ci build.

But an additional option is needed to disable the failure.

autoantwort:
I have added such an option now

src/vcpkg/install.cpp
```diff
@@ -542,8 +543,10 @@ namespace vcpkg::Install
         {OPTION_CLEAN_DOWNLOADS_AFTER_BUILD, "Clean downloads after building each package"},
         {OPTION_PROHIBIT_BACKCOMPAT_FEATURES,
          "(experimental) Fail install if a package attempts to use a deprecated feature"},
+        {OPTION_ONLY_WARN_UNSUPPORTED,
+         "Only issue a warning if a port or feature is unsupported regarding to support expressions"},
```
strega-nil-ms, line 547:
```suggestion
         "Instead of erroring on an unsupported port, continue with a warning."},
```

src/vcpkg/install.cpp
```diff
@@ -558,6 +561,8 @@ namespace vcpkg::Install
         {OPTION_CLEAN_PACKAGES_AFTER_BUILD, "Clean packages after building each package"},
         {OPTION_CLEAN_DOWNLOADS_AFTER_BUILD, "Clean downloads after building each package"},
         {OPTION_MANIFEST_NO_DEFAULT_FEATURES, "Don't install the default features from the manifest."},
+        {OPTION_ONLY_WARN_UNSUPPORTED,
+         "Only issue a warning if a port or feature is unsupported regarding to support expressions"},
```
strega-nil-ms, line 565:
```suggestion
         "Instead of erroring on an unsupported port, continue with a warning."},
```

strega-nil-ms (changes requested):
Looking good!

src/vcpkg/commands.upgrade.cpp
```diff
@@ -22,10 +22,13 @@ namespace vcpkg::Commands::Upgrade
...
     static constexpr StringLiteral OPTION_NO_DRY_RUN = "no-dry-run";
     static constexpr StringLiteral OPTION_KEEP_GOING = "keep-going";
+    static constexpr StringLiteral OPTION_ONLY_WARN_UNSUPPORTED = "only-warn-unsupported";
 
-    static constexpr std::array<CommandSwitch, 2> INSTALL_SWITCHES = {{
+    static constexpr std::array<CommandSwitch, 3> INSTALL_SWITCHES = {{
         {OPTION_NO_DRY_RUN, "Actually upgrade"},
         {OPTION_KEEP_GOING, "Continue installing packages on failure"},
+        {OPTION_ONLY_WARN_UNSUPPORTED,
+         "Only issue a warning if a port or feature is unsupported regarding to support expressions"},
```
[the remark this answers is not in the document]
strega-nil-ms: Question: does upgrading to a port that is unsupported error out _before_, or _after_ we start removing ports?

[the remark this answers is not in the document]
autoantwort: The support expressions are evaluated while computing the installation plan, that is done before removing ports.

include/vcpkg/dependencies.h
```diff
@@ -25,6 +25,12 @@ namespace vcpkg
 
 namespace vcpkg::Dependencies
 {
+    enum class SupportExpressionAction : bool
```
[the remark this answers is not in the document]
autoantwort: Theoretically this is wrong, because features also have support expressions. But I think we can ignore this fact.

src/vcpkg/dependencies.cpp
```diff
@@ -837,16 +841,58 @@ namespace vcpkg::Dependencies
...
+                        auto maybe_paragraph =
+                            clust.get_scfl_or_exit().source_control_file->find_feature(spec.feature());
+                        Checks::check_maybe_upgrade(VCPKG_LINE_INFO,
+                                                    maybe_paragraph.has_value(),
+                                                    "Package %s does not have a %s feature",
+                                                    spec.name(),
+                                                    spec.feature());
+
+                        supports_expression = &maybe_paragraph.get()->supports_expression;
+                    }
```
BillyONeal, line 875: It looks like this could be extracted into a function like this?:

```
const PlatformExpression::Expr* get_applicable_supports_expression(Cluster& clust, FeatureSpec spec)
```

?

src/vcpkg/dependencies.cpp
```diff
@@ -1895,13 +1944,62 @@ namespace vcpkg::Dependencies
                         "with detailed steps to reproduce the problem.");
                 }
 
+                const auto get_vars = [&]() {
```
BillyONeal, line 1947: It seems like this would be better as just another member function (with a parameter `spec`) than as a lambda?

BillyONeal (approved):
Only extreme nits here.

src/vcpkg/dependencies.cpp
autoantwort: I made this a member function of the cluster

src/vcpkg/dependencies.cpp
autoantwort: Done

strega-nil-ms (changes requested):
Nearly there!

strega-nil-ms (approved):

autoantwort:
I assume a approval from @ras0219-msft is also needed?

strega-nil-ms:
Given two yesses, merging now. Thanks @autoantwort :D

autoantwort:
Maybe you are missing a git pull. I can add `supports` expressions to features and it is definitively working

autoantwort:
>  While ./vcpkg format-manifest ports/port/vcpkg.json simply removes the supports line.

But this seems to be a bug

autoantwort:
> > While ./vcpkg format-manifest ports/port/vcpkg.json simply removes the supports line.
> 
> But this seems to be a bug

Would be fixed by #240

src/vcpkg/dependencies.cpp
```diff
@@ -837,16 +861,41 @@ namespace vcpkg::Dependencies
...
                 }
+                else
+                {
+                    auto supports_expression = clust.get_applicable_supports_expression(spec);
+                    if (supports_expression && !supports_expression.get()->is_empty())
+                    {
+                        if (!supports_expression.get()->evaluate(
+                                m_var_provider.get_dep_info_vars(spec.spec()).value_or_exit(VCPKG_LINE_INFO)))
+                        {
+                            const auto msg = Strings::format("%s[%s] is only supported on '%s'",
```
JackBoosY, line 884: I strongly against this changes!
Since our document does not specify whether to use a whitelist or a blacklist for the supports field, this modification will cause many ports to fail to build on the community triplet.
Please note that this field is only used for pipeline test to judge whether the result is reasonable.

autoantwort: Support expressions supports black and whitelists. For example `!windows` is a blacklist while `osx & linux` is a whitelist. But yeah we should suggest using blacklists to not fail on community triplets unless we know that a port really only works the given set of triplets. 

>  Please note that this field is only used for pipeline test to judge whether the result is reasonable.

Until this PR. Now it can replaces `vcpkg_fail_port_install` calls to get a better user experience. You don't have to build all dependencies to see that the port does not support your triplet.  
If a port is generally supported by a triplet but our ci don't want to build it, you should use `ci.baseline.txt` because the port in general supports the triplet.

BillyONeal:
* `s/blacklist/denylist/g`
* `s/whitelist/allowlist/g`

src/vcpkg/dependencies.cpp
BillyONeal: >Please note that this field is only used for pipeline test to judge whether the result is reasonable.

The intent is that ci.baseline.txt is the 'specific to our pipelines' thing, while `supports` reflects behavior from the upstream sources.

JackBoosY: Therefore, I suggest changes this error to warning message.

JackBoosY: This also breaks option `--only-downloads` and `--keep-going`.

autoantwort: You can use the option `--allow-unsupported`

wrobelda:
@autoantwort I think there's some issue with it in CI/CD. Specifically, in Microsoft/vcpkg#21491 I [added](https://github.com/microsoft/vcpkg/pull/21491/files#diff-a3169b4d3e1319b8fcd9a9695cb1deb242b43ae0eac80a8618c6f4dd5e56d6ff):
 
```
"features": {
     "libmount": {
       "description": "Used by the UDisks backend on Linux",
       "supports": "linux",
       "dependencies": [
         "libmount"
       ]
     }
   }
```

However, it appears that because of it the Windows CI/CD now skips the `kf5build`, since `libmount` is only available on Linux:
https://dev.azure.com/vcpkg/public/_build/results?buildId=63246&view=logs&jobId=7922e5c4-0103-5f8f-ad17-45ce9bb98e80&j=7922e5c4-0103-5f8f-ad17-45ce9bb98e80&t=491b9f02-7edc-5990-cda1-511e95a3768e

dg0yt:
`"platform": "linux"`

dg0yt:
No e2e test for this?

dg0yt:
> However, it appears that because of it the Windows CI/CD now skips the `kf5solid` build altogether, apparently since `libmount` is only available on Linux:

Note that the skip is not because libmount is part of the default features of kf5solid, but because libmount is part of the features requested by depending port kf5io, unconditionally.

ras0219-msft:
Which means the immediate fix is to add a supports clause to `kf5io`; but this is definitely tricky to detect reliably and we need some better diagnostics.

src/vcpkg/dependencies.cpp
JackBoosY: This also breaks command `depend-info` which should nevert fails.

dg0yt: > This also breaks command `depend-info` which should nevert fails.

I don't think depend-info is broken. It works as expected for the target triplet, listing the unsupported dependency when needed.

JackBoosY: @dg0yt depend-info is only used to view dependencies without affecting the build, so I don't think we should stop it here.

dg0yt: @JackBoosY The current behaviour allows me to inspect unsupported dependencies for arm-uwp before attempting to push to CI from x64-linux. This is beneficial.

JackBoosY: > @JackBoosY The current behaviour allows me to inspect unsupported dependencies for arm-uwp before attempting to push to CI from x64-linux. This is beneficial.

At lest depend-info doesn't support extra arg `--allow-unsupported`.

dg0yt: .. and it doesn't allow multiple package names, to limit the feature set of dependencies. :disappointed:

JackBoosY: And this doesn't exists in the usage info.

dg0yt: > And this doesn't exists in the usage info.

I don't understand that point.

JackBoosY: > I don't understand that point.

This option should be printed in the usage info if we need to add it:
```
PS F:\vcpkg> .\vcpkg.exe depend-info vtk[*] --allow-unsupported
Unknown option(s) for command 'depend-info':
    '--allow-unsupported'

Example:
  vcpkg depend-info sqlite3

Options:
  --dot                           Creates graph on basis of dot
  --dgml                          Creates graph on basis of dgml
  --show-depth                    Show recursion depth in output
  --max-recurse=...               Set max recursion depth, a value of -1 indicates no limit
  --sort=...                      Set sort order for the list of dependencies, accepted values are:
                                  lexicographical, topological (default), reverse
  --triplet=<t>                   Specify the target architecture triplet. See 'vcpkg help triplet'
                                  (default: %VCPKG_DEFAULT_TRIPLET%)
  --host-triplet=<t>              Specify the host architecture triplet. See 'vcpkg help triplet'
                                  (default: %VCPKG_DEFAULT_HOST_TRIPLET%)
  --overlay-ports=<path>          Specify directories to be used when searching for ports
                                  (also: %VCPKG_OVERLAY_PORTS%)
  --overlay-triplets=<path>       Specify directories containing triplets files
                                  (also: %VCPKG_OVERLAY_TRIPLETS%)
  --binarysource=<path>           Add sources for binary caching. See 'vcpkg help binarycaching'
  --x-asset-sources=<path>        Add sources for asset caching. See 'vcpkg help assetcaching'
  --downloads-root=<path>         Specify the downloads root directory
                                  (default: %VCPKG_DOWNLOADS%)
  --vcpkg-root=<path>             Specify the vcpkg root directory
                                  (default: %VCPKG_ROOT%)
  --x-buildtrees-root=<path>      (Experimental) Specify the buildtrees root directory
  --x-install-root=<path>         (Experimental) Specify the install root directory
  --x-packages-root=<path>        (Experimental) Specify the packages root directory
  --x-json                        (Experimental) Request JSON output
```

JackBoosY: Fix this in https://github.com/microsoft/vcpkg-tool/pull/300.

JackBoosY: Another issue: https://github.com/microsoft/vcpkg/issues/22278

PR #193, opened 2021-09-11
remove telemetry line by line

playgithub:
A branch without telemetry is welcome.


autoantwort:
You can simply disable the metrics at runtime?

playgithub:
> You can simply disable the metrics at runtime?

I think everyone that choose to disable telemetry prefers this way.

With the ability of telemetry, though disabled, I'm not sure if there is a backdoor.

And there are some codes like
```
if (!was_tracked)
{
    was_tracked = true;
    // do metrics
}
```

ras0219-msft:
Thanks for this PR!

If you'd like to maintain this in your own fork, we have no objection. However, we do not intend to maintain a unique branch of the codebase to support this particular implementation.

We would be happy to accept any PRs that fix issues with our existing mechanisms to disable telemetry though! If you find any cases where we are incorrectly re-enabling telemetry, those are absolutely a bug and should be fixed.

playgithub:
> Finally, if you do want to maintain your own fork, I recommend modifying the fundamental apis in metrics.cpp only and leaving the rest of the codebase untouched. This will make it much easier to merge upstream changes in the future (e.g. when we add, change, or remove calls from the main codebase).

I'd like to remove anything about metric for purity, it is not hard to maintain.
BTW I'm using Linux now, but I have some interest to see how telemetry works.

playgithub:
For tech, the master branch should be  without telemetry, and another branch for telemetry.
For competition advantages, no branch without telemetry.

rouisibra:
I'm not a fan of telemetry for open source software as well since we can just post the issue on GitHub and fix it.

closed source software may need telemetry since the the developers need to fix everything by themselves.

PR #194, opened 2021-09-12
fix enforced telemetry

playgithub:


playgithub:
Why can't the simple fix pass the checks?

mathisloge:
> Why can't the simple fix pass the checks?

Seems like the formatting is not correct. You can access the pipeline by clicking on the failed run and then on "View more details on Azure Pipelines" to access azure. Choose the run which failed and there you should see the error messages

src/vcpkg/vcpkglib.cpp
```diff
@@ -112,11 +112,8 @@ namespace vcpkg
             return; // File already in the new format
         }
 
-        if (!was_tracked)
-        {
-            was_tracked = true;
+        if (was_tracked)
```
ras0219-msft, line 115: This is a misunderstanding -- `was_tracked` is a boolean for whether this property has been submitted at least once. It is _not_ a setting for whether metrics are enabled or not.

The way metrics are intended to work in vcpkg is that the apis (`track_property`) are called unconditionally and whether they are actually sent is controlled centrally.

ras0219-msft (changes requested):
I believe this PR is based on a misunderstanding of the implementation of vcpkg -- could you please provide a repro that actually causes the unexpected behavior?

src/vcpkg/vcpkglib.cpp
playgithub: OK

If `was_tracked` is `true`,
`LockGuardPtr<Metrics>(g_metrics)->track_property("listfile", "update to new format");` will not be called.
Is it as expected?

PR #197, opened 2021-09-19
user_troubleshooting_message: Include git commit of the vcpkg repo and suggest updating with `git pull`

autoantwort:


strega-nil-ms:
Is there a reason we aren't just using `rev-parse HEAD`?

autoantwort:
Because I want to know the date. I want to know if the user is using a recent version of the vcpkg repo or an old one. And only the hash does not help (Yeah I can lookup the commit and check the date, but doing this for every issue is very painful).

dg0yt (commented):
How does it interact with custom registries or overlay ports? They may change the given port and/or dependencies.

autoantwort:
> How does it interact with custom registries or overlay ports? They may change the given port and/or dependencies.

Thats right, my ultimate goal is that vcpkg outputs a ready to copy paste github issue markdown code. In that you can then also put config files and other important information. The only problem currently is to catch the error logs. They are printed here:
https://github.com/microsoft/vcpkg/blob/827a54c404e67205262e6db032fb2c1a5e6fd097/scripts/cmake/vcpkg_execute_required_process.cmake#L132
I would simply search for the `See logs for more information` string but I assume @strega-nil-ms wants a guid based solution?

src/vcpkg/build.cpp
```diff
@@ -1315,22 +1315,26 @@ namespace vcpkg::Build
...
         auto vcpkg_update_cmd = ".\\vcpkg";
 #else
         auto vcpkg_update_cmd = "./vcpkg";
 #endif
-        return Strings::format("Please ensure you're using the latest portfiles with `%s update`, then\n"
+        auto description = paths.git_describe_head();
+        return Strings::format("Please ensure you're using the latest portfiles with `git pull` and `%s update`, then\n"
                                "submit an issue at https://github.com/Microsoft/vcpkg/issues including:\n"
                                "  Package: %s\n"
+                               "  Vcpkg-tool version: %s\n"
```
strega-nil-ms, line 1329: Also, this isn't _really_ related to the PR, but this should be (and the one below) `s/Vcpkg/vcpkg/`

autoantwort: I wanted to be consistent with the other lines. Should I then also change `Package` to `package`

strega-nil-ms: If you like; `vcpkg` is the correct title case for the vcpkg name. it should never be capitalized, even at the front of a sentence or in a title.

strega-nil-ms (approved):

autoantwort:
@BillyONeal can you do the second review ? :)

PR #199, opened 2021-09-26
Allow specification of msvc toolset version in triplet file

Chronial:
It is currently not possible to specify the minor version of the msvc toolset used to build ports. Vcpkg always chooses the newest installed version.

This is a problem for us, since we would like every developer in our team to build the same thing with the same tool. We also have a shared binary cache, which won't be used if a developer has a different "newest toolset" installed than the others.

This PR fixes that issue by introducing a `VCPKG_PLATFORM_TOOLSET_VERSION` variable to the triplet file. It can specify the version of the toolset to use to any precision. So toolset version `14.25.28610` would match a `VCPKG_PLATFORM_TOOLSET_VERSION` of `14`, `14.24` or `14.25.28610`.

This is only supported for Visual Studio >= 2017.

This fixes https://github.com/microsoft/vcpkg/issues/11317


src/vcpkg/vcpkgpaths.cpp
```diff
@@ -1057,45 +1057,53 @@ namespace vcpkg
...
+            error_message += " for toolset version " + *tsvf;
         }
 
-        Checks::check_exit(VCPKG_LINE_INFO, !candidates.empty(), "No suitable Visual Studio instances were found");
-        return *candidates.front();
+        Checks::check_exit(VCPKG_LINE_INFO, !candidates.empty(), error_message + ".");
 
+        if (tsv && vsp)
+        {
+            return *candidates.back();
```
Chronial, line 1101: This seems rather odd behavior – why would we want the oldest available version in this case, instead of the newest? This line does not introduce this behavior, but preserves it as it was before.

Chronial: Ah, I see, I missed that. What I don't see is why that assert should be here. And it isn't on my machine:
```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Tools\MSVC>dir /B /W
14.22.27905
14.25.28610
14.29.30037
```

Custom toolchain file `x64-windows-test`:
```
set(VCPKG_VISUAL_STUDIO_PATH [[C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional]])
set(VCPKG_PLATFORM_TOOLSET v142)
```

master vcpkg:
```
$ ./vcpkg.exe install zlib:x64-windows-test
    zlib[core]:x64-windows-ts -> 1.2.11#12
Error: vcpkg has crashed; no additional details are available.
The source line is D:\a\_work\1\s\src\vcpkg\vcpkgpaths.cpp(1076)
```

I assume it's fine if I don't add that crash back in. I'll just use `front()` in this case.

strega-nil-ms (approved):

ras0219-msft:
Thank you so much for this contribution and sorry for the long delays in merging!

PR #215, opened 2021-10-12
Localization PR

strega-nil:
v exciting

### DIFF
--- a/include/vcpkg/base/messages.h
+++ b/include/vcpkg/base/messages.h
@@ -138,11 +138,28 @@ namespace vcpkg::msg
         friend LocalizedString detail::internal_vformat(int index, fmt::format_args args);
     };
 
-    void write_unlocalized_text_to_stdout(Color c, StringView sv);
-    inline void write_newline_to_stdout()
+    struct LocalizedStringMapLess
     {
-        write_unlocalized_text_to_stdout(Color::None, "\n");
-    }
+        using is_transparent = void;
+        bool operator()(const LocalizedString& lhs, const LocalizedString& rhs) const
+        {
+            return lhs.data() < rhs.data();
+        }
+        bool operator()(LocalizedStringView lhs, LocalizedStringView rhs) const
+        {
+            return lhs.data() < rhs.data();
+        }
+        bool operator()(const LocalizedString& lhs, LocalizedStringView rhs) const
+        {
+            return lhs.data() < rhs.data();
+        }
+        bool operator()(LocalizedStringView lhs, const LocalizedString& rhs) const
+        {
+            return lhs.data() < rhs.data();
+        }
+    };
+
+    void write_unlocalized_text_to_stdout(Color c, StringView sv);
 
     template<class Message, class... Tags, class... Ts>
     LocalizedString format(Message, detail::MessageArgument<Tags, Ts>... args)
@@ -153,6 +170,11 @@ namespace vcpkg::msg
         return detail::internal_vformat(
             Message::index,
             fmt::make_format_args(fmt::arg(Tags::name(), *args.parameter)...));
+    }
+
+    inline void println()
+    {
+        write_unlocalized_text_to_stdout(Color::None, "\n");
     }
 
     inline void print(Color c, LocalizedStringView sv)
@@ -166,23 +188,23 @@ namespace vcpkg::msg
     inline void println(Color c, LocalizedStringView sv)
     {
         write_unlocalized_text_to_stdout(c, sv);
-        write_newline_to_stdout();
+        println();
     }
     inline void println(Color c, const LocalizedString& s)
     {
         write_unlocalized_text_to_stdout(c, s.data());
-        write_newline_to_stdout();
+        println();
     }
 
     inline void println(LocalizedStringView sv)
     {
         write_unlocalized_text_to_stdout(Color::None, sv);
-        write_newline_to_stdout();
+        println();
     }
     inline void println(const LocalizedString& s)
     {
         write_unlocalized_text_to_stdout(Color::None, s.data());
-        write_newline_to_stdout();
+        println();
     }
     inline void print(LocalizedStringView sv)
     {
--- a/src/vcpkg/base/system.cpp
+++ b/src/vcpkg/base/system.cpp
@@ -390,7 +390,7 @@ namespace vcpkg::Debug
         {
             msg::write_unlocalized_text_to_stdout(Color::None, "[DEBUG] ");
             msg::write_unlocalized_text_to_stdout(Color::None, sv);
-            msg::write_newline_to_stdout();
+            msg::println();
         }
     }
     void print(StringView sv)
--- a/include/vcpkg/base/expected.h
+++ b/include/vcpkg/base/expected.h
@@ -3,13 +3,15 @@
 #include <vcpkg/base/checks.h>
 #include <vcpkg/base/lineinfo.h>
 #include <vcpkg/base/stringliteral.h>
-#include <vcpkg/base/system.print.h>
+#include <vcpkg/base/messages.h>
 
 #include <system_error>
 #include <type_traits>
 
 namespace vcpkg
 {
+    DECLARE_MESSAGE(ExpectedValueWasError, (), "", "value was error");
+
     template<class Err>
     struct ErrorHolder
     {
@@ -24,7 +26,7 @@ namespace vcpkg
         const Err& error() const { return m_err; }
         Err& error() { return m_err; }
 
-        StringLiteral to_string() const { return "value was error"; }
+        msg::LocalizedString format() const { return msg::format(msgExpectedValueWasError); }
 
     private:
         bool m_is_error;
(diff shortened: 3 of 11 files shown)


CMakeLists.txt
```diff
@@ -121,6 +123,23 @@ endif()
 
 include(GNUInstallDirs)
 
+# ====================
+# === Dependencies ===
+# ====================
+
+include(FetchContent)
+FetchContent_Declare(
```
ras0219-msft, line 131: I know it's "documenting the standard library", but we should have a comment about how to point this at a pre-cloned copy of fmt.

src/vcpkg/base/downloads.cpp
```diff
@@ -476,13 +482,14 @@ namespace vcpkg::Downloads
     }
 #endif
 
+    DECLARE_AND_REGISTER_MESSAGE(DownloadBasicErrorWithInfo, (msg::url, msg::error), "{LOCKED}", "{url}: {error}");
```
ras0219-msft, line 485: Docs on what `"{LOCKED}"` means?

src/vcpkg/base/messages.cpp
```diff
@@ -0,0 +1,283 @@
...
+                   ? static_cast<DWORD>(-1)
+                   : static_cast<DWORD>(size);
+    }
+
+    void write_unlocalized_text_to_stdout(Color c, StringView sv)
+    {
+        if (sv.empty()) return;
+
+        auto handle = ::GetStdHandle(STD_OUTPUT_HANDLE);
+        if (is_console(handle))
```
ras0219-msft, line 39: This should be cached; we don't want to call `GetConsoleMode()` every time we print a message.

src/vcpkg/base/messages.cpp
```diff
@@ -0,0 +1,283 @@
...
+            {
+                DWORD written = 0;
+                check_write(::WriteFile(handle, pointer, size_to_write(size), &written, nullptr));
+                pointer += written;
+                size -= written;
+            }
+        }
+    }
+#else
+    static void write_all(int fd, const char* ptr, size_t to_write)
```
ras0219-msft, line 83:
```suggestion
    static void write_all(const char* ptr, size_t to_write)
```
The body assumes stdout, so it should not take a parameter.

src/vcpkg/base/messages.cpp
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    void write_unlocalized_text_to_stdout(Color c, StringView sv)
+    {
+        static constexpr char reset_color_sequence[] = {'\033', '[', '0', 'm'};
+
+        if (sv.empty()) return;
+
+        bool reset_color = false;
+        if (::isatty(STDOUT_FILENO) && c != Color::None)
```
ras0219-msft, line 105: If we are introducing tty detection, we need to give the user a way to override it (e.g. `--color` or `--no-color`).

src/vcpkg/base/messages.cpp
```diff
@@ -0,0 +1,283 @@
...
+            write_all(STDOUT_FILENO, reset_color_sequence, sizeof(reset_color_sequence));
+        }
+    }
+#endif
+
+    namespace
+    {
+        struct Messages
+        {
+            // this is basically a SoA - each index is:
```
ras0219-msft, line 126:
```suggestion
            // this is an SoA - each index is:
```

src/vcpkg/base/messages.cpp
```diff
@@ -0,0 +1,283 @@
...
+            // {
+            //   name
+            //   default_string
+            //   localization_comment
+            //   localized_string
+            // }
+            // requires: names.size() == default_strings.size() == localized_strings.size()
+            std::vector<StringLiteral> names;
```
ras0219-msft, line 134:
```suggestion
            std::vector<StringLiteral> names;  // const after startup
```

src/vcpkg/base/messages.cpp
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    void threadunsafe_initialize_context()
+    {
+        Messages& m = messages();
+        if (m.initialized)
+        {
+            write_unlocalized_text_to_stdout(Color::Error,
+                                 "double-initialized message context; this is a very serious bug in vcpkg\n");
+            Checks::exit_fail(VCPKG_LINE_INFO);
```
ras0219-msft, line 158:
```suggestion
            Checks::exit_with_message(VCPKG_LINE_INFO, "Attempted to double-initialized message context; this is a vcpkg internal bug");
```

src/vcpkg/base/messages.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+        Messages& m = messages();
+        std::vector<std::string> names_without_localization;
+
+        for (::size_t index = 0; index < m.names.size(); ++index)
+        {
+            const auto& name = m.names[index];
+            if (auto p = message_map.get(m.names[index]))
+            {
+                m.localized_strings[index] = p->string().to_string();
```
ras0219-msft, line 175: Needs error handling of `p` not being a string.

src/vcpkg/base/messages.cpp
```diff
@@ -0,0 +1,283 @@
...
+            }
+            else if (Debug::g_debugging)
+            {
+                // we only want to print these in debug
+                names_without_localization.push_back(name);
```
ras0219-msft, line 180: I don't think `--debug` should immediately imply "never localize anything". Maybe there is a need for a separate `--debug-locale` or some other method of getting this output.

include/vcpkg/base/messages.h
```diff
@@ -0,0 +1,259 @@
...
+        // REQUIRES: index < last_message_index()
+        StringView get_message_name(::size_t index);
+        // REQUIRES: index < last_message_index()
+        StringView get_default_format_string(::size_t index);
+        // REQUIRES: index < last_message_index()
+        StringView get_localization_comment(::size_t index);
+    }
+
+    // load from "locale_base/${language}.json"
+    void threadunsafe_initialize_context(const Filesystem& fs, StringView language, const Path& locale_base);
```
ras0219-msft, line 75: Noting that the user could set language to `C:\Users\My User\foo` and get that loaded as a language file.

src/vcpkg/base/messages.cpp
```diff
@@ -0,0 +1,283 @@
...
+        threadunsafe_initialize_context(message_map.first.object());
+    }
+
+    ::size_t detail::number_of_messages() { return messages().names.size(); }
+
+    ::size_t detail::startup_register_message(StringLiteral name, StringLiteral format_string, StringLiteral comment)
+    {
+        Messages& m = messages();
+
+        auto name_is_used = std::find(m.names.begin(), m.names.end(), name);
```
ras0219-msft, line 217: This is `O(N^2)` -- we should not be doing `O(N^2)` things on every startup. Instead, let's only check this when generating the localization file (which will be part of CI).

src/vcpkg/base/system.cpp
```diff
@@ -372,4 +372,22 @@ namespace vcpkg
...
 {
     std::atomic<bool> g_debugging(false);
+
+    void println(StringView sv)
+    {
+        if (g_debugging)
+        {
+            msg::write_text_to_stdout(Color::None, "[DEBUG] ");
+            msg::write_text_to_stdout(Color::None, sv);
+            msg::write_text_to_stdout(Color::None, "\n");
```
ras0219-msft, line 382:
```suggestion
            msg::write_text_to_stdout(Color::None, Strings::concat("[DEBUG] ", sv, "\n");
```

Message printing performance on Windows is _atrocious_ and in `--debug` mode combining these print calls will run significantly faster.

src/vcpkg/base/system.cpp
```diff
@@ -372,4 +372,22 @@ namespace vcpkg
...
+        }
+    }
+    void print(StringView sv)
+    {
+        if (g_debugging)
+        {
+            msg::write_text_to_stdout(Color::None, "[DEBUG] ");
```
ras0219-msft, line 389: Ditto as above.

src/vcpkg/binarycaching.cpp
```diff
@@ -1360,6 +1367,46 @@ namespace vcpkg
 
 namespace
 {
+    DECLARE_AND_REGISTER_MESSAGE(BcPathsMustBeAbsolute, (), "", "expected arguments: path arguments for binary config strings must be absolute");
```
ras0219-msft, line 1370: This really makes me wonder whether the regex based solutions are better; this does seems pretty awful for code review and maintenance being so far out of context.

src/vcpkg/base/messages.cpp
strega-nil: this is a misunderstanding of how this works.

src/vcpkg/binarycaching.cpp
strega-nil: I am not going to implement regex based solutions. I think it would be a major mistake.

PR #221, opened 2021-10-16
Update spdx license lists

autoantwort:
Update to latest master commit 0af22f869f8b0906097cfac90ee0516992e8939f


Generate-SpdxLicenseList.ps1
```diff
@@ -39,7 +39,9 @@ function Transform-JsonFile {
         "// Data downloaded from $Uri",
         "// Generated by scripts/Generate-SpdxLicenseList.ps1",
         "{")
-    $json.$OuterName | ForEach-Object {
+    $json.$OuterName |
+        Sort-Object -Property $Id |
```
strega-nil-ms, line 43:
```suggestion
        Sort-Object -Property $Id -Culture '' |
```

This needs to use the culture invariant culture.

PR #223, opened 2021-10-20
require exact version match for CMake and PowerShell

Be-ing:
CMake is included in the ABI info, so an exact version match is required
for binary caching to work for sharing a cache between machines with
different environments. Refer to the comment in the code for more
details.


Osyotr:
I can understand why CMake version is tracked, but why is PowerShell? I very much like the idea of having a switch to disable ABI tracking for these tools.

Be-ing:
I think adding another option would be the worst way this could be solved.

> I can understand why CMake version is tracked

Can you explain? I don't understand what that has to do with ABI.

Be-ing:
Could a maintainer please take a look at this? This is holding back Mixxx from switching from `vcpkg archive` to manifest mode, which we are trying to use so we can run Qt5 and Qt6 builds on the same branch while we transition to Qt6.

Osyotr:
> Can you explain? I don't understand what that has to do with ABI.

You are right, it does not. My guess is that it has something to do with how library is consumed (config file generated for specific version etc.)

Be-ing:
Pinging on this again. We have had several new contributors trying to build Mixxx on ARM macOS but not succeeding with the legacy `vcpkg archive` setup. We have had success building with vcpkg manifest mode on ARM macOS, but we cannot merge support for manifest mode until this issue with binary caching is resolved.

Be-ing:
Superseded by #234

PR #242, opened 2021-11-01
Add support for .a file check on Windows

JackBoosY:
Currently, we have allowed to build Linux static libraries through cross-compilation in Windows. This PR will add support for static library post check.
- [x] Add filtering code for .a files
- [ ] Add check ELF format code

Fixes https://github.com/microsoft/vcpkg/issues/21060


dg0yt:
#21060 is reporting mingw naming convention (`.dll.a`) on x64-windows. This is a valid port error, not a check issue. 
And please don't use the term "Linux static library" unless you mean linux binaries.

PR #248, opened 2021-11-03
Add new fields to vcpkg configuration

vicroms:
This PR adds new metadata to the top-level of the vcpkg configuration object.


src/vcpkg/configuration.cpp
```diff
@@ -53,21 +209,77 @@ namespace
...
     {
     }
 
+    static Json::Object serialize_configuration_impl(const Configuration& config)
+    {
+        constexpr static StringLiteral REGISTRY_PACKAGES = "packages";
+
+        auto serialize_packages_list = [](vcpkg::View<std::string> packages) {
+            Json::Array ret;
+            for (auto pkg : packages)
```
strega-nil-ms, line 233:
```suggestion
            for (const auto& pkg : packages)
```

strega-nil-ms: although I also think this helper function isn't super necessary

src/vcpkg/configuration.cpp
```diff
@@ -53,21 +209,77 @@ namespace
...
+        }
+
+        auto reg_view = config.registry_set.registries();
+        if (reg_view.size() > 0)
+        {
+            auto& reg_arr = obj.insert(ConfigurationDeserializer::REGISTRIES, Json::Array());
+            for (const auto& reg : reg_view)
+            {
+                auto reg_obj = reg.implementation().serialize();
+                reg_obj.insert(REGISTRY_PACKAGES, serialize_packages_list(reg.packages()));
```
strega-nil-ms, line 258:
```suggestion
                auto& packages = reg_obj.insert(REGISTRY_PACKAGES, Json::Array{});
                for (const auto& pkg : reg.packages())
                    packages.push_back(pkg);
```

something like this instead?

src/vcpkg/configuration.cpp
```diff
@@ -53,21 +209,77 @@ namespace
...
+                reg_arr.push_back(std::move(reg_obj));
+            }
+        }
+
+        for (const auto& el : config.ce_configuration)
```
strega-nil-ms, line 263: we may want to consider having a "standard order" for CE configuration as well.

src/vcpkg/registries.cpp
```diff
@@ -1013,6 +1019,38 @@ namespace
...
+    Json::Object obj{RegistryImplementation::serialize()};
+    obj.insert(RegistryImplDeserializer::BASELINE, Json::Value::string(m_baseline_identifier));
+    return obj;
+}
+
+Json::Object GitRegistry::serialize() const
+{
+    Json::Object obj{RegistryImplementation::serialize()};
+    obj.insert(RegistryImplDeserializer::BASELINE, Json::Value::string(m_baseline_identifier));
+    obj.insert(RegistryImplDeserializer::REPO, Json::Value::string(m_repo));
```
strega-nil-ms, line 1042: I think I'd prefer baseline to be the last field? What do others think?

autoantwort:
What are this for fields?

src/vcpkg/registries.cpp
vicroms: I agree with you

src/vcpkg/configuration.cpp
vicroms: Seems like a good idea, this is the order I prefer: 

```
{
  "default-registry": {},
  "registries": [],
  "message": "",
  "warning": "",
  "error": "",
  "settings": {},
  "apply": {},
  "requires": {},
  "see-also": {},
  "demands": {}
}
```

But I don't have any strong attachments to it. 
Except for one thing: since `demands` can recurse infinitely, I think they should be at the bottom.

BillyONeal (approved):
This rocks, thank you so much :D

PR #262, opened 2021-11-10
return verbose error when json contains object with duplicated keys

stforek:
Fixes https://github.com/microsoft/vcpkg/issues/20837.

It looks that parsing vcpkg.json may require more error handling.


vicroms (approved):

ras0219-msft (approved):

ras0219-msft:
Thanks for the PR!

PR #263, opened 2021-11-10
remove CMake and PowerShell versions from ABI info

Be-ing:
Neither of these actually affect the ABI. Including these in the ABI info is problematic because it requires using the exact same
versions of these tools on build servers as developers' local build environments to get binary caching to use cached packages
built on a server.

This is an alternative solution to https://github.com/microsoft/vcpkg/issues/16615. Unlike #223 it avoids vcpkg needing to download exact versions of CMake and PowerShell if a newer version than vcpkg's minimum requirement is already available from the system.

I don't particularly care whether this or #223 is merged; I just want some solution to https://github.com/microsoft/vcpkg/issues/16615.

### DIFF
--- a/src/vcpkg/build.cpp
+++ b/src/vcpkg/build.cpp
@@ -1087,12 +1087,6 @@ namespace vcpkg::Build
             }
         }
 
-        abi_tag_entries.emplace_back("cmake", paths.get_tool_version(Tools::CMAKE));
-
-#if defined(_WIN32)
-        abi_tag_entries.emplace_back("powershell", paths.get_tool_version("powershell-core"));
-#endif
-
         auto& helpers = paths.get_cmake_script_hashes();
         auto portfile_contents = fs.read_contents(port_dir / "portfile.cmake", VCPKG_LINE_INFO);
         for (auto&& helper : helpers)


ras0219-msft:
We've merged #223; we're hesitant to exclude CMake and PowerShell from the abi hashes because they have proven in the past to be significant in non-backwards-compatible ways.

Thanks for the PR! I'm closing since I believe that's the resolution you'd like based on the OP.

Be-ing:
I'm okay with this being closed but I don't see that #223 was merged?

Be-ing:
Oh I see that #234 was merged instead. I was not aware of that PR. I would have appreciated a comment on #223 notifying me there was a PR opened for an alternative solution.

ras0219-msft:
So sorry! I didn't look closely enough at the numbers. Does #234 address your concerns sufficiently?

Be-ing:
#234 does address my concerns but I still think it is not a great user experience to have to disable two options to get caching working as expected. Refer to my [comment on #234](https://github.com/microsoft/vcpkg-tool/pull/234#issuecomment-968411451) for more; let's continue discussing it there.

PR #266, opened 2021-11-11
Add pre-release tags and build metadata to relaxed versions

vicroms:
This PR makes the following changes:

- Cherry-picks @ras0219-msft changes  to remove regular expressions
- Merges logical parsing and comparing of relaxed and semantic versions
- Adds new tests for semver conformance

### DIFF
--- a/src/vcpkg-test/dependencies.cpp
+++ b/src/vcpkg-test/dependencies.cpp
@@ -123,7 +123,7 @@ static void check_name_and_version(const Dependencies::InstallPlanAction& ipa,
     }
 }
 
-static void check_semver_version(const ExpectedS<Versions::SemanticVersion>& maybe_version,
+static void check_semver_version(const ExpectedS<Versions::DotVersion>& maybe_version,
                                  const std::string& version_string,
                                  const std::string& prerelease_string,
                                  uint64_t major,
@@ -141,7 +141,7 @@ static void check_semver_version(const ExpectedS<Versions::SemanticVersion>& may
     CHECK(actual_version.identifiers == identifiers);
 }
 
-static void check_relaxed_version(const ExpectedS<Versions::RelaxedVersion>& maybe_version,
+static void check_relaxed_version(const ExpectedS<Versions::DotVersion>& maybe_version,
                                   const std::vector<uint64_t>& version)
 {
     auto actual_version = unwrap(maybe_version);
@@ -631,19 +631,19 @@ TEST_CASE ("version install diamond relaxed", "[versionplan]")
 
 TEST_CASE ("version parse semver", "[versionplan]")
 {
-    auto version_basic = Versions::SemanticVersion::from_string("1.2.3");
+    auto version_basic = Versions::semver_from_string("1.2.3");
     check_semver_version(version_basic, "1.2.3", "", 1, 2, 3, {});
 
-    auto version_simple_tag = Versions::SemanticVersion::from_string("1.0.0-alpha");
+    auto version_simple_tag = Versions::semver_from_string("1.0.0-alpha");
     check_semver_version(version_simple_tag, "1.0.0", "alpha", 1, 0, 0, {"alpha"});
 
-    auto version_alphanumeric_tag = Versions::SemanticVersion::from_string("1.0.0-0alpha0");
+    auto version_alphanumeric_tag = Versions::semver_from_string("1.0.0-0alpha0");
     check_semver_version(version_alphanumeric_tag, "1.0.0", "0alpha0", 1, 0, 0, {"0alpha0"});
 
-    auto version_complex_tag = Versions::SemanticVersion::from_string("1.0.0-alpha.1.0.0");
+    auto version_complex_tag = Versions::semver_from_string("1.0.0-alpha.1.0.0");
     check_semver_version(version_complex_tag, "1.0.0", "alpha.1.0.0", 1, 0, 0, {"alpha", "1", "0", "0"});
 
-    auto version_complexer_tag = Versions::SemanticVersion::from_string("1.0.0-alpha.1.x.y.z.0-alpha.0-beta.l-a-s-t");
+    auto version_complexer_tag = Versions::semver_from_string("1.0.0-alpha.1.x.y.z.0-alpha.0-beta.l-a-s-t");
     check_semver_version(version_complexer_tag,
                          "1.0.0",
                          "alpha.1.x.y.z.0-alpha.0-beta.l-a-s-t",
@@ -652,7 +652,7 @@ TEST_CASE ("version parse semver", "[versionplan]")
                          0,
                          {"alpha", "1", "x", "y", "z", "0-alpha", "0-beta", "l-a-s-t"});
 
-    auto version_ridiculous_tag = Versions::SemanticVersion::from_string("1.0.0----------------------------------");
+    auto version_ridiculous_tag = Versions::semver_from_string("1.0.0----------------------------------");
     check_semver_version(version_ridiculous_tag,
                          "1.0.0",
                          "---------------------------------",
@@ -661,44 +661,44 @@ TEST_CASE ("version parse semver", "[versionplan]")
                          0,
                          {"---------------------------------"});
 
-    auto version_build_tag = Versions::SemanticVersion::from_string("1.0.0+build");
+    auto version_build_tag = Versions::semver_from_string("1.0.0+build");
     check_semver_version(version_build_tag, "1.0.0", "", 1, 0, 0, {});
 
-    auto version_prerelease_build_tag = Versions::SemanticVersion::from_string("1.0.0-alpha+build");
+    auto version_prerelease_build_tag = Versions::semver_from_string("1.0.0-alpha+build");
     check_semver_version(version_prerelease_build_tag, "1.0.0", "alpha", 1, 0, 0, {"alpha"});
 
-    auto version_invalid_incomplete = Versions::SemanticVersion::from_string("1.0-alpha");
+    auto version_invalid_incomplete = Versions::semver_from_string("1.0-alpha");
     CHECK(!version_invalid_incomplete.has_value());
 
-    auto version_invalid_leading_zeroes = Versions::SemanticVersion::from_string("1.02.03-alpha+build");
+    auto version_invalid_leading_zeroes = Versions::semver_from_string("1.02.03-alpha+build");
     CHECK(!version_invalid_leading_zeroes.has_value());
 
-    auto version_invalid_leading_zeroes_in_tag = Versions::SemanticVersion::from_string("1.0.0-01");
+    auto version_invalid_leading_zeroes_in_tag = Versions::semver_from_string("1.0.0-01");
     CHECK(!version_invalid_leading_zeroes_in_tag.has_value());
 
-    auto version_invalid_characters = Versions::SemanticVersion::from_string("1.0.0-alpha#2");
+    auto version_invalid_characters = Versions::semver_from_string("1.0.0-alpha#2");
     CHECK(!version_invalid_characters.has_value());
 }
 
 TEST_CASE ("version parse relaxed", "[versionplan]")
 {
-    auto version_basic = Versions::RelaxedVersion::from_string("1.2.3");
+    auto version_basic = Versions::relaxed_from_string("1.2.3");
     check_relaxed_version(version_basic, {1, 2, 3});
 
-    auto version_short = Versions::RelaxedVersion::from_string("1");
+    auto version_short = Versions::relaxed_from_string("1");
     check_relaxed_version(version_short, {1});
 
     auto version_long =
-        Versions::RelaxedVersion::from_string("1.20.300.4000.50000.6000000.70000000.80000000.18446744073709551610");
+        Versions::relaxed_from_string("1.20.300.4000.50000.6000000.70000000.80000000.18446744073709551610");
     check_relaxed_version(version_long, {1, 20, 300, 4000, 50000, 6000000, 70000000, 80000000, 18446744073709551610u});
 
-    auto version_invalid_characters = Versions::RelaxedVersion::from_string("1.a.0");
+    auto version_invalid_characters = Versions::relaxed_from_string("1.a.0");
     CHECK(!version_invalid_characters.has_value());
 
-    auto version_invalid_identifiers_2 = Versions::RelaxedVersion::from_string("1.1a.2");
+    auto version_invalid_identifiers_2 = Versions::relaxed_from_string("1.1a.2");
     CHECK(!version_invalid_identifiers_2.has_value());
 
-    auto version_invalid_leading_zeroes = Versions::RelaxedVersion::from_string("01.002.003");
+    auto version_invalid_leading_zeroes = Versions::relaxed_from_string("01.002.003");
     CHECK(!version_invalid_leading_zeroes.has_value());
 }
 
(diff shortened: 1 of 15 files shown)


src/vcpkg-test/dependencies.cpp
```diff
@@ -707,10 +705,10 @@ TEST_CASE ("version parse relaxed", "[versionplan]")
 TEST_CASE ("version parse date", "[versionplan]")
 {
     auto version_basic = Versions::DateVersion::from_string("2020-12-25");
-    check_date_version(version_basic, "2020-12-25", "", {});
+    check_date_version(version_basic, "2020-12-25", {});
 
     auto version_identifiers = Versions::DateVersion::from_string("2020-12-25.1.2.3");
```
BillyONeal, line 710: Hmmm by our current docs this shouldn't parse, it doesn't meet the date contract. (But resolving things in favor of what the code does rather than what the docs say is probably reasonable)

src/vcpkg/versions.cpp
```diff
@@ -50,93 +51,185 @@ namespace vcpkg::Versions
         return hash<string>()(key.port_name) ^ (hash<string>()(key.version.to_string()) >> 1);
     }
 
-    ExpectedS<RelaxedVersion> RelaxedVersion::from_string(const std::string& str)
+    // 0|[1-9][0-9]*
+    static const char* parse_skip_number(const char* const s, uint64_t* const n)
```
BillyONeal, line 55: I observe all places calling this do something of the form:

```
x = fn(x, ...)
```

so you may want to consider making the first param an in_out parameter.

src/vcpkg/versions.cpp
```diff
@@ -50,93 +51,185 @@ namespace vcpkg::Versions
...
-                                  return as_numeric(strval).value_or_exit(VCPKG_LINE_INFO);
-                              })};
+        if (ch < '1' || ch > '9')
+        {
+            return nullptr;
+        }
+        size_t i = 1;
+        for (; Parse::ParserBase::is_ascii_digit(s[i]); ++i)
+        {
+        }
```
BillyONeal, line 70:
```suggestion
        while (Parse::ParserBase::is_ascii_digit(s[i]))
        {
            ++i;
        }
```

src/vcpkg/versions.cpp
```diff
@@ -50,93 +51,185 @@ namespace vcpkg::Versions
...
+        if (Parse::ParserBase::is_lower_alpha(ch) || Parse::ParserBase::is_upper_alpha(ch) || ch == '-')
         {
-            ret.prerelease_string = ret.version_string.substr(prerelease_found + 1);
-            ret.identifiers = Strings::split(ret.prerelease_string, '.');
-            ret.version_string.resize(prerelease_found);
+            // matched alpha identifier
+            ++cur;
+            for (; Parse::ParserBase::is_alphanumdash(*cur); ++cur)
+            {
+            }
```
BillyONeal, line 116:
```suggestion
            do
            {
                ++cur;
            } while(Parse::ParserBase::is_alphanumdash(*cur));
```

src/vcpkg/versions.cpp
```diff
@@ -50,93 +51,185 @@ namespace vcpkg::Versions
...
+        // Suggested regex by semver.org
+        // std::regex semver_scheme_match("^(0|[1-9][0-9]*)\\.(0|[1-9][0-9]*)\\.(0|[1-9][0-9]*)"
+        //                                "(?:-((?:0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9][0-9]*|[0-9]"
+        //                                "*[a-zA-Z-][0-9a-zA-Z-]*))*))?"
+        //                                "(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$");
+
+        SemanticVersion ret;
+        ret.original_string = str;
+        do
+        {
```
BillyONeal, line 143: Instead of this do-nothing-do in order to make break into poor-man's-goto, consider extracting a function so that you can use return for that purpose? e.g.:

```
bool from_string_impl(SemanticVersion& ret, const std::string& str) {
    ret.original_string = str;
    // the contents of the do loop go here
}

ExpectedS<SemanticVersion> SemanticVersion::from_string(const std::string& str) {
    SemanticVersion ret;
    if (from_string_impl(ret, str)) { return ret; }
    return Strings::format("Error: String `%s` is not a valid Semantic Version string, consult https://semver.org",
                                str);
}

File: src/vcpkg/versions.cpp [hunk 6]
```diff
@@ -50,93 +51,185 @@ namespace vcpkg::Versions
...
+            // pre-release
+            if (*cur == '-')
+            {
+                ++cur;
+                const char* const start_of_prerelease = cur;
+                for (;;)
+                {
+                    const auto start_identifier = cur;
+                    cur = skip_prerelease_identifier(cur);
+                    if (!cur) break;
```
BillyONeal, line 165: Here's a great example that would be made better with return since it has the cascaded seemingly meaningless extra if on 170.

src/vcpkg/versions.cpp
```diff
@@ -50,93 +51,185 @@ namespace vcpkg::Versions
...
+            if (*cur != '+') break;
+            ++cur;
+            // build
+            for (;;)
+            {
+                if (!Parse::ParserBase::is_alphanumdash(*cur)) break;
+                ++cur;
+                for (; Parse::ParserBase::is_alphanumdash(*cur); ++cur)
+                {
+                }
```
BillyONeal, line 184:
```suggestion
                do {
                    ++cur;
                } while(Parse::ParserBase::is_alphanumdash(*cur));
```

src/vcpkg/versions.cpp
```diff
@@ -50,93 +51,185 @@ namespace vcpkg::Versions
...
+            valid |= str[7] != '-';
+            valid |= Parse::ParserBase::is_ascii_digit(str[8]);
+            valid |= Parse::ParserBase::is_ascii_digit(str[9]);
+            if (!valid) break;
+            ret.version_string.assign(str.c_str(), 10);
+
+            const char* cur = str.c_str() + 10;
+            for (;;)
+            {
+                if (!cur || *cur != '.') break;
```
BillyONeal, line 220:
```suggestion
            while(cur && *cur == '.')
            {
```

src/vcpkg/versions.cpp
```diff
@@ -50,93 +51,185 @@ namespace vcpkg::Versions
...
-        return ret;
+        return Strings::format("Error: String `%s` is not a valid Semantic Version string, consult https://semver.org",
+                               str);
     }
 
     ExpectedS<DateVersion> DateVersion::from_string(const std::string& str)
     {
-        std::regex date_scheme_match("([0-9]{4}-[0-9]{2}-[0-9]{2})(\\.(0|[1-9][0-9]*))*");
-        if (!std::regex_match(str, date_scheme_match))
+        do
```
BillyONeal, line 198: Ditto poor-man's-goto

src/vcpkg/versions.cpp
```diff
@@ -50,93 +51,185 @@ namespace vcpkg::Versions
...
 
-    ExpectedS<SemanticVersion> SemanticVersion::from_string(const std::string& str)
+    ExpectedS<RelaxedVersion> RelaxedVersion::from_string(const std::string& str)
     {
-        // Suggested regex by semver.org
-        std::regex semver_scheme_match("^(0|[1-9][0-9]*)\\.(0|[1-9][0-9]*)\\.(0|[1-9][0-9]*)"
-                                       "(?:-((?:0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9][0-9]*|[0-9]"
-                                       "*[a-zA-Z-][0-9a-zA-Z-]*))*))?"
-                                       "(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$");
+        // (0|[1-9][0-9]*)(\.(0|[1-9][0-9]*))*;
```
BillyONeal, line 77: Is the regex supposed to end with a semicolon here?

src/vcpkg/versions.cpp
```diff
@@ -50,93 +51,185 @@ namespace vcpkg::Versions
...
+            cur = parse_skip_number(cur, &numbers.back());
+            if (!cur || (*cur != 0 && *cur != '.'))
+            {
+                return Strings::format(
+                    "Error: String `%s` must only contain dot-separated numeric values without leading zeroes.", str);
+            }
+            if (*cur == 0)
+            {
+                break;
+            }
```
BillyONeal, line 93:
```suggestion
            if (cur) {
              if (*cur == 0) { break; }
              if (*cur == '.') {
              ++cur; continue;
              }
            }
            return Strings::format(
                    "Error: String `%s` must only contain dot-separated numeric values without leading zeroes.", str);
```

?

src/vcpkg/versions.cpp
```diff
@@ -50,93 +51,185 @@ namespace vcpkg::Versions
...
+    }
 
-        auto build_found = ret.version_string.find('+');
-        if (build_found != std::string::npos)
+    // 0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*
+    // equivalent:
+    static const char* skip_prerelease_identifier(const char* const s)
+    {
+        auto cur = s;
+        for (; Parse::ParserBase::is_ascii_digit(*cur); ++cur)
```
BillyONeal, line 106:
```suggestion
        while (Parse::ParserBase::is_ascii_digit(*cur)) { ++cur; }
```

src/vcpkg/versions.cpp
vicroms: Probably not, the parsing changes are cherry-picked from Robert's PR to speed-up search.
I think the `;` is a left-over from commenting a call to `std::regex_match()`.

src/vcpkg-test/dependencies.cpp
vicroms: I remember these were part of the initial spec draft and at some point even allowed for things like `2020-12-25.nightly` and `2020-12-25-nightly`,  but we decided on only allowing numeric identifiers after `.` to solve the sort behavior problem. If we decided to drop them entirely it was probably after the code had been implemented and I never got around to remove it.

src/vcpkg/versions.cpp
ras0219-msft: Yep, that's my bad.

src/vcpkg/versions.cpp
ras0219-msft: It's funny, I implemented all of these as for loops because I _thought_ I remembered you saying you preferred
`for (;;) {}` over `while (1) {}`.

src/vcpkg/versions.cpp
ras0219-msft: Imho it would be _much_ cleaner to just use `goto`, but I figured it would be less taxing on code review to avoid it.

src/vcpkg/versions.cpp
ras0219-msft: I felt this was cleaner due to reducing indirection (ret value + in param + out param instead of ret value + in_out param)

ras0219-msft:
@vicroms I pushed changes to your branch to address @BillyONeal's comments on the cherry-picked version parser. I addressed the do-nothing-do loops by transforming them into proper goto's (the original design) -- if strongly desired I can extract them into functions, but I think this is the clearest and simplest implementation.

BillyONeal:
I prefer for for infinite loops but these aren’t.

On Nov 13, 2021, at 11:35 AM, Robert Schumacher ***@***.***> wrote:

﻿

@ras0219-msft commented on this pull request.

________________________________

In src/vcpkg/versions.cpp<https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fmicrosoft%2Fvcpkg-tool%2Fpull%2F266%23discussion_r748758111&data=04%7C01%7Cbion%40microsoft.com%7Ca4244cec817440159cb608d9a6dcb96c%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637724289212820849%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=pT4%2B56TgasPx71IJpnDl8BkT%2Fvs%2FasuAnYUBFQYR3LY%3D&reserved=0>:

>
-        auto build_found = ret.version_string.find('+');
-        if (build_found != std::string::npos)
+    // 0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*
+    // equivalent:
+    static const char* skip_prerelease_identifier(const char* const s)
+    {
+        auto cur = s;
+        for (; Parse::ParserBase::is_ascii_digit(*cur); ++cur)


It's funny, I implemented all of these as for loops because I thought I remembered you saying you preferred
for (;;) {} over while (1) {}.

vicroms:
Thanks @ras0219-msft!

After some investigation I noticed that we can safely merge semver and relaxed into the same class without much pain. It is the test code that will be the most affected.

vicroms:
Intentionally pushed a failing test in order to discuss what the correct output should be.

src/vcpkg-test/dependencies.cpp
```diff
@@ -663,54 +663,56 @@ TEST_CASE ("version parse semver", "[versionplan]")
...
 
-    auto version_invalid_characters = Versions::RelaxedVersion::from_string("1.a.0");
-    CHECK(!version_invalid_characters.has_value());
+    check_relaxed_version(Versions::relaxed_from_string("1.0.0-alpha"), {1, 0, 0}, {"alpha"});
 
-    auto version_invalid_identifiers_2 = Versions::RelaxedVersion::from_string("1.1a.2");
-    CHECK(!version_invalid_identifiers_2.has_value());
+    // TODO: Currently this test fails, the produced identifiers are { "alpha-0", "1" }
+    // but my understanding is that the output should be as in this test.
+    check_relaxed_version(Versions::relaxed_from_string("1.0.0-alpha-0.1"), {1, 0, 0}, {"alpha", "0", "1"});
```
vicroms, line 699: @ras0219-msft @BillyONeal 

I intentionally pushed this failing test so we can discuss what the correct output should be.

src/vcpkg-test/dependencies.cpp
```diff
@@ -727,27 +729,35 @@ TEST_CASE ("version parse date", "[versionplan]")
...
     std::sort(std::begin(versions), std::end(versions), [](const auto& lhs, const auto& rhs) -> bool {
         return Versions::compare(lhs, rhs) == Versions::VerComp::lt;
     });
 
+    std::string str;
+    for (auto v : versions)
+    {
+        Strings::append(str, v.original_string + "\n");
+    }
+
```
vicroms, line 760:
```suggestion
```

src/vcpkg-test/dependencies.cpp
vicroms: It seems that the output was correct all along. 
See https://github.com/semver/semver/issues/78

```suggestion
    check_relaxed_version(Versions::relaxed_from_string("1.0.0-alpha-0.1"), {1, 0, 0}, {"alpha-0", "1"});
```

ras0219-msft (approved):
I took the opportunity to extract the "lexographical spaceship" algorithm.

src/vcpkg/versions.cpp
```diff
@@ -50,93 +49,171 @@ namespace vcpkg::Versions
...
+                ++cur;
+            }
+            if (*cur == 0) return ret;
+            if (*cur == '.')
+            {
+                ++cur;
+            }
+            else
+            {
+                goto invalid_semver;
```
BillyONeal, line 170:
```suggestion
                return nullopt;
```

Ditto the others. I'm not opposed to goto in concept but it seems to not be adding value here.

src/vcpkg/versions.cpp
```diff
@@ -50,93 +49,171 @@ namespace vcpkg::Versions
...
+            }
+            if (*cur != 0) goto invalid_date;
+
+            return ret;
         }
 
-        return ret;
+    invalid_date:
+
+        return Strings::format("Error: String `%s` is not a valid date version."
```
BillyONeal, line 213: How about extract this as `std::string format_invalid_date_message(const std::string& str)`, then the returns can become `return format_invalid_date_message(str);` and then you don't need goto.

Or do a transform like done with `relaxed_from_string` already where there's a wrapping function.

BillyONeal (approved):
The gotos are better than cascaded breaks but I think it would still be better to make them returns.

ras0219-msft:
@vicroms I'm going ahead and pushing merge because I want to resolve conflicts with https://github.com/microsoft/vcpkg-tool/pull/265

PR #267, opened 2021-11-12
Support media queries binary logical operator syntax in platform expressions

markle11m:
This change implements support in platform expressions for binary logical operators (and, or) as defined by [Media Queries](https://www.w3.org/TR/css3-mediaqueries/#syntax).

We are extending current platform expressions as follows:
* for logical-and, 'and' is a synonym for '&'
* for logical-or, ',' is accepted, but has lower precedence than logical-and
  * currently using both '&' and '|' in the same expression is disallowed unless parens are used to identify evaluation precedence; that constraint is preserved, but '&', 'and', and ',' can be freely mixed in a platform expression without the need for parens

The existing platform expression parser:
* does a left-to-right linear scan of the expression string
* assumes that logical operators are individual chars ('!', '&', '|')
* consumption/processing of binary logical operators is split between multiple methods (expr(), expr_binary()) and tightly couple with assumptions about valid syntax
* does not create a binary expression tree
  * instead expr_binary() gathers consecutive instances of the same operation into the expression node (e.g., "A & B & C" yields <op_and, vector<A,B,C>>)
* expr_binary() support backwards compatibility for chained vcpkg operators '&' and '|' (e.g., '&&&') 
* expr_binary() also detects the error case of mixing '&' and '|'
* operator precedence is supported only explicitly (using parens, e.g., '(A & B) | C')

To support the media queries syntax, the following changes have been made:
* moved all processing of logical-operators to a new function expr_operator()
  * removed the assumption that a logical operator is represented by an individual char (e.g., 'and')
  * added parsing support for 'and' and ','
* added support for implicit operator precedence (the low-precedence or-operator ',')

Implementation details:
* Added 16 unit tests for platform expressions, covering both existing and new syntax
* Extended ExprKind to support low_precedence_or, empty and invalid operators to represent a more complete set of parse states after scanning for an expected operator
* Added expr_operator() to extract the next operator from the input string. This is a destructive operation as it removes the chars parsed from the input text.
  * backwards compatible support for chained vcpkg operators '&' and '|' moved here from expr_binary()
  * expr_operator adds support for parsing non-single-char operators (e.g., "and")
    * note: 'and' is odd in that it's multiple chars and looks like an identifier, so it required special handling
* Updated inaccurate error messages
* Updated expr_binary() to template on ExprKind instead of char to support new operators and restrictions 
  * chained vcpkg operator handling moved to expr_operator()
  * process ',' as 'platform_expression , platform_expression' to handle precedence order (as opposed to '|', which is 
    implemented as 'plaform_expression_not | platform_expression_not'
  * consecutive instance gathering is supported for 'and' and ','

### DIFF
--- a/src/vcpkg-test/platform-expression.cpp
+++ b/src/vcpkg-test/platform-expression.cpp
@@ -421,6 +421,17 @@ TEST_CASE ("platform-expressions without whitespace", "[platform-expression]")
     CHECK_FALSE(expr.evaluate({{"VCPKG_CMAKE_SYSTEM_NAME", "Linux"}, {"VCPKG_TARGET_ARCHITECTURE", "arm64"}}));
 }
 
+TEST_CASE ("ambiguities without whitespace", "[platform-expression]")
+{
+    // Operator keywords ("and","not") require a break to separate them from identifiers
+    // In these cases, strings containing an operator keyword parse as an identifier, not as a unary/binary expression
+    auto m_expr = parse_expr("!windowsandandroid");
+    CHECK(m_expr);
+
+    m_expr = parse_expr("notwindows");
+    CHECK(m_expr);
+}
+
 TEST_CASE ("invalid logic expression, unexpected character", "[platform-expression]")
 {
     auto m_expr = parse_expr("windows arm");
--- a/src/vcpkg-test/platform-expression.cpp
+++ b/src/vcpkg-test/platform-expression.cpp
@@ -340,6 +340,43 @@ TEST_CASE ("weird platform-expressions whitespace", "[platform-expression]")
     CHECK(expr.evaluate({{"VCPKG_CMAKE_SYSTEM_NAME", "Linux"}}));
 }
 
+TEST_CASE ("platform-expressions no whitespace", "[platform-expression]")
+{
+    auto m_expr = parse_expr("!windows|linux|arm");
+    REQUIRE(m_expr);
+    auto& expr = *m_expr.get();
+
+    CHECK_FALSE(expr.evaluate({{"VCPKG_CMAKE_SYSTEM_NAME", ""}}));
+    CHECK(expr.evaluate({{"VCPKG_CMAKE_SYSTEM_NAME", ""}, {"VCPKG_TARGET_ARCHITECTURE", "arm"}}));
+    CHECK(expr.evaluate({{"VCPKG_CMAKE_SYSTEM_NAME", "Linux"}}));
+
+    m_expr = parse_expr("!windows&!arm&!x86");
+    REQUIRE(m_expr);
+    expr = *m_expr.get();
+
+    CHECK_FALSE(expr.evaluate({{"VCPKG_CMAKE_SYSTEM_NAME", ""}}));
+    CHECK_FALSE(expr.evaluate({{"VCPKG_CMAKE_SYSTEM_NAME", "WindowsStore"}}));
+    CHECK(expr.evaluate({{"VCPKG_CMAKE_SYSTEM_NAME", "Linux"}}));
+    CHECK_FALSE(expr.evaluate({
+        {"VCPKG_CMAKE_SYSTEM_NAME", "Linux"},
+        {"VCPKG_TARGET_ARCHITECTURE", "arm"},
+    }));
+
+    m_expr = parse_expr("windows,!arm&linux&(x86|x64)");
+    REQUIRE(m_expr);
+    expr = *m_expr.get();
+
+    CHECK(expr.evaluate({{"VCPKG_CMAKE_SYSTEM_NAME", ""}}));
+    CHECK(expr.evaluate({{"VCPKG_CMAKE_SYSTEM_NAME", "WindowsStore"}}));
+    CHECK_FALSE(expr.evaluate({{"VCPKG_CMAKE_SYSTEM_NAME", "Linux"}}));
+    CHECK_FALSE(expr.evaluate({{"VCPKG_CMAKE_SYSTEM_NAME", "Darwin"}}));
+    CHECK_FALSE(expr.evaluate({{"VCPKG_TARGET_ARCHITECTURE", ""}}));
+    CHECK(expr.evaluate({{"VCPKG_CMAKE_SYSTEM_NAME", "Linux"}, {"VCPKG_TARGET_ARCHITECTURE", "x86"}}));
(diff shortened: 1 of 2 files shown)


src/vcpkg/platform-expression.cpp
```diff
@@ -139,26 +142,114 @@ namespace vcpkg::PlatformExpression
             // platform-expression =
             // | platform-expression-not
             // | platform-expression-and
-            // | platform-expression-or ;
+            // | platform-expression-or
+            // | platform-expression-low-precedence-or ;
```
strega-nil, line 146: We need to write the real syntax; this isn't fully correct.

strega-nil (changes requested):

src/vcpkg/platform-expression.cpp
strega-nil: (probably in vcpkg)

markle11m: Thanks. I will incorporate your suggested changes (in vcpkg) and update this code.

strega-nil: thanks mark!

src/vcpkg-test/platform-expression.cpp
```diff
@@ -152,6 +152,48 @@ TEST_CASE ("platform-expression-and", "[platform-expression]")
...
+    CHECK(expr.evaluate({{"VCPKG_CMAKE_SYSTEM_NAME", "Linux"}}));
+    CHECK_FALSE(expr.evaluate({
+        {"VCPKG_CMAKE_SYSTEM_NAME", "Linux"},
+        {"VCPKG_TARGET_ARCHITECTURE", "arm"},
+    }));
+}
+
+TEST_CASE ("platform-expression-and-multiple", "[platform-expression]")
+{
+    auto m_expr = parse_expr("!windows & !arm & !x86");
```
ras0219-msft, line 172: I don't see any tests for "minimum whitespace" examples:
`windows&arm`
`windows,arm&!linux`
etc.
These are supported today (though `format-manifest` inserts spaces)

src/vcpkg/platform-expression.cpp
```diff
@@ -139,26 +142,114 @@ namespace vcpkg::PlatformExpression
...
+                        // but has lower precedence
+                        next();
+                        return ExprKind::op_low_precedence_or;
+                    }
+                    case 'a':
+                    {
+                        // { "and", optional-whitespace, platform-expression-not }
+                        // "and" is a synonym of "&"
+                        std::string name = match_zero_or_more(is_identifier_char).to_string();
+                        if (!name.empty() && (name == "and"))
```
ras0219-msft, line 233:
```suggestion
                        if (name == "and")
```
`== "and"` implies `!empty()`

src/vcpkg/platform-expression.cpp
```diff
@@ -139,26 +142,114 @@ namespace vcpkg::PlatformExpression
...
+                        {
+                            return ExprKind::op_and;
+                        }
+
+                        // Invalid alphanumeric strings or strings other than "and" are errors.
+                        if (name.empty())
+                        {
+                            add_error("unexpected character in logic expression");
```
ras0219-msft, line 241:
```suggestion
                            Checks::unreachable(VCPKG_LINE_INFO);
```
Because we matched an 'a' in the case statement, it should be impossible for `name` to be empty (if I'm reading the code correctly).

src/vcpkg-test/platform-expression.cpp
markle11m: Good catch. Added minimum whitespace test.

src/vcpkg/platform-expression.cpp
markle11m: You are correct. At that point in the code, cur() is 'a' and based on the logic in match_zero_or_more(), at least the 'a' must be returned. I will modify the code to remove the unnecessasry name.empty() checks.

src/vcpkg/platform-expression.cpp
markle11m: after a few iterations, proposed EBNF update is [posted](https://github.com/microsoft/vcpkg/pull/21444), additional tests added and comments updated

src/vcpkg-test/platform-expression.cpp
```diff
@@ -174,10 +376,194 @@ TEST_CASE ("weird platform-expressions whitespace", "[platform-expression]")
...
+    m_expr = parse_expr("windows andMORE arm");
+    CHECK_FALSE(m_expr);
+    m_expr = parse_expr("windows and+ arm");
+    CHECK_FALSE(m_expr);
+    m_expr = parse_expr("windows and& arm");
+    CHECK_FALSE(m_expr);
+
+    m_expr = parse_expr("notANY windows");
+    CHECK_FALSE(m_expr);
+    m_expr = parse_expr("not! windows");
```
ras0219-msft, line 565: Why is this one invalid? Shouldn't this parse as `!!windows`?

src/vcpkg/platform-expression.cpp
```diff
@@ -223,43 +308,97 @@ namespace vcpkg::PlatformExpression
...
+                {
+                    std::string name = match_zero_or_more(is_identifier_char).to_string();
+
+                    // "not"
+                    if (name == "not")
+                    {
+                        // optional-whitespace,
+                        skip_whitespace();
+                        // platform-expression-simple
+                        return std::make_unique<ExprImpl>(ExprKind::op_not, expr_simple());
```
ras0219-msft, line 321: I think this disagrees with the EBNF above -- it says the continuation should be `platform-expression-unary-keyword-operand`, not `platform-expression-simple`

src/vcpkg-test/platform-expression.cpp
markle11m: That's a good question. At first glance, one would expect `not! windows` to be the same as `!!windows` and evaluate to windows. But the code currently fails to part both of those expressions. It also doesn't appear to be a valid parse according to the EBNF. 
So I think you're right and this should be valid, but we'll need to update the grammar when we do.

src/vcpkg/platform-expression.cpp
markle11m: You are correct. I have fixed the comments.

ras0219-msft (approved):

PR #270, opened 2021-11-13
Implement bundle.usegitregistry (continuation of read-only vcpkg)

ras0219-msft:
This bundle setting replaces the `BuiltinRegistry` with a `GitRegistry` pointed at `github.com/Microsoft/vcpkg`, enabling functional tool distribution without requiring the entire port history.

Some additional implementation notes:
- I enabled e2e tests outside the registries.ps1 suite to use the user cache -- these e2e tests are not testing the registries code, they are testing other behaviors and shouldn't need to constantly purge and re-download git repos.
- Management of `"builtin-baseline"` has been moved into `load_configuration()`, since it's morally the same as `"vcpkg-configuration"`
- This mode _requires_ a configured baseline even for commands like "search". Future work should address this.


BillyONeal (approved):

ras0219-msft:
Thanks!

PR #272, opened 2021-11-16
[vcpkg-hash] use current working directory to get a file

srgbtl:
Fixes: https://github.com/microsoft/vcpkg/issues/4791

### DIFF
--- a/src/vcpkg/commands.hash.cpp
+++ b/src/vcpkg/commands.hash.cpp
@@ -20,8 +20,7 @@ namespace vcpkg::Commands::Hash
     {
         (void)args.parse_arguments(COMMAND_STRUCTURE);
 
-        Path file_to_hash = args.command_arguments[0];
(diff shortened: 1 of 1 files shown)


src/vcpkg/commands.hash.cpp
```diff
@@ -20,7 +20,8 @@ namespace vcpkg::Commands::Hash
     {
         (void)args.parse_arguments(COMMAND_STRUCTURE);
 
-        const Path file_to_hash = args.command_arguments[0];
+        Path file_to_hash = args.command_arguments[0];
+        if (file_to_hash.is_relative()) file_to_hash = (paths.original_cwd / file_to_hash).lexically_normal();
```
ras0219-msft, line 24:
```suggestion
        auto file_to_hash = (paths.original_cwd / args.command_arguments[0]).lexically_normal();
```

`Path::operator/` does nothing if the right hand side is absolute, so no need to guard against it.

ras0219-msft (changes requested):

ras0219-msft:
This LGTM, thanks for the fix!

PR #273, opened 2021-11-17
Add 'embeddedsha' to record tested baseline for bundle

ras0219-msft:
The primary change in this PR is very small: Add `$bundle.embeddedsha` to enable read-only vcpkg to provide useful error messages around baselines.

While investigating the general problem, I discovered several other functions that were generally 'unsafe' (they depend upon `root / ".git"`), but in the single context they're used are safe. I therefore refactored things to move those functions closer to their contexts and out of the public `VcpkgPaths::` interface.

### DIFF
--- a/include/vcpkg/vcpkgpaths.h
+++ b/include/vcpkg/vcpkgpaths.h
@@ -115,7 +115,11 @@ namespace vcpkg
         Path community_triplets;
         Path scripts;
         Path prefab;
+
+    private:
         Path builtin_ports;
+
+    public:
         Path builtin_registry_versions;
 
         Path tools;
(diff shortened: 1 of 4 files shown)


autoantwort:
One main aspect of #197 was to get the date of the used vcpkg version for issues. You now always see in the issues how old the vcpkg version is they used to build a port (for example `vcpkg version: 1a5075b 2021-11-17 (2 hours ago)`). This is very helpful if people are using a (very) old version of vcpkg and you know the failure was fixed in a new version. 

This feature gets removed with this PR

ras0219-msft:
I see, I assumed it was primarily about the commit id. Let me see about adding date infromation.

BillyONeal (approved):
As long as @autoantwort is happy I'm happy

ras0219-msft:
I've restored the date display behavior for existing cases, however the read-only bundle probably needs a different treatment (the date information needs to come from somewhere else).

PR #282, opened 2021-11-22
[platform-expression] fix compare()

maksqwe:


ras0219-msft (approved):

ras0219-msft:
Thanks for the fix!

What bad behavior caused you to identify this bug? We'd like to add a test case to improve coverage.

maksqwe:
Just saw this typo in the code.
Test case: when do some search using "vcpkg search ..." this function compare two expressions, for example in ffmpeg vcpkg.json there are platforms which return wrong result from compare().

```JSON
{
  "name": "ffmpeg",
  "default-features": false,
  "features": [
    "opengl"
  ],
  "platform": "!uwp & !(windows & arm)"
},
{
  "name": "ffmpeg",
  "default-features": false,
  "features": [
    "dav1d"
  ],
  "platform": "!(uwp | arm | x86 | osx)"
},
```
Complexy of `!uwp & !(windows & arm)` is 7, when `!(uwp | arm | x86 | osx)` is 6 and has bigger length.

PR #286, opened 2021-11-29
[hash] Downgrade Hash to a BasicCommand

ras0219-msft:


BillyONeal (approved):
Approved once the test that asserts about which commands exist is fixed.

PR #290, opened 2021-12-01
Fix the cmake's output for vcpkg_get_dep_info()

tlemo:
Set `CLICOLOR=0` to force cmake to output "plain" text (without ANSI escape codes)
(see [Terminal.c](https://github.com/Kitware/CMake/blob/29c80aec94b3607f3aae82170f0b3d0804de2aea/Source/kwsys/Terminal.c#L176-L182))

On Linux (and likely on other non-Windows platforms), CMake may "colorize" the output depending on the terminal type. This creates problems when `vcpkg` attempts to parse the output from CMake ([ex](https://github.com/microsoft/vcpkg-tool/blob/1b7edfa7196228006613d96657492e78d3e22043/src/vcpkg/cmakevars.cpp#L260-L264))

This PR would fix https://github.com/microsoft/vcpkg/issues/20430

### DIFF
--- a/src/vcpkg.cpp
+++ b/src/vcpkg.cpp
@@ -187,7 +187,13 @@ int main(const int argc, const char* const* const argv)
 #endif
     set_environment_variable("VCPKG_COMMAND", get_exe_path_of_current_process().generic_u8string());
     
-    // See http://bixense.com/clicolors
+    // Prevent child processes (ex. cmake) from producing "colorized"
+    // output (which may include ANSI escape codes), since it would
+    // complicate parsing the output.
+    //
+    // See http://bixense.com/clicolors for the semantics associated with
+    // the CLICOLOR and CLICOLOR_FORCE env variables
+    //
     set_environment_variable("CLICOLOR_FORCE", {});
     set_environment_variable("CLICOLOR", "0");
 
--- a/src/vcpkg.cpp
+++ b/src/vcpkg.cpp
@@ -186,7 +186,7 @@ int main(const int argc, const char* const* const argv)
     }
 #endif
     set_environment_variable("VCPKG_COMMAND", get_exe_path_of_current_process().generic_u8string());
-    
+
     // Prevent child processes (ex. cmake) from producing "colorized"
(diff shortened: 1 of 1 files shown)


dg0yt:
It looks inconsistent to unset one variable and to set the other to '0'.

I still think that the right fix is to not use console output for the cmake output which is meant to be parsed, but write a regular file instead. This would allow cmake to use color as desired by the user for other output.

dg0yt:
I guess the test for CLICOLOR didn't work because vcpkg CI is not really a terminal. CLICOLOR_FORCE makes cmake ignore the terminal property. So the problem is valid, and this PR is a possible solution.

tlemo:
> Semi-duplicate of #261 but there dg0yt tried to write a test to repro the problem and could not

Constructing a repro should be straightforward. The following conditions can trigger the issue:
1. A project using the vcpkg manifest
2. Non-Windows platform + a color supporting terminal
3. Use `make` as generator (by default, cmake would attempt to "colorize" the output)
4. After a cmake configure + build, modify a cmake script than run `make` 

The last step will trigger cmake configure, which would invoke the vcpkg installation, which should hit the issue.

> It looks inconsistent to unset one variable and to set the other to '0'.

I agree, it does seem a bit inconsistent, although the inconsistency is external - that's the logic that CMake implements:`CLICOLOR_FORCE` can force colorization on (but not off), and `CLICOLOR=1` can force it off (but not on). There may be some history behind this convoluted logic.

Arguably, there are some better fixes: redirect cmake output to a file instead of a pipe (like @dg0yt suggested) or set the `CLICOLOR=0` just for the child processes (not vckpg itself). The latter seems to require additional plumbing to add support for exec with env variables on all platforms, which doesn't seem to be there.

Ultimately, the fix proposed here is simple and safe and would eliminate a jarring behavior when vcpkg manifests are used on Linux. It doesn't prevent a more elaborate fix in the future.

ras0219-msft (approved):
LGTM

ras0219-msft:
LGTM, thanks!

PR #293, opened 2021-12-04
Adding support for AWS S3 binary caching

dave-juicelabs:


dave-juicelabs:
While this is being reviewed, is there a way for me to use this in my own personal builds? Since vcpkg now downloads a built version, I am unsure of how I would be able to integrate this.

mathisloge:
> While this is being reviewed, is there a way for me to use this in my own personal builds? Since vcpkg now downloads a built version, I am unsure of how I would be able to integrate this.

you can just replace the downloaded version of vcpkg with your built version

src/vcpkg/binarycaching.cpp
```diff
@@ -1120,6 +1120,186 @@ namespace
...
+    {
+        Command cmd;
+        cmd.string_arg("aws").string_arg("s3").string_arg("cp").string_arg(aws_object).path_arg(archive);
+        const auto out = cmd_execute_and_capture_output(cmd);
+        if (out.exit_code == 0)
+        {
+            return true;
+        }
+
+        print2(Color::warning, "aws failed to download with exit code: ", out.exit_code, '\n', out.output);
```
strega-nil-ms, line 1156: This should be using the new localized output; at the top of the file:

```cxx
DECLARE_AND_REGISTER_MESSAGE(AwsFailedToDownload, (msg::value, msg::output), "aws failed to download with exit code: {value}\n{output}");
```

(then add a new `DECLARE_MSG_ARG` in messages.h for `output`)

Then, replace this with:

```suggestion
        msg::println(Color::warning, msgAwsFailedToDownload, msg::value = out.exit_code, msg::output = out.output);
```

src/vcpkg/binarycaching.cpp
```diff
@@ -1120,6 +1120,186 @@ namespace
...
+
+                    clean_prepare_dir(fs, paths.package_dir(action.spec));
+                    url_paths.emplace_back(make_aws_path(prefix, *abi),
+                                           make_temp_archive_path(paths.buildtrees(), action.spec));
+                    url_indices.push_back(idx);
+                }
+
+                if (url_paths.empty()) break;
+
+                print2("Attempting to fetch ", url_paths.size(), " packages from AWS.\n");
```
strega-nil-ms, line 1203: same as above - replace with localized output.

src/vcpkg/tools.cpp
```diff
@@ -567,6 +567,35 @@ gsutil version: 4.58
         }
     };
 
+    struct AWSCliProvider : ToolProvider
```
strega-nil-ms, line 570:
```suggestion
    struct AwsCliProvider : ToolProvider
```

Acronyms should be treated as a single word for the purposes of camel case, like in [rust].

[rust]: https://rust-lang.github.io/api-guidelines/naming.html

strega-nil-ms (changes requested):
Some minor changes, but otherwise, this looks great :)

strega-nil-ms:
cc @ras0219-msft as the owner of binary caching.

dave-juicelabs:
Ahh, yeah that looks a lot more reasonable.

Pospelove:
Can't wait for AWS S3 support :)

ras0219-msft (approved):
I believe this also will require adding the AWS CLI to `vcpkgTools.xml` in the vcpkg repo.

Otherwise, LGTM. Thanks for the PR!

strega-nil-ms:
thanks y'all!

PR #298, opened 2021-12-08
Clarify "default" feature

ras0219-msft:
This PR extracts out core refactoring exposed by https://github.com/microsoft/vcpkg-tool/pull/295. Specifically, this PR seeks to clarify where in the code we are using "parsed-notation" (no features implies defaults, `[core]` implies not defaults) versus internal notation (`[default]` implies defaults).

The primary change to accomplish this is the introduction of `struct InternalFeatureSet : std::vector<std::string>` to denote when this handoff occurs. By type:
- `Dependency` is parse-notation
- `ParsedQualifiedSpecifier` is parse-notation
- `FullPackageSpec` is internal-notation
- `FeatureSpec` is internal-notation

Therefore, the core conversions to internal notation happen as part of `Dependency::to_full_spec()` and `ParsedQualifiedSpecifier::to_full_spec()` (shared via `normalize_feature_list()`). `filter_dependencies()` uses `Dependency::to_full_spec()`.


include/vcpkg/base/expected.h
```diff
@@ -198,25 +200,25 @@ namespace vcpkg
...
 
-        template<class F, class U = map_t<F>>
-        U then(F f) const&
+        template<class F, class... Args, class U = std::invoke_result_t<F, const_ref_type, Args&&...>>
+        U then(F f, Args&&... args) const&
         {
             if (has_value())
             {
-                return f(*m_t.get());
+                return std::invoke(f, *m_t.get(), static_cast<Args&&>(args)...);
```
strega-nil-ms, line 208: Doesn't this require C++17? Can we actually use C++17?

include/vcpkg/base/expected.h
```diff
@@ -198,25 +200,25 @@ namespace vcpkg
             }
         }
 
-        template<class F, class U = map_t<F>>
-        U then(F f) const&
+        template<class F, class... Args, class U = std::invoke_result_t<F, const_ref_type, Args&&...>>
```
strega-nil-ms, line 203: Why don't we just use:

```c++
template <class F, class... Args>
std::invoke_result_t<F, const_ref_type, Args&&...> then(F f, Args&&... args) const&;
```

include/vcpkg/base/expected.h
```diff
@@ -198,25 +200,25 @@ namespace vcpkg
...
             }
             else
             {
                 return U{error(), expected_right_tag};
             }
         }
 
-        template<class F, class U = move_map_t<F>>
-        U then(F f) &&
+        template<class F, class... Args, class U = std::invoke_result_t<F, move_ref_type, Args&&...>>
```
strega-nil-ms, line 216: same questions as above

include/vcpkg/packagespec.h
```diff
@@ -92,6 +93,12 @@ namespace vcpkg
         std::string m_feature;
     };
 
+    /// In an internal feature set, "default" represents default features and missing "core" has no semantic
+    struct InternalFeatureSet : std::vector<std::string>
```
strega-nil-ms, line 97: why are we inheriting from a standard type here?

include/vcpkg/packagespec.h
```diff
@@ -157,6 +149,9 @@ namespace vcpkg
 
         Json::Object extra_info;
 
+        /// @param implicit_default adds "default" if "core" not present.
+        FullPackageSpec to_full_spec(Triplet target, Triplet host, bool implicit_default) const;
```
strega-nil-ms, line 153: could we make `implicit_default` an enumerator?

include/vcpkg/packagespec.h
```diff
@@ -180,6 +175,12 @@ namespace vcpkg
         Optional<std::vector<std::string>> features;
         Optional<std::string> triplet;
         Optional<PlatformExpression::Expr> platform;
+
+        /// @param implicit_default add "default" if "core" is not present
+        /// @return nullopt on success. On failure, caller should supplement returned string with more context.
+        ExpectedS<FullPackageSpec> to_full_spec(Triplet default_triplet, bool implicit_default) const;
```
strega-nil-ms, line 181: especially since it seems to be used multiple times

strega-nil-ms (commented):
That's what I see for now

include/vcpkg/base/expected.h
ras0219-msft: I assume that with the new "ship binaries" approach, as long as it passes CI it should be sufficient.

include/vcpkg/base/expected.h
ras0219-msft: I think I originally had some `decltype`s in there that needed the arguments. Will fix.

include/vcpkg/packagespec.h
ras0219-msft: I wanted existing algorithms to work and didn't see value in forwarding all the member functions.

This is (almost) purely annotation; in all other ways it should act like a vector.

src/vcpkg/packagespec.cpp
```diff
@@ -91,13 +48,58 @@ namespace vcpkg
         return left.name() == right.name() && left.triplet() == right.triplet();
     }
 
-    ExpectedS<Features> Features::from_string(const std::string& name)
+    static constexpr StringLiteral s_illegal_platform_spec = "platform qualifier is not allowed in this context";
+    static constexpr StringLiteral s_illegal_features = "features are not allowed in this context";
```
strega-nil-ms, line 52: Shouldn't these be localized?

strega-nil-ms (commented):
minor stuff, but mostly looks really good! Thanks @ras0219-msft!

include/vcpkg/packagespec.h
```diff
@@ -101,37 +108,22 @@ namespace vcpkg
...
             : package_spec(std::move(spec)), features(std::move(features))
         {
         }
 
-        std::vector<FeatureSpec> to_feature_specs(const std::vector<std::string>& default_features,
-                                                  const std::vector<std::string>& all_features) const;
-
-        static ExpectedS<FullPackageSpec> from_string(const std::string& spec_as_string, Triplet default_triplet);
+        /// Splats into individual FeatureSpec's
+        void expand_to(std::vector<FeatureSpec>& out) const;
```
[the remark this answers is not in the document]
ras0219-msft: Renamed to `expand_fspecs_to()` -- I find `write` to not be directional enough (is it reading from `out` and writing to the fullpackagespec)?

strega-nil-ms (approved):

PR #305, opened 2021-12-20
cmd_execute_and_stream_data: Determine right exit codes

autoantwort:
According to https://linux.die.net/man/3/pclose:
> The pclose() function waits for the associated process to terminate and returns the exit status of the command as returned by wait4(2). 

And from https://linux.die.net/man/2/wait4:
> Return Value
As for waitpid(2).  

And from https://linux.die.net/man/2/waitpid:
> WIFEXITED(status)  
> ____ returns true if the child terminated normally, that is, by calling exit(3) or _exit(2), or by returning from main(). 
WEXITSTATUS(status)  
       ____ returns the exit status of the child. This consists of the least significant 8 bits of the status argument that the child specified in a call to exit(3) or _exit(2) or as the argument for a return statement in main(). This macro should only be employed if WIFEXITED returned true. 
WIFSIGNALED(status)  
       ____ returns true if the child process was terminated by a signal. 
WTERMSIG(status)  
       ____ returns the number of the signal that caused the child process to terminate. This macro should only be employed if WIFSIGNALED returned true. 
WCOREDUMP(status)  
       ____ returns true if the child produced a core dump. This macro should only be employed if WIFSIGNALED returned true. This macro is not specified in POSIX.1-2001 and is not available on some UNIX implementations (e.g., AIX, SunOS). Only use this enclosed in #ifdef WCOREDUMP ... #endif. 
WIFSTOPPED(status)  
       ____ returns true if the child process was stopped by delivery of a signal; this is only possible if the call was done using WUNTRACED or when the child is being traced (see ptrace(2)). 
WSTOPSIG(status)  
       ____ returns the number of the signal which caused the child to stop. This macro should only be employed if WIFSTOPPED returned true.

### DIFF
--- a/src/vcpkg/base/system.process.cpp
+++ b/src/vcpkg/base/system.process.cpp
@@ -9,6 +9,7 @@
 
 #include <ctime>
 #include <future>
+#include <sstream>
 
 #if defined(__APPLE__)
 #include <mach-o/dyld.h>
@@ -773,7 +774,11 @@ namespace vcpkg
                                     const Environment& env)
     {
         const auto timer = ElapsedTimer::create_started();
-
+        const auto thread_id = []() {
+            std::ostringstream ss;
+            ss << std::this_thread::get_id();
+            return ss.str();
+        }();
 #if defined(_WIN32)
         using vcpkg::g_ctrl_c_state;
 
@@ -803,7 +808,7 @@ namespace vcpkg
                                   .extract();
         }
 
-        Debug::print("popen(", actual_cmd_line, ")\n");
+        Debug::print(thread_id, ": popen(", actual_cmd_line, ")\n");
         // Flush stdout before launching external process
         fflush(stdout);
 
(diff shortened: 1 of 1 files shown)


autoantwort:
And include the thread prefix in the output so that you can see which debug messages belongs together when multiple threads are used to unpack the binary cache

src/vcpkg/base/system.process.cpp
```diff
@@ -773,7 +774,11 @@ namespace vcpkg
                                     const Environment& env)
     {
         const auto timer = ElapsedTimer::create_started();
-
+        const auto thread_id = []() {
+            std::ostringstream ss;
+            ss << std::this_thread::get_id();
+            return ss.str();
+        }();
```
strega-nil-ms, line 781:
```suggestion
        const auto thread_id = std::to_string(std::this_thread::get_id());
```

ras0219-msft:
This seems trivially positive to me, thanks for catching this pretty serious bug!

PR #306, opened 2021-12-20
Retry popen on 127 osx

autoantwort:
Close https://github.com/microsoft/vcpkg/issues/21905 bullet point 1.  

Code 127 normally means `command not found`, but that does not really makes sense here. Sometimes (not often) `unzip` returns 127 and sometimes not, when you retry it the return code is 0. I have no idea what the problem could be, but this implements a workaround by retrying the command once on mac when the return code is 127.

### DIFF
--- a/src/vcpkg/base/system.process.cpp
+++ b/src/vcpkg/base/system.process.cpp
@@ -809,38 +809,52 @@ namespace vcpkg
         }
 
         Debug::print(thread_id, ": popen(", actual_cmd_line, ")\n");
-        // Flush stdout before launching external process
-        fflush(stdout);
+        int exit_code;
+        for (int i = 1; i <= 2; ++i)
+        {
+            // Flush stdout before launching external process
+            fflush(stdout);
 
-        const auto pipe = popen(actual_cmd_line.c_str(), "r");
-        if (pipe == nullptr)
-        {
-            return 1;
-        }
-        char buf[1024];
-        // Use fgets because fread will block until the entire buffer is filled.
-        while (fgets(buf, 1024, pipe))
-        {
-            data_cb(StringView{buf, strlen(buf)});
-        }
+            const auto pipe = popen(actual_cmd_line.c_str(), "r");
+            if (pipe == nullptr)
+            {
+                return 1;
+            }
+            [[maybe_unused]] bool got_output = false;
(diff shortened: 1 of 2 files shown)


src/vcpkg/base/system.process.cpp
```diff
@@ -773,7 +774,11 @@ namespace vcpkg
                                     const Environment& env)
     {
         const auto timer = ElapsedTimer::create_started();
-
+        const auto thread_id = []() {
+            std::ostringstream ss;
+            ss << std::this_thread::get_id();
+            return ss.str();
+        }();
```
strega-nil-ms, line 781: This is not a very good way to do this...

src/vcpkg/base/system.process.cpp
```diff
@@ -803,32 +808,59 @@ namespace vcpkg
...
+                exit_code = WTERMSIG(exit_code);
+            }
+            else if (WIFSTOPPED(exit_code))
+            {
+                exit_code = WSTOPSIG(exit_code);
+            }
+#ifdef __APPLE__
+            if (exit_code == 127 && !got_output)
+            {
+                Debug::print(thread_id, ": pclose returned 127, try again \n");
```
strega-nil-ms, line 853:
```suggestion
                Debug::print(std::to_string(std::this_thread::get_id()), ": pclose returned 127, try again \n");
```

strega-nil-ms (changes requested):
There are some clear (minor) changes needed, but I want someone else to take a look and see if we want to do this.

strega-nil-ms (approved):
nevermind; I am so annoyed that this is the best way to get a string representation of a thread ID.

ras0219-msft:
I merged https://github.com/microsoft/vcpkg-tool/pull/305 since it seems obviously positive. I am less convinced about arbitrarily retrying: child processes are not guaranteed to be idempotent.

If we want to implement a retry around "unzip" specifically on mac, I think that would be reasonable. It should not be implemented for _all_ processes though.

ras0219-msft (changes requested):
See above comment.

ras0219-msft (approved):

ras0219-msft:
Thanks, LGTM!

PR #308, opened 2021-12-21
upgrade: Make --keep-going the default

autoantwort:
Closes [#17922](https://github.com/microsoft/vcpkg/discussions/17922)


vkaytsanov (approved):
Ship it

strega-nil-ms:
I like this choice; however, it definitely requires discussion.

src/vcpkg/commands.upgrade.cpp
```diff
@@ -24,11 +24,13 @@ namespace vcpkg::Commands::Upgrade
...
     static constexpr StringLiteral OPTION_NO_DRY_RUN = "no-dry-run";
     static constexpr StringLiteral OPTION_KEEP_GOING = "keep-going";
+    static constexpr StringLiteral OPTION_STOP_AT_FAILURE = "stop-at-failure";
     static constexpr StringLiteral OPTION_ALLOW_UNSUPPORTED_PORT = "allow-unsupported";
 
-    static constexpr std::array<CommandSwitch, 3> INSTALL_SWITCHES = {{
+    static constexpr std::array<CommandSwitch, 4> INSTALL_SWITCHES = {{
         {OPTION_NO_DRY_RUN, "Actually upgrade"},
-        {OPTION_KEEP_GOING, "Continue installing packages on failure"},
+        {OPTION_KEEP_GOING, "The default (deprecated)"},
```
strega-nil-ms, line 32:
```suggestion
        {OPTION_KEEP_GOING, "Continue installing packages on failure (default)"},
```

src/vcpkg/commands.upgrade.cpp
```diff
@@ -24,11 +24,13 @@ namespace vcpkg::Commands::Upgrade
...
+        {OPTION_STOP_AT_FAILURE, "Stop installing packages on failure"},
```
strega-nil-ms, line 33:
```suggestion
        {OPTION_NO_KEEP_GOING, "Stop installing packages on failure"},
```

src/vcpkg/commands.upgrade.cpp
```diff
@@ -54,8 +56,16 @@ namespace vcpkg::Commands::Upgrade
...
         const ParsedArguments options = args.parse_arguments(COMMAND_STRUCTURE);
 
+        if (Util::Sets::contains(options.switches, OPTION_KEEP_GOING))
+        {
+            print2(Color::warning,
+                   "The option `",
+                   OPTION_KEEP_GOING,
+                   "` is the default now. You don't have to pass this option anymore.");
+        }
+
```
strega-nil-ms, line 66:
```suggestion
```

it's not standard practice to warn on passing the "default" for an option.

src/vcpkg/commands.upgrade.cpp
```diff
@@ -54,8 +56,16 @@ namespace vcpkg::Commands::Upgrade
 
         const ParsedArguments options = args.parse_arguments(COMMAND_STRUCTURE);
```
strega-nil-ms, line 58: Also, make sure that you check for both `OPTION_KEEP_GOING` and `OPTION_NO_KEEP_GOING` and error if both are passed.

src/vcpkg/commands.upgrade.cpp
```diff
@@ -24,11 +24,13 @@ namespace vcpkg::Commands::Upgrade
 
     static constexpr StringLiteral OPTION_NO_DRY_RUN = "no-dry-run";
     static constexpr StringLiteral OPTION_KEEP_GOING = "keep-going";
+    static constexpr StringLiteral OPTION_STOP_AT_FAILURE = "stop-at-failure";
```
strega-nil-ms, line 27:
```suggestion
    static constexpr StringLiteral OPTION_NO_KEEP_GOING = "no-keep-going";
```

this is the standard.

strega-nil-ms (changes requested):
Thanks, just some minor stuff.

BillyONeal (approved):

strega-nil-ms (approved):

PR #309, opened 2021-12-21
platform expressions: add freebsd and openbsd

autoantwort:
"Close"/Implements https://github.com/microsoft/vcpkg/discussions/21964


strega-nil-ms (approved):

strega-nil-ms:
Thanks!

PR #310, opened 2021-12-21
Improve format of platform expressions

autoantwort:
Closes https://github.com/microsoft/vcpkg/issues/22080


strega-nil-ms:
LGTM! Thanks @autoantwort

PR #311, opened 2021-12-21
Auto create issue on install failure

autoantwort:
Closes https://github.com/microsoft/vcpkg/discussions/17490


autoantwort:
For example https://github.com/microsoft/vcpkg/issues/22206 and https://github.com/microsoft/vcpkg/issues/22207 were created by this

strega-nil-ms (commented):
This is really cool! I want @ras0219-msft to take a look first (so it probably won't get merged until january), but I'm definitely a :+1:.

src/vcpkg/base/system.cpp
```diff
@@ -367,6 +368,21 @@ namespace vcpkg
...
+#if defined(_WIN32)
+        return "windows";
+#elif defined(__APPLE__)
+        return "osx";
+#elif defined(__FreeBSD__)
+        return "freebsd";
+#elif defined(__OpenBSD__)
+        return "openbsd";
+#else
+        return "linux";
```
strega-nil-ms, line 383:
```suggestion
#elif defined(__linux__)
        return "linux";
#else
        return "unknown"
```

strega-nil-ms (changes requested):
We shouldn't assume linux, since someone might be running this on z/OS for example.

autoantwort:
Can be merged now :)

BillyONeal (approved):

PR #312, opened 2021-12-23
Fix crash message, the variable has the name `version`

autoantwort:
Currently vcpkg crashes while displaying the crash message :D 
For example:
```
libc++abi: terminating with uncaught exception of type fmt::v8::format_error: argument not found
[1]    41566 abort      vcpkg install --enforce-port-checks --allow-unsupported
```


strega-nil-ms:
Thanks @autoantwort :)

PR #315, opened 2022-01-02
fix link

autoantwort:
Fix [#22289](https://github.com/microsoft/vcpkg/issues/22289)


strega-nil-ms:
thanks @autoantwort !

PR #317, opened 2022-01-04
Fix issue #15376 - Prefab export: invalid Java package name

holmescn:
fix issue #15376 of vcpkg


holmescn:
This PR is closed since export boost to prefab directly is not recommended. If one would like to export boost as prefab, please do it manually.

PR #327, opened 2022-01-20
Fix crash when detecting binary cache

JackBoosY:
https://github.com/microsoft/vcpkg-tool/blob/abe9562a24a2e8fff13af310ef7d6164202350fa/src/vcpkg/binarycaching.cpp#L1451-L1455
https://github.com/microsoft/vcpkg-tool/blob/abe9562a24a2e8fff13af310ef7d6164202350fa/src/vcpkg/binarycaching.cpp#L1416-L1429

`m_status` maybe empty so `cache_status[idx]` maybe a nullptr.

### DIFF
--- a/src/vcpkg/binarycaching.cpp
+++ b/src/vcpkg/binarycaching.cpp
@@ -1457,8 +1457,9 @@ namespace vcpkg
         std::vector<CacheAvailability> results{actions.size()};
         for (size_t idx = 0; idx < results.size(); ++idx)
         {
-            results[idx] = cache_status[idx]->get_available_provider() ? CacheAvailability::available
-                                                                       : CacheAvailability::unavailable;
+            if (cache_status[idx])
+                results[idx] = cache_status[idx]->get_available_provider() ? CacheAvailability::available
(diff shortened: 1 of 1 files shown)


src/vcpkg/binarycaching.cpp
```diff
@@ -1457,8 +1457,9 @@ namespace vcpkg
         std::vector<CacheAvailability> results{actions.size()};
         for (size_t idx = 0; idx < results.size(); ++idx)
         {
-            results[idx] = cache_status[idx]->get_available_provider() ? CacheAvailability::available
-                                                                       : CacheAvailability::unavailable;
+            if (cache_status[idx])
+                results[idx] = cache_status[idx]->get_available_provider() ? CacheAvailability::available
+                                                                           : CacheAvailability::unavailable;
```
JackBoosY, line 1462: Or
```suggestion
                results[idx] = cache_status[idx] ? (cache_status[idx]->get_available_provider() ? CacheAvailability::available
                                                                           : CacheAvailability::unavailable) : CacheAvailability::unavailable;
```

PR #332, opened 2022-01-24
Improve printed filesystem error message

autoantwort:
For example before vcpkg failed with:
```
read_contents(): No such file or directory
```
Now vcpkg prints:
```
/Users/leanderSchulten/git_projekte/vcpkg-tool/src/vcpkg/build.cpp(1349): read_contents(""): No such file or directory
```

### DIFF
--- a/src/vcpkg/base/files.cpp
+++ b/src/vcpkg/base/files.cpp
@@ -72,8 +72,7 @@ namespace
                                                  std::initializer_list<StringView> args)
     {
         auto arguments = args.size() == 0 ? "()" : "(\"" + Strings::join("\", \"", args.begin(), args.end()) + "\")";
-        Checks::exit_with_message(
-            li, Strings::concat(li.file_name, "(", li.line_number, "): ", call_name, arguments, ": ", ec.message()));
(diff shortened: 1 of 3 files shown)


BillyONeal (approved):
LGTM

src/vcpkg/base/files.cpp
```diff
@@ -71,8 +71,9 @@ namespace
                                                  StringView call_name,
                                                  std::initializer_list<StringView> args)
     {
+        auto arguments = args.size() == 0 ? "()" : "(\"" + Strings::join("\", \"", args.begin(), args.end()) + "\")";
         Checks::exit_with_message(
-            li, Strings::concat(call_name, "(", Strings::join(", ", args.begin(), args.end()), "): ", ec.message()));
+            li, Strings::concat(li.file_name, "(", li.line_number, "): ", call_name, arguments, ": ", ec.message()));
```
ras0219-msft, line 76: Since we already have an implementation of printing line information inside `Checks::` (if --debug was passed), it seems like it would be better to extract this out as `Checks::exit_with_message_and_line()` which will always print line information.

ras0219-msft (approved):

ras0219-msft:
Thanks!

PR #335, opened 2022-01-25
[Windows] Make stdout-*.log readable (open in shared for reading mode) and non-buffered

mkhon:
Allow to view stdout-*.log files while build is in progress


BillyONeal (approved):
LGTM thanks!

ras0219-msft (approved):

PR #339, opened 2022-02-02
[VS] allow installations without the english language pack

strega-nil:
See title.

This still requests `VSLANG=1033`, since there may be some minor build systems that don't support non-english language packs; however, we just don't fail when the english language pack isn't installed.

### DIFF
--- a/src/vcpkg/base/system.process.cpp
+++ b/src/vcpkg/base/system.process.cpp
@@ -386,6 +386,9 @@ namespace vcpkg
             new_path += Strings::format(";%s", extra_env.find("PATH")->second);
         env_cstr.append(Strings::to_utf16(new_path));
         env_cstr.push_back(L'\0');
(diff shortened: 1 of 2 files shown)


ras0219-msft (approved):

Neumann-A:
This breaks `vcpkg_copy_pdbs` which requires english output:

https://github.com/microsoft/vcpkg/blob/de176433e9a8769eed0e43d61758f4cdc1dc6e20/scripts/cmake/vcpkg_copy_pdbs.cmake#L21
https://github.com/microsoft/vcpkg/blob/de176433e9a8769eed0e43d61758f4cdc1dc6e20/scripts/cmake/vcpkg_copy_pdbs.cmake#L31

cc @BillyONeal

autoantwort:
It is still broken: https://github.com/microsoft/vcpkg/issues/41400

PR #342, opened 2022-02-05
[#22041] vcpkg install - cmake version check should NOT reject valid response.

Sand2Silicon:
Proposed fix for:  [#22041](https://github.com/microsoft/vcpkg/issues/22041)

As concisely stated by @BRKVings - There are two types of output:
- cmake3 version x.x.x
- cmake version x.x.x
But the current code can't handle the first case.

This minor annoyance, just turned into a hard-blocker for me. I am proposing a change, which works in a minimal-repro program in Visual Studio 2022. I'm not familiar with your code-base. Could someone please test and promote, or propose an alternative that is inline VCPkg style guidelines?

### DIFF
--- a/src/vcpkg/tools.cpp
+++ b/src/vcpkg/tools.cpp
@@ -11,6 +11,7 @@
 #include <vcpkg/archives.h>
 #include <vcpkg/tools.h>
 #include <vcpkg/vcpkgpaths.h>
+#include <regex>
 
 namespace vcpkg
 {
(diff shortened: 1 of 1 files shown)


Sand2Silicon:
@PhoebeHui I took a shot at resolving the bug #22041 I reported. I need some help finishing this. I can not find the buildlog, from azure pipelines and do not see or understand what error are being reported by these checks. Can someone please assist?

PR #343, opened 2022-02-05
(Fix 22041) - vcpkg install - cmake version check should NOT reject valid response. (2nd try)

Sand2Silicon:
Proposed fix for: https://github.com/microsoft/vcpkg/issues/22041

As concisely stated by @BRKVings - There are two types of output:
- cmake3 version x.x.x
- cmake version x.x.x
But the current code can't handle the first case.

This minor annoyance, just turned into a hard-blocker for me. I am proposing a change, which works in a minimal-repro program in Visual Studio 2022. I'm not familiar with your code-base. Could someone please test and promote, or propose an alternative that is inline VCPkg style guidelines?

(2nd try).

### DIFF
--- a/src/vcpkg/tools.cpp
+++ b/src/vcpkg/tools.cpp
@@ -11,6 +11,7 @@
 #include <vcpkg/archives.h>
 #include <vcpkg/tools.h>
 #include <vcpkg/vcpkgpaths.h>
+#include <regex>
 
 namespace vcpkg
 {
@@ -316,8 +317,11 @@ cmake version 3.10.2
 
 CMake suite maintained and supported by Kitware (kitware.com/cmake).
                 */
-            return {Strings::find_exactly_one_enclosed(rc.output, "cmake version ", "\n").to_string(),
-                    expected_left_tag};
+            
+            // There are two expected output formats to handle: "cmake3 version x.x.x" and "cmake version x.x.x"
+            auto simplifiedOutput = std::regex_replace(rc.output, std::regex("cmake3"), "cmake");   
+            return { Strings::find_exactly_one_enclosed(simplifiedOutput, "cmake version ", "\n").to_string(),
+                    expected_left_tag };
         }
     };
 
--- a/src/vcpkg/tools.cpp
+++ b/src/vcpkg/tools.cpp
@@ -319,7 +319,7 @@ CMake suite maintained and supported by Kitware (kitware.com/cmake).
                 */
 
             // There are two expected output formats to handle: "cmake3 version x.x.x" and "cmake version x.x.x"
-            auto simplifiedOutput = std::regex_replace(rc.output, std::regex("cmake3"), "cmake");
+            auto simplifiedOutput = Strings::replace_all(rc.output, "cmake3", "cmake");
             return { Strings::find_exactly_one_enclosed(simplifiedOutput, "cmake version ", "\n").to_string(),
                     expected_left_tag };
         }
--- a/src/vcpkg/tools.cpp
+++ b/src/vcpkg/tools.cpp
@@ -317,9 +317,9 @@ cmake version 3.10.2
 
 CMake suite maintained and supported by Kitware (kitware.com/cmake).
                 */
-            
+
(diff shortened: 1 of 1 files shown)


src/vcpkg/tools.cpp
```diff
@@ -316,8 +317,11 @@ cmake version 3.10.2
 
 CMake suite maintained and supported by Kitware (kitware.com/cmake).
                 */
-            return {Strings::find_exactly_one_enclosed(rc.output, "cmake version ", "\n").to_string(),
-                    expected_left_tag};
+
+            // There are two expected output formats to handle: "cmake3 version x.x.x" and "cmake version x.x.x"
+            auto simplifiedOutput = std::regex_replace(rc.output, std::regex("cmake3"), "cmake");
```
autoantwort, line 322:
```suggestion
            auto simplifiedOutput = Strings::replace_all(rc.output, "cmake3", "cmake");
```

src/vcpkg/tools.cpp
```diff
@@ -11,6 +11,7 @@
 #include <vcpkg/archives.h>
 #include <vcpkg/tools.h>
 #include <vcpkg/vcpkgpaths.h>
+#include <regex>
```
autoantwort, line 14:
```suggestion
```

Then you can also remove the include :)

Sand2Silicon: Yes, I got it! Thank you Autoantwort!

Sand2Silicon:
@autoantwort Thank you for the code-review help. Does anyone know what the failed checks are about? The Azure error messages are unclear to me, something about an unexpected diff?

autoantwort:
Yeah the code is wrongly formatted:
```patch
diff --git a/src/vcpkg/tools.cpp b/src/vcpkg/tools.cpp
index 89d2b1c9..23343269 100644
--- a/src/vcpkg/tools.cpp
+++ b/src/vcpkg/tools.cpp
@@ -319,8 +319,8 @@ CMake suite maintained and supported by Kitware (kitware.com/cmake).
 
             // There are two expected output formats to handle: "cmake3 version x.x.x" and "cmake version x.x.x"
             auto simplifiedOutput = Strings::replace_all(rc.output, "cmake3", "cmake");
-            return { Strings::find_exactly_one_enclosed(simplifiedOutput, "cmake version ", "\n").to_string(),
-                    expected_left_tag };
+            return {Strings::find_exactly_one_enclosed(simplifiedOutput, "cmake version ", "\n").to_string(),
+                    expected_left_tag};
         }
     };
 
```

Sand2Silicon:
> Yeah the code is wrongly formatted:
Thank you! Fixed.

ras0219-msft (approved):

Sand2Silicon:
Is there any final action I have to take to get this merged in? 
Once it is merged, will it immediately be reflected in a fresh `git clone https://github.com/Microsoft/vcpkg.git vcpkg; vcpkg/bootstrap.sh`, or not until a release? I know you're busy but I'm about to share 3 months of work with my team and without this it will fail on our Linux build server.

Sand2Silicon:
> No further action needed, thank you for the PR!
> 
> This unfortunately will not be reflected in a fresh clone until we push out the next build of the tool. We just pushed a new version last Friday, so we are likely to wait until either at least this Friday or next to push another build.

@ras0219-msft Thank you so much I appreciate it! Also thanks for clarifying the time-line to set my expectations. This has been a great first experience with your community!

> 
> Is it possible for you to use a custom-built copy of the tool for your demo?

Team has never used VCPkg before, so I've set the automated build script to `git clone vcpkg; and bootstrap` it before launching CMake (with VCPkg in Manifest mode). Is it possible for me to edit `vcpkg\scripts\bootstrap.sh` to pull a CI or Nightly build of the tool from your build-system instead of the current release? The script could inject that before running bootstrap.sh easily.

ras0219-msft:
We don't sign our CI/PR builds, so we don't make them available to the general public. However, you could try forking the `vcpkg-tool` repo, creating your own release, and uploading your own build of the tool. Then it should be pretty easy to modify bootstrap to pull your own build from there.

Sand2Silicon:
Sounds like a plan! Thank you!

PR #362, opened 2022-02-17
manifest install: Cache compiler information for change detection

autoantwort:
Fixes https://github.com/microsoft/vcpkg/pull/20549#issuecomment-937422751, https://github.com/microsoft/vcpkg/issues/14025 and https://github.com/microsoft/vcpkg/issues/29857.

The noop case now needs 110 ms instead of 2.2 seconds on my arm mac.  
And from 2.0 sec to 18 ms when build as release


src/vcpkg/install.cpp
```diff
@@ -795,6 +799,12 @@ namespace vcpkg::Install
                                  "",
                                  "Error: The option --{option} is not supported in manifest mode.");
 
+    DECLARE_AND_REGISTER_MESSAGE(StampNotChanged,
+                                 (msg::option),
+                                 "",
+                                 "Installation skipped. Everything seems to be installed. To disable this check, don't "
```
strega-nil-ms, line 805: I really do not love this message; it should be a simple "No changes detected" or something, there's no need to make it complex.

src/vcpkg/install.cpp
```diff
@@ -1003,6 +1019,50 @@ namespace vcpkg::Install
             {
                 extended_overlay_ports.push_back(paths.builtin_ports_directory().native());
             }
+
+            std::string hash;
+            {
```
strega-nil-ms, line 1024: I'm not sure we want to calculate the hash if the `check-stamp` option isn't passed

src/vcpkg/install.cpp
```diff
@@ -1003,6 +1019,50 @@ namespace vcpkg::Install
...
+                auto timer = ElapsedTimer::create_started();
+                auto hasher = Hash::get_hasher_for(Hash::Algorithm::Sha256);
+                for (auto&& path : args.overlay_ports)
+                {
+                    for (auto&& file : fs.get_regular_files_recursive(path, VCPKG_LINE_INFO))
+                    {
+                        hasher->add_bytes(fs.read_contents(file, VCPKG_LINE_INFO));
```
strega-nil-ms, line 1031: cc @ras0219 don't we do something interesting for the hashing of port contents?

ras0219-msft: Yes, this hashing method doesn't include file names so renaming a file in an overlay will not result in a rebuild.

ras0219-msft:
A few independent thoughts:

### MSBuild parity

I think we should consider changing our CMake integration to use the same algorithm as MSBuild. Currently, MSBuild considers:

1. Timestamp check on vcpkg.json
2. Timestamp check on vcpkg-configuration.json
3. Triplet name
4. Host triplet name

We _should_ add "full vcpkg command line" to that list in both cases. My concern would be only that this will not pick up changes to overlay port contents or to the triplet files.

### Locking

Right now, vcpkg _generally_ performs a bunch of file locking in manifest mode that will cause this "lightweight" check to be globally serialized. For CMake usage this is less important since it's often called only once for the whole repo, but for more complex scenarios it could be problematic.

### Avoid hashing file contents

If the intent of this is to be a lightweight check, we should avoid hashing file contents altogether. Instead, I would suggest:
1. For every file under consideration, compare their modified timestamp to the stamp's modified timestamp.
2. If the timestamps disagree (a file is newer than the stamp), report a mismatch.
3. For every file under consideration, hash the path. This ensures deleting files will be tracked correctly. (In practice I think this is easiest by hashing the list of paths)
4. Read the stamp contents and compare against the paths hash. If not matching, report a mismatch.

If there's a mismatch, perform the install. Once the install completes, write the new paths hash to the stamp (which updates the timestamp).

include/vcpkg/installedpaths.h
```diff
@@ -20,6 +20,7 @@ namespace vcpkg
         Path vcpkg_dir_info() const { return vcpkg_dir() / "info"; }
         Path vcpkg_dir_updates() const { return vcpkg_dir() / "updates"; }
         Path lockfile_path() const { return vcpkg_dir() / "vcpkg-lock.json"; }
+        Path stampfile_path() const { return vcpkg_dir() / "stampfile"; }
```
BillyONeal, line 23: Hmmm... this is almost certainly not the correct path; it probably should be part of the installed tree rather than the vcpkg_root.

In particular, the vcpkg_root may be readonly.

(The lockfile is OK because it's a read / lock operation applied to the file rather than a write)

autoantwort: `vcpkg_dir()` is the `vcpkg` dir inside of the install folder.

autoantwort: See https://github.com/microsoft/vcpkg-tool/blob/b3b6577286e8c7791aa572c204ca2699ddc0de21/src/vcpkg/vcpkgpaths.cpp#L415-L435

BillyONeal: Megaderp, thanks

zig-for:
It would be really neat if this were merged, this is really slowing down my builds.

autoantwort:
@BillyONeal Any news? :)

BillyONeal:
I think for some reason I had in my mind that someone else specific was to review this one. Sorry, I don't have updated news for you.

autoantwort:
Thank you for the update :) Maybe you can poke the person. 
This change would mean a big improvement in usability for cmake users (100 times faster configure runs).

BillyONeal:
> Thank you for the update :) Maybe you can poke the person.

To clarify, I had in my mind that someone was supposed to look at this but that doesn't appear to be the case

autoantwort:
Ah ok

src/vcpkg/install.cpp
```diff
@@ -1070,6 +1079,79 @@ namespace vcpkg
...
+                {
+                    auto maybe_git_sha = paths.get_current_git_sha();
+                    if (maybe_git_sha)
+                    {
+                        hasher->add_bytes(maybe_git_sha.value_or_exit(VCPKG_LINE_INFO));
+                    }
+                }
+                hash = hasher->get_hash();
+
+                auto have_file_change = [&]() {
```
BillyONeal, line 1101: I'm really not a fan if declaring a lambda like this, can it be extracted?

src/vcpkg/install.cpp
```diff
@@ -1070,6 +1079,79 @@ namespace vcpkg
                 extended_overlay_ports.emplace_back(paths.builtin_ports_directory().native());
             }
 
+            std::string hash;
+            {
+                auto hasher = Hash::get_hasher_for(Hash::Algorithm::Sha256);
+                hasher->add_bytes(default_triplet.to_string());
+                hasher->add_bytes(host_triplet.to_string());
```
BillyONeal, line 1086: Can `add_bytes` be taught about `StringView`?

src/vcpkg/install.cpp
```diff
@@ -1070,6 +1079,79 @@ namespace vcpkg
...
+                for (const auto& feature : features)
+                {
+                    hasher->add_bytes(feature);
+                }
+                if (paths.get_registry_set().is_default_builtin_registry())
+                {
+                    auto maybe_git_sha = paths.get_current_git_sha();
+                    if (maybe_git_sha)
```
BillyONeal, line 1094: I'm not sure just giving up and not recording the hash if we aren't in a git repo is the right behavior. We need the git part to detect that none of the depended-upon stuff changed.

src/vcpkg/install.cpp
```diff
@@ -1070,6 +1079,79 @@ namespace vcpkg
...
+                    {
+                        hasher->add_bytes(maybe_git_sha.value_or_exit(VCPKG_LINE_INFO));
+                    }
+                }
+                hash = hasher->get_hash();
+
+                auto have_file_change = [&]() {
+                    auto last_install = fs.last_write_time(paths.installed().stampfile_path(), VCPKG_LINE_INFO);
```
BillyONeal, line 1102: `paths.installed()` and `paths.installed().stampfile_path()` are repeated a lot; introduce locals?

BillyONeal (commented):
I think a change like this needs a spec for how the stamp is calculated where we justify that that is in fact all the inputs that could possibly affect the output.

Do we have an understanding of where the cost delta comes from? I would substantially prefer a way for users to explicitly opt-in to reusing compiler detection across invocations which would feed into our existing abi_hash diffing mechanism, for instance.

autoantwort:
> Do we have an understanding of where the cost delta comes from? 

Mostly hashing the compiler. See https://github.com/microsoft/vcpkg/discussions/17429

autoantwort:
> I would substantially prefer a way for users to explicitly opt-in to reusing compiler detection across invocations which would feed into our existing abi_hash diffing mechanism, for instance.

That would help a lot. I guess *reusing compiler detection across invocations* is the same as `set(VCPKG_DISABLE_COMPILER_TRACKING ON)`. The runtime of `set(VCPKG_DISABLE_COMPILER_TRACKING ON)` and `--check-stamp` is nearly the same.

autoantwort:
@BillyONeal I have now implemented this (Currently not behind a command line switch).

autoantwort:
@BillyONeal The following change to the main repo is needed. But I guess we have to move the detect_compiler files to this repo so that the new vcpkg-tool works with old vcpkg commits. 

```diff
diff --git a/scripts/detect_compiler/CMakeLists.txt b/scripts/detect_compiler/CMakeLists.txt
index c55da9ebb..db7973e7b 100644
--- a/scripts/detect_compiler/CMakeLists.txt
+++ b/scripts/detect_compiler/CMakeLists.txt
@@ -19,6 +19,8 @@ message("#COMPILER_HASH#${COMPILER_HASH}")
 message("#COMPILER_C_HASH#${C_HASH}")
 message("#COMPILER_C_VERSION#${CMAKE_C_COMPILER_VERSION}")
 message("#COMPILER_C_ID#${CMAKE_C_COMPILER_ID}")
+message("#COMPILER_C_PATH#${CMAKE_C_COMPILER}")
 message("#COMPILER_CXX_HASH#${CXX_HASH}")
 message("#COMPILER_CXX_VERSION#${CMAKE_CXX_COMPILER_VERSION}")
 message("#COMPILER_CXX_ID#${CMAKE_CXX_COMPILER_ID}")
+message("#COMPILER_CXX_PATH#${CMAKE_CXX_COMPILER}")
```

autoantwort:
Should it be a feature flag like compiler_tracking or a "normal" command line switch?

cenit:
any news on this PR?

autoantwort:
According to https://discord.com/channels/400588936151433218/687365466422902841/1038192801491464192 it will probably not be merged and they will not look at it before march

autoantwort:
No. There are concerns about cache invalidation.  For example if you do the following with this PR:
1. `vcpkg install --cache-compiler-information`
2. Update your compiler 
3. `vcpkg install --cache-compiler-information` # This does check that the compiler was updated and reinstalls
4. `export CXX=<path_to_different_compiler>`
5. `vcpkg install --cache-compiler-information` # This would not do anything, since the cached compiler was not updated
6. Change `vcpkg.json`
7. Run `vcpkg install`. This would use the compiler abi from 3., but the compiler from 4. ⚡   

So a solution could be to only use `--cache-compiler-information` to detect changes. If there are changes, rerun compiler detection and install packages. This would fix point 7. 

Another solution is to cache the compiler and use it later (For example by sharing the CMakeCache.txt). This would use the compiler from point 3 in point 7. 

Another more simple solution would be to only check if the vcpkg.json file was changed. The point 3 would do nothing and point 7 would work normally.

src/vcpkg/commands.build.cpp
```diff
@@ -666,14 +783,60 @@ namespace vcpkg
                                            const PreBuildInfo& pre_build_info,
                                            const Toolset& toolset)
     {
+        static constexpr auto vcpkg_json = R"--(
```
BillyONeal, line 786: Now that we have https://github.com/microsoft/vcpkg/pull/36398 do you mind if I replace this attempt to move detect_compiler out with just using what @data-queue added?

include/vcpkg/installedpaths.h
```diff
@@ -22,6 +22,10 @@ namespace vcpkg
         Path vcpkg_dir_updates() const { return vcpkg_dir() / "updates"; }
         Path lockfile_path() const { return vcpkg_dir() / "vcpkg-lock.json"; }
         Path triplet_dir(Triplet t) const { return m_root / t.canonical_name(); }
+        Path compiler_info_cache_file(Triplet t) const
+        {
+            return vcpkg_dir() / "compiler_info_cache" / t.canonical_name();
```
BillyONeal, line 27: This needs to be in buildtrees as the vcpkg_dir() might be not writable.

BillyONeal: Oh this is in `InstalledPaths` damnit

BillyONeal:
@BillyONeal

I've convinced myself we should take this because changing the compiler under an installed tree is already doomed.

@ras0219-msft (paraphrased)

The thing I am concerned about is corrupting binary caches. If the installed tree is toast you can recover by deleting it and reinstalling, but if binary caches are toast recovering is difficult.

If the meat of this is saying the hash algorithm is too slow, we don't need to skip the cmake detection to cache the hash. We can have a database file saying "this path was this filesize / last mod time / sha".

We could also hash the first MB of the compiler or something like that.

@BillyONeal

OK, that makes sense to me. What if we kept the compiler from the metadata cache and used that to lock the version?

@ras0219-msft (paraphrased)

That would resolve my concerns.

src/vcpkg/commands.build.cpp
autoantwort: You can do it :)

PatrickKa:
Any update/news on this?

BillyONeal:
> Any update/news on this?

Given that the discussion in https://github.com/microsoft/vcpkg-tool/pull/362#issuecomment-2060041817 isn't resolved I don't see progress happening here.

I did some more testing and it looks like the majority of the time is not SHA1-ing the compiler binaries, but is instead CMake detecting which compiler to use. Note that I intentionally used clang in this example, which is a 116MB binary, rather than cl.exe which is like 3MB:

```console
PS D:\vcpkg> Measure-Command { cmake.exe -DCURRENT_PORT_DIR=D:\vcpkg\scripts\detect_compiler -DCURRENT_BUILDTREES_DIR=D:\vcpkg\buildtrees -DCURRENT_PACKAGES_DIR=D:\vcpkg\packages\detect_compiler_x64-windows -D_HOST_TRIPLET=x64-windows -DCMD=BUILD -DDOWNLOADS=D:\vcpkg-downloads -DTARGET_TRIPLET=x64-windows -DTARGET_TRIPLET_FILE=D:\vcpkg\triplets\x64-windows.cmake -DVCPKG_ROOT_DIR=D:\vcpkg -DPACKAGES_DIR=D:\vcpkg\packages -DBUILDTREES_DIR=D:\vcpkg\buildtrees -D_VCPKG_INSTALLED_DIR=D:\vcpkg\installed -DVCPKG_MANIFEST_INSTALL=OFF -DVCPKG_BASE_VERSION=2999-12-12 -DDO_HASH=OFF -P .\scripts\ports.cmake }
[1/1] "C:/Program Files/CMake/bin/cmake.exe" -E chdir ".." "C:/Program Files/CMake/bin/cmake.exe" "D:/vcpkg/scripts/detect_compiler" "-G" "Ninja" "-DCMAKE_BUILD_TYPE=Release" "-DCMAKE_INSTALL_PREFIX=D:\vcpkg\packages\detect_compiler_x64-windows" "-DDO_HASH=OFF" "-DCMAKE_MAKE_PROGRAM=D:/vcpkg-downloads/tools/ninja/1.10.2-windows/ninja.exe" "-DBUILD_SHARED_LIBS=ON" "-DVCPKG_CHAINLOAD_TOOLCHAIN_FILE=D:/vcpkg/scripts/toolchains/windows.cmake" "-DVCPKG_TARGET_TRIPLET=x64-windows" "-DVCPKG_SET_CHARSET_FLAG=ON" "-DVCPKG_PLATFORM_TOOLSET=" "-DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON" "-DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=ON" "-DCMAKE_FIND_PACKAGE_NO_SYSTEM_PACKAGE_REGISTRY=ON" "-DCMAKE_INSTALL_SYSTEM_RUNTIME_LIBS_SKIP=TRUE" "-DCMAKE_VERBOSE_MAKEFILE=ON" "-DVCPKG_APPLOCAL_DEPS=OFF" "-DCMAKE_TOOLCHAIN_FILE=D:/vcpkg/scripts/buildsystems/vcpkg.cmake" "-DCMAKE_ERROR_ON_ABSOLUTE_INSTALL_DESTINATION=ON" "-DVCPKG_CXX_FLAGS=" "-DVCPKG_CXX_FLAGS_RELEASE=" "-DVCPKG_CXX_FLAGS_DEBUG=" "-DVCPKG_C_FLAGS=" "-DVCPKG_C_FLAGS_RELEASE=" "-DVCPKG_C_FLAGS_DEBUG=" "-DVCPKG_CRT_LINKAGE=dynamic" "-DVCPKG_LINKER_FLAGS=" "-DVCPKG_LINKER_FLAGS_RELEASE=" "-DVCPKG_LINKER_FLAGS_DEBUG=" "-DVCPKG_TARGET_ARCHITECTURE=x64" "-DCMAKE_INSTALL_LIBDIR:STRING=lib" "-DCMAKE_INSTALL_BINDIR:STRING=bin" "-D_VCPKG_ROOT_DIR=D:\vcpkg" "-DZ_VCPKG_ROOT_DIR=D:\vcpkg" "-D_VCPKG_INSTALLED_DIR=D:\vcpkg\installed" "-DVCPKG_MANIFEST_INSTALL=OFF"
-- The C compiler identification is Clang 18.1.6 with GNU-like command-line
-- The CXX compiler identification is Clang 18.1.6 with GNU-like command-line
#COMPILER_HASH#
#COMPILER_C_HASH#
#COMPILER_C_VERSION#18.1.6
#COMPILER_C_ID#Clang
#COMPILER_C_PATH#C:/Program Files/LLVM/bin/clang.exe
#COMPILER_CXX_HASH#
#COMPILER_CXX_VERSION#18.1.6
#COMPILER_CXX_ID#Clang
#COMPILER_CXX_PATH#C:/Program Files/LLVM/bin/clang++.exe
-- Configuring done (1.2s)
-- Generating done (0.0s)
CMake Warning:
  Manually-specified variables were not used by the project:

    BUILD_SHARED_LIBS
    CMAKE_INSTALL_BINDIR
    CMAKE_INSTALL_LIBDIR
    _VCPKG_ROOT_DIR


-- Build files have been written to: D:/vcpkg/buildtrees/x64-windows-rel


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 1
Milliseconds      : 332
Ticks             : 13323311
TotalDays         : 1.54204988425926E-05
TotalHours        : 0.000370091972222222
TotalMinutes      : 0.0222055183333333
TotalSeconds      : 1.3323311
TotalMilliseconds : 1332.3311

PS D:\vcpkg> Measure-Command { cmake.exe -DCURRENT_PORT_DIR=D:\vcpkg\scripts\detect_compiler -DCURRENT_BUILDTREES_DIR=D:\vcpkg\buildtrees -DCURRENT_PACKAGES_DIR=D:\vcpkg\packages\detect_compiler_x64-windows -D_HOST_TRIPLET=x64-windows -DCMD=BUILD -DDOWNLOADS=D:\vcpkg-downloads -DTARGET_TRIPLET=x64-windows -DTARGET_TRIPLET_FILE=D:\vcpkg\triplets\x64-windows.cmake -DVCPKG_ROOT_DIR=D:\vcpkg -DPACKAGES_DIR=D:\vcpkg\packages -DBUILDTREES_DIR=D:\vcpkg\buildtrees -D_VCPKG_INSTALLED_DIR=D:\vcpkg\installed -DVCPKG_MANIFEST_INSTALL=OFF -DVCPKG_BASE_VERSION=2999-12-12 -DDO_HASH=ON -P .\scripts\ports.cmake }
[1/1] "C:/Program Files/CMake/bin/cmake.exe" -E chdir ".." "C:/Program Files/CMake/bin/cmake.exe" "D:/vcpkg/scripts/detect_compiler" "-G" "Ninja" "-DCMAKE_BUILD_TYPE=Release" "-DCMAKE_INSTALL_PREFIX=D:\vcpkg\packages\detect_compiler_x64-windows" "-DDO_HASH=ON" "-DCMAKE_MAKE_PROGRAM=D:/vcpkg-downloads/tools/ninja/1.10.2-windows/ninja.exe" "-DBUILD_SHARED_LIBS=ON" "-DVCPKG_CHAINLOAD_TOOLCHAIN_FILE=D:/vcpkg/scripts/toolchains/windows.cmake" "-DVCPKG_TARGET_TRIPLET=x64-windows" "-DVCPKG_SET_CHARSET_FLAG=ON" "-DVCPKG_PLATFORM_TOOLSET=" "-DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON" "-DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=ON" "-DCMAKE_FIND_PACKAGE_NO_SYSTEM_PACKAGE_REGISTRY=ON" "-DCMAKE_INSTALL_SYSTEM_RUNTIME_LIBS_SKIP=TRUE" "-DCMAKE_VERBOSE_MAKEFILE=ON" "-DVCPKG_APPLOCAL_DEPS=OFF" "-DCMAKE_TOOLCHAIN_FILE=D:/vcpkg/scripts/buildsystems/vcpkg.cmake" "-DCMAKE_ERROR_ON_ABSOLUTE_INSTALL_DESTINATION=ON" "-DVCPKG_CXX_FLAGS=" "-DVCPKG_CXX_FLAGS_RELEASE=" "-DVCPKG_CXX_FLAGS_DEBUG=" "-DVCPKG_C_FLAGS=" "-DVCPKG_C_FLAGS_RELEASE=" "-DVCPKG_C_FLAGS_DEBUG=" "-DVCPKG_CRT_LINKAGE=dynamic" "-DVCPKG_LINKER_FLAGS=" "-DVCPKG_LINKER_FLAGS_RELEASE=" "-DVCPKG_LINKER_FLAGS_DEBUG=" "-DVCPKG_TARGET_ARCHITECTURE=x64" "-DCMAKE_INSTALL_LIBDIR:STRING=lib" "-DCMAKE_INSTALL_BINDIR:STRING=bin" "-D_VCPKG_ROOT_DIR=D:\vcpkg" "-DZ_VCPKG_ROOT_DIR=D:\vcpkg" "-D_VCPKG_INSTALLED_DIR=D:\vcpkg\installed" "-DVCPKG_MANIFEST_INSTALL=OFF"
-- The C compiler identification is Clang 18.1.6 with GNU-like command-line
-- The CXX compiler identification is Clang 18.1.6 with GNU-like command-line
#COMPILER_HASH#0f4a11f3a8b3c6eee0390888b2ca585daca373eb
#COMPILER_C_HASH#cf5d8e7225d21ffe029bdf032edaaf8a66ea0594
#COMPILER_C_VERSION#18.1.6
#COMPILER_C_ID#Clang
#COMPILER_C_PATH#C:/Program Files/LLVM/bin/clang.exe
#COMPILER_CXX_HASH#cf5d8e7225d21ffe029bdf032edaaf8a66ea0594
#COMPILER_CXX_VERSION#18.1.6
#COMPILER_CXX_ID#Clang
#COMPILER_CXX_PATH#C:/Program Files/LLVM/bin/clang++.exe
-- Configuring done (1.6s)
-- Generating done (0.0s)
CMake Warning:
  Manually-specified variables were not used by the project:

    BUILD_SHARED_LIBS
    CMAKE_INSTALL_BINDIR
    CMAKE_INSTALL_LIBDIR
    _VCPKG_ROOT_DIR


-- Build files have been written to: D:/vcpkg/buildtrees/x64-windows-rel


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 1
Milliseconds      : 753
Ticks             : 17538336
TotalDays         : 2.0299E-05
TotalHours        : 0.000487176
TotalMinutes      : 0.02923056
TotalSeconds      : 1.7538336
TotalMilliseconds : 1753.8336

PS D:\vcpkg>
```

So the hashing is approximately `(1.33-1.75)/1.33 ~= 31%` of the time. Not nothing, but not the earthshattering difference in the original PR description.

Given that @ras0219-msft does not want to accept any change that skips the CMake compiler detection over the potential to damage binary caches, I am unsure what changes could be made to get this PR into a state we could merge.

BillyONeal:
To clarify, the problematic scenario is:

1. Run vcpkg with this setting turned on.
2. Change something about the environment that causes a different compiler to be selected.
3. Run vcpkg with this setting turned on again. It sees that the compiler that was detected before has not changed, so it skips running the compiler detection step and uses the old compiler's SHAs.
4. Binary caching uploads bits built with a different compiler as if they were built with the old compiler.
5. Future binary cache hits with the old compiler are now poisoned by the new compiler.

so we would have to have high confidence that anything that could possibly change which compiler CMake could pick invalidated this detection, and it seems unlikely to me that we will ever reach that level of confidence given how lots of how CMake does this is a black box.

autoantwort:
So the hashing is approximately (1.33-1.75)/1.33 ~= -31% of the time. Not nothing, but not the earthshattering difference in the original PR description. 
I have tested it on my mac: 
Baseline noop: 1.4s
Baseline noop without hashing: 0.5s 
This PR noop: ~0.1s 

> Given that @ras0219-msft does not want to accept any change that skips the CMake compiler detection over the potential to damage binary caches, I am unsure what changes could be made to get this PR into a state we could merge.

I already sketched a solution but got no real response. If this is the only think blocking this PR I will start implementing that :)

BillyONeal:
> I already sketched a solution but got no real response.

Sorry I missed that, can you point specifically?

autoantwort:
https://github.com/microsoft/vcpkg-tool/pull/362#issuecomment-1752192106 
Especially:
> So a solution could be to only use --cache-compiler-information to detect changes. If there are changes, rerun compiler detection and install packages. This would fix point 7.

BillyONeal:
> [#362 (comment)](https://github.com/microsoft/vcpkg-tool/pull/362#issuecomment-1752192106) Especially:
> 
> > So a solution could be to only use --cache-compiler-information to detect changes. If there are changes, rerun compiler detection and install packages. This would fix point 7.

I'm not sure how such changes are going to be possible in a way that truly sells that it's safe.

BillyONeal:
> If the meat of this is saying the hash algorithm is too slow, we don't need to skip the cmake detection to cache the hash. We can have a database file saying "this path was this filesize / last mod time / sha".

Would you be willing to do just this part? It would fix your Mac test and is clearly less controversial..

src/vcpkg/commands.set-installed.cpp
```diff
@@ -217,7 +243,32 @@ namespace vcpkg
...
+                }
+                compute_all_abis(paths, action_plan, cmake_vars, {});
+                used_cached_compiler_info = false;
+            }
+        }
         adjust_action_plan_to_status_db(action_plan, status_db);
+        if (used_cached_compiler_info)
+        {
+            Checks::check_exit(VCPKG_LINE_INFO, action_plan.empty());
+        }
```
autoantwort, line 271: Now the cached compiler info is only used to detect if there is something to install, if something gets installed, fresh compiler info are used. 
(if you update your compiler the cache is not used)

miss-programgamer:
At the risk of annoying everyone who's posted here before: any update on this? It's been more than three years since this pull request was created...

BillyONeal:
> At the risk of annoying everyone who's posted here before: any update on this? It's been more than three years since this pull request was created...

You should see the majority of the benefit after https://github.com/microsoft/vcpkg/pull/42994 and https://github.com/microsoft/vcpkg-tool/pull/1558

autoantwort:
> You should see the majority of the benefit after

I wouldn't say that. This PR vs latest release on windows (8x faster) and mac (14x faster): 
image
image

ras0219-msft:
In absolute terms: saves 2s on Windows and ~0.5s on MacOS

autoantwort:
Is this blocked because of a technical issue or by your sparse review time?

autoantwort:
Would you consider merging it when it is hidden behind a feature flag that is disabled by default?

PR #367, opened 2022-02-19
cmake: add target generate-message-map

autoantwort:
Maybe mark this target as `ALL` so that it is build by default


strega-nil-ms:
I want this to wait for #364 since that will add another line to this. Also, adding tag requires:discussion for the thursday meeting (although I'm in favor)

CMakeLists.txt
```diff
@@ -229,6 +229,17 @@ if(VCPKG_PDB_SUFFIX)
     set_property(TARGET vcpkg PROPERTY PDB_NAME "vcpkg${VCPKG_PDB_SUFFIX}")
 endif()
 
+# === Target: generate-message-map ===
+
+add_custom_target(generate-message-map
+   COMMAND vcpkg x-generate-default-message-map --no-allow-incorrect-comments locales/messages.json
+   COMMAND vcpkg x-generate-default-message-map --no-output-comments locales/messages.en.json
+   WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
+   DEPENDS vcpkg
```
strega-nil-ms, line 238: this is unnecessary, since this already depends on the `vcpkg` target:

```suggestion
```

strega-nil-ms (approved):

PR #400, opened 2022-03-01
use tar if available.

Neumann-A:


src/vcpkg/archives.cpp
```diff
@@ -140,7 +140,26 @@ namespace
                  Strings::case_insensitive_ascii_equals(ext, ".7z") ||
                  Strings::case_insensitive_ascii_equals(ext, ".exe"))
         {
-            win32_extract_with_seven_zip(paths, archive, to_path);
+            auto tars = paths.get_filesystem().find_from_PATH("tar");
+#if defined(_WIN32)
+            Util::erase_remove_if(tars, [](auto& elem) {
+                if (vcpkg::Strings::contains(elem,"\\system32\\tar.exe"))  {
+                    return false; 
+                }
```
Neumann-A, line 148: this is just to be explicit. can be removed if requested.

src/vcpkg/archives.cpp
```diff
@@ -140,7 +140,26 @@ namespace
...
+                else if (vcpkg::Strings::contains(elem,"\\usr\\bin\\tar.exe")) {
```
Neumann-A, line 149: uncertain how this will effect MinGW?

dg0yt:
> Yep, and that's exactly what we use once we have cmake -- the problem is that getting cmake requires unzipping.

FTR, there also self-extracting ".sh" downloads for CMake. However I don't know their hidden dependencies.
https://cmake.org/download/

PR #411, opened 2022-03-03
re-enable disabled warnings which are no longer relevant

strega-nil:
try to find all the warnings that are still an issue in modern VS

Also, standardize on `VCPKG_MSVC_WARNING`


BillyONeal (approved):
I would like to see the commit title when merged to be something like:

`pragma warning => VCPKG_MSVC_WARNING`

to indicate that we aren't just blindly removing stuff.

PR #439, opened 2022-03-16
[regex-ripout] Part 3/4

strega-nil:
The final regex-ripout shall be part of @BillyONeal's tools PR; this rips regex out of `commands.integrate.cpp` and `install.cpp`.


src/vcpkg-test/integrate.cpp
```diff
@@ -0,0 +1,66 @@
...
+    REQUIRE(res.has_value());
+    CHECK(*res.get() == 1);
+
+    res = Integrate::find_targets_file_version("<!-- ver 1 -->");
+    CHECK_FALSE(res.has_value());
+}
+
+TEST_CASE ("get_bash_source_completion_lines", "[integrate]")
+{
+
```
strega-nil, line 60: [need to implement these tests]

src/vcpkg/commands.integrate.cpp
```diff
@@ -14,6 +14,101 @@
...
+        std::vector<std::string> matches;
+        auto first = contents.begin();
+        const auto last = contents.end();
+        while (first != last)
+        {
+            const auto end_of_line = std::find(first, last, '\n');
+            const auto line = Strings::trim(StringView{first, end_of_line});
+            if (Strings::starts_with(line, "source") && Strings::ends_with(line, "scripts/vcpkg_completion.bash"))
+            {
+                matches.emplace_back(line.begin(), line.end());
```
BillyONeal, line 57:
```suggestion
                matches.emplace_back(line.data(), line.size());
```

The char*,size_t constructor for string is slightly more efficient because it doesn't need metaprogramming to determine that the input is a contiguous range.

src/vcpkg/install.cpp
```diff
@@ -621,11 +621,53 @@ namespace vcpkg::Install
...
 
-    CMakeUsageInfo get_cmake_usage(const Filesystem& fs, const InstalledPaths& installed, const BinaryParagraph& bpgh)
+    static constexpr StringLiteral ADD_LIBRARY_CALL = "add_library(";
+
+    static const char* find_add_library(const char* real_first, const char* first, const char* last)
+    {
+        for (;;)
+        {
+            first = std::search(first, last, ADD_LIBRARY_CALL.begin(), ADD_LIBRARY_CALL.end());
+            if (first == real_first || first == last || !Parse::ParserBase::is_word_char(*(first - 1)))
```
BillyONeal, line 631:
```suggestion
            if (first == real_first || first == last || !Parse::ParserBase::is_word_char(first[-1]))
```

src/vcpkg/install.cpp
```diff
@@ -621,11 +621,53 @@ namespace vcpkg::Install
...
+            {
+                return first;
```
BillyONeal, line 633: The caller skips over ADD_LIBRARY_CALL.size(), should that be here so that only this function need see the variable ADD_LIBRARY_CALL?

src/vcpkg/install.cpp
```diff
@@ -662,16 +704,11 @@ namespace vcpkg::Install
...
+                        auto targets = get_cmake_add_library_names(contents);
+                        if (!targets.empty())
                         {
-                            auto match = *next;
-                            auto& targets = library_targets[find_package_name];
-                            if (std::find(targets.cbegin(), targets.cend(), match[1]) == targets.cend())
-                                targets.push_back(match[1]);
-                            ++next;
+                            auto& all_targets = library_targets[find_package_name];
+                            all_targets.insert(all_targets.end(), targets.begin(), targets.end());
```
BillyONeal, line 711:
```suggestion
                            all_targets.insert(all_targets.end(), std::make_move_iterator(targets.begin()), std::make_move_iterator(targets.end()));
```

Needs `#include <iterator>` if not already done in this file.

src/vcpkg/install.cpp
```diff
@@ -740,30 +777,27 @@ namespace vcpkg::Install
...
+                            msg, "    target_link_libraries(main PRIVATE ", Strings::join(" ", targets), ")\n\n");
                     }
                     else
                     {
-                        auto omitted = library_target_pair.second.size() - 4;
-                        library_target_pair.second.erase(library_target_pair.second.begin() + 4,
-                                                         library_target_pair.second.end());
+                        auto omitted = targets.size() - 4;
+                        library_target_pair.second.erase(targets.begin() + 4, targets.end());
                         msg += Strings::format("    # Note: %zd target(s) were omitted.\n"
```
BillyONeal, line 797: fmtize?

BillyONeal (approved):
If possible I would like someone with a better understanding of shell things to look at the shell things.

src/vcpkg/install.cpp
BillyONeal: (Also, I hate every map::operator[] I've ever seen and this is no exception but that's not changed here so :P )

src/vcpkg/commands.integrate.cpp
```diff
@@ -14,6 +14,101 @@
...
+        auto first = contents.begin();
+        const auto last = contents.end();
+        for (;;)
+        {
+            first = std::search(first, last, VERSION_START.begin(), VERSION_START.end());
+            if (first == last)
+            {
+                break;
+            }
+            first += VERSION_START.size();
```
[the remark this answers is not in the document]
strega-nil: probably

src/vcpkg/commands.integrate.cpp
```diff
@@ -14,6 +14,101 @@
...
+            auto version_end = std::search(first, last, VERSION_END.begin(), VERSION_END.end());
```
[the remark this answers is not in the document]
strega-nil: This does work, see the tests.

src/vcpkg/install.cpp
strega-nil: I know this is equivalent, but I _really_ don't like array indexing a pointer; I'd prefer to treat it as an un-indexable iterator

src/vcpkg/install.cpp
strega-nil: you will get no disagreement from me

src/vcpkg/install.cpp
strega-nil: This should likely be non-localized, so I'll make it `{Locked}`.

src/vcpkg/install.cpp
BillyONeal: > I'd prefer to treat it as an un-indexable iterator

Note that random-access iterators require `op[]`: http://eel.is/c++draft/random.access.iterators

image

No additional change requested; just wanted to note this.

PR #456, opened 2022-03-26
General object storage interface for binary caching.

day253:
I would like to see the aws and gcs merged such that the tool exe is passed in if the intent is that they truly have the same interface so that we don't forget to update one when changes are made to the other.

This PR is to enable vcpkg users using general object storage interface for binary caching.


day253:
#435

day253:
Could you help me about the pipeline? I changed nearly nothing but the pipeline is failed. #457 @BillyONeal

day253:
> Could you help me about the pipeline? I changed nearly nothing but the pipeline is failed. #457 @BillyONeal

I this pull request #460 will resolve my problem.

vicroms:
Hi @day253 

For PRs that include localizable strings you don't need to provide translations for each language, we have a team of translators at Microsoft that takes care of that. If you want to submit translation recommendations please open an issue in the main vcpkg repository.

The process to add localizable strings is to declare them with `DECLARE_AND_REGISTER_MESSAGE` in your code. Then you can just run the `generate-message-map` custom CMake target.

PowerShell Example:
```
# from the vcpkg-tool directory
mkdir build; cd build
cmake .. -G Ninja
cmake --build . --target generate-message-map
```

This will generate updated `locales/messages.json` and `locales/messages.en.json` that you can commit in your PR.

day253:
> Hi @day253 
> 
> 
> 
> For PRs that include localizable strings you don't need to provide translations for each language, we have a team of translators at Microsoft that takes care of that. If you want to submit translation recommendations, please open an issue in the main vcpkg repository.
> 
> 
> 
> The process to add localizable strings is to:
> 
> 1)  declare them with `DECLARE_AND_REGISTER_MESSAGE` in your code 
> 
> 2) run the `generate-message-map` CMake target.
> 
> 
> 
> PowerShell Example:
> 
> ```
> 
> # from the vcpkg-tool directory
> 
> mkdir build; cd build
> 
> cmake .. -G Ninja
> 
> cmake --build . --target generate-message-map
> 
> ```
> 
> 
> 
> This will generate updated `locales/messages.json` and `locales/messages.en.json` that you can commit in your PR.
> 
> 
> 
> P.S.: We also have a useful `fmt` CMake target that will automatically format all code files using vcpkg's format rules.

Thanks a lot. I have fixed all the review comments. Could please help me review all my changes?

day253:
cc @ras0219-msft

coryan (approved):

day253:
cc @strega-nil-ms @vicroms @BillyONeal

src/vcpkg/binarycaching.cpp
```diff
@@ -25,19 +25,22 @@ using namespace vcpkg;
...
+    DECLARE_AND_REGISTER_MESSAGE(RestoredPackagesFromVendor,
+                                 (msg::count, msg::elapsed, msg::vendor),
                                  "",
-                                 "Restored {count} packages from AWS servers in {elapsed}");
-    DECLARE_AND_REGISTER_MESSAGE(AwsUploadedPackages, (msg::count), "", "Uploaded binaries to {count} AWS servers");
+                                 "Restored {count} packages from {vendor} servers in {elapsed}");
+    DECLARE_AND_REGISTER_MESSAGE(UploadedPackagesToVendor,
+                                 (msg::count, msg::vendor),
+                                 "",
+                                 "Uploaded binaries to {count} {vendor} servers");
```
vicroms, line 43:
```suggestion
                                 "Uploaded {count} binaries to {vendor} servers");
```

src/vcpkg/binarycaching.cpp
```diff
@@ -1095,222 +1066,154 @@ namespace
...
+                         msgObjectStorageToolFailedToDownload,
+                         msg::exit_code = out.exit_code,
+                         msg::tool_name = m_tool);
+            msg::write_unlocalized_text_to_stdout(Color::warning, out.output);
+            return false;
         }
-        const auto out = cmd_execute_and_capture_output(cmd);
-        if (out.exit_code == 0)
+
+        bool download_file(const std::string& object, const Path& archive) const override
```
vicroms, line 1119: Feels like `upload_file()` and `download_file()` could be abstracted into `bool copy_file(source, destination)`.  But I'm not against this bit of repetition so not asking for changes at this time.

ras0219-msft (approved):
I applied Victor's comment and also applied a handful of small tweaks:
- Use literals directly instead of saving them as std::string members
- Use StringLiteral/StringView over std::string
- Factor out `Command{paths.get_tool(...)}` as a member function.
- `msgObjectStorageToolFailed` is used for failures to upload OR download, so I made the message more generic

LGTM, thanks for the PR!

src/vcpkg/binarycaching.cpp
```diff
@@ -1094,222 +1069,141 @@ namespace
...
-            return Strings::concat(prefix, abi, ".zip");
-        }
-
-        void prefetch(View<Dependencies::InstallPlanAction> actions, View<CacheStatus*> cache_status) const override
-        {
-            auto& fs = paths.get_filesystem();
+        StringLiteral vendor() const override { return "AWS"; }
 
-            const auto timer = ElapsedTimer::create_started();
+        Command command() const { return Command{paths.get_tool_exe(Tools::AWSCLI)}; }
```
vicroms, line 1152: Should this include the `--no-sign-request` logic? That way it doesn't have to be repeated in every command.

ras0219-msft: I'm not versed enough with the aws cli to know whether it accepts that option on every subcommand

PR #459, opened 2022-03-27
Add --create-dirs to curl call in download_files_inner

horenmar:
Without this option, curl does not create intermediate directories
and thus trying to retrieve binary cache from az blob storage
fails if the `buildtrees` and its various subdirectories for ports
are not already present.

The alternative is for vcpkg to create the intermediate dirs
explicitly, but we already use the option in `try_download_file`,
so there is no reason not to use it here as well.

-----

Tests: not sure how to add tests for this, if you have any ideas
I don't mind adding them.

### DIFF
--- a/src/vcpkg/base/downloads.cpp
+++ b/src/vcpkg/base/downloads.cpp
@@ -353,6 +353,7 @@ namespace vcpkg
 
             Command cmd;
             cmd.string_arg("curl")
+                .string_arg("--create-dirs")
                 .string_arg("--location")
                 .string_arg("-w")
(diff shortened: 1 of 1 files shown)


horenmar:
This is almost definitely wrong, I happily went by the guid_marker and apparently multiple functions have the same one.

strega-nil-ms:
Thanks!

horenmar:
Btw, when will version with this change be released? This is a blocker for us.

strega-nil-ms:
@horenmar we're gonna hotfix https://github.com/microsoft/vcpkg/issues/23862 then crank the release

PR #465, opened 2022-03-29
Change guid_marker in download_files_inner to be unique

horenmar:
As far as I can tell, these are used as a combination of simple
disambiguation for the different curl callsites and a normalized
output prefix to be stripped later.

However, two functions, `url_heads_inner` and `download_files_inner`
ended up with the same guid, so I generated a new one for the
latter, so that they are unique again.


vicroms (approved):

ras0219-msft (approved):
LGTM and should have no functional change.

PR #486, opened 2022-04-04
[telemetry] Add additional context for some commands

vicroms:
To better understand how some of our commands are being used we need to capture additional telemetry.


src/vcpkg/commands.add.cpp
```diff
@@ -45,22 +47,30 @@ namespace
...
                                    msgAddArtifactOnlyOne,
                                    msg::command_line = "vcpkg add artifact");
-            std::string ce_args[] = {"add", args.command_arguments[1]};
+
+            auto artifact_name = args.command_arguments[1];
+
+            auto metrics = LockGuardPtr<Metrics>(g_metrics);
+            metrics->track_property("command_name", "add");
+            metrics->track_property("command_context", "artifact");
+            metrics->track_property("command_args", Hash::get_string_hash(artifact_name, Hash::Algorithm::Sha256));
```
vicroms, line 71: I think we need a file with the constants for "command_*" properties, since more commands will probably need to handle telemetry this way.

src/vcpkg/commands.add.cpp
```diff
@@ -45,22 +47,29 @@ namespace
...
-        args.parse_arguments(AddCommandStructure);
+        (void)args.parse_arguments(AddCommandStructure);
         auto&& selector = args.command_arguments[0];
+
         if (selector == "artifact")
         {
             Checks::msg_check_exit(VCPKG_LINE_INFO,
                                    args.command_arguments.size() <= 2,
                                    msgAddArtifactOnlyOne,
                                    msg::command_line = "vcpkg add artifact");
```
vicroms, line 64: Daniel pointed out that the ce's `add` does accept multiple arguments.
We should fix this behavior in a different PR.

dan-shaw (approved):

strega-nil-ms (approved):

BillyONeal (approved):

PR #517, opened 2022-04-22
Implement consistent registry access for vcpkg command accessing ports

theblackunknown:
## Description

This PR implements the behavior to consistently access a port and its versions for vcpkg command discussed in microsoft/vcpkg#18967

I consider this PR as a draft for now as I have only testing code changes on `x-add-version` and a new command `format-port` like it was suggested in the initial PR, and I am looking for feedbacks whether this is going into the right direction or not.

Although I have implement this wanted behavior as a separate struct `CommandRegistryPaths` containing `vcpkg::Path`, I feel like this would be more coherent if implemented directly as additional method on `vcpkg::RegistryImplementation` and `vcpkg::RegistryEntry`.
However as this is my first PR I lack context and experience with the code base to decide this is a good direction or not.

NOTE: This is only a partial implementation over the initial spec (e.g. read-only registry not implemented).


theblackunknown:
@ras0219-msft I take your emoji as a good sign, I'll keep this work going into the same direction !
Please share any feeedbacks whether this would be better to augment `RegistryXXX` classes instead.

PR #528, opened 2022-04-27
Fix hueristic -> heuristic typo in messages

horenmar:
I was poking around when I noticed the typo, so here it is fixed.


ras0219-msft:
Could you revert all changes to the `.lcl` files and the `messages.X.json` files other than `messages.json` and `messages.en.json`? All of these other files are owned by the localization team and will be automatically updated as part of the next localization batch.

horenmar:
I mean I can, but doesn't that leave the code in a state where the localization is temporarily broken?

ras0219-msft:
Yes; but:
1. We won't "ship" this repo into Microsoft/vcpkg before the localization files have been regenerated.
2. The localization isn't currently "shipping" yet

horenmar:
Ok

ras0219-msft:
Thanks a bunch for the continued fixes, they're very appreciated!

PR #538, opened 2022-05-10
get pipeline results

autoantwort:


src/vcpkg/sourceparagraph.cpp
```diff
@@ -1544,7 +1544,18 @@ namespace vcpkg
 
     static bool is_dependency_trivial(const Dependency& dep)
     {
-        return dep.features.empty() && dep.platform.is_empty() && dep.extra_info.is_empty() &&
+        if (Util::Vectors::contains(dep.features, "default"))
```
ras0219-msft, line 1547: I believe triviality now depends on the `depend-defaults` setting, because it changes the definition of `"zlib"` to imply `"default-features": false`.

That means that `"default-features": true` is no longer trivial if `"depend-defaults": true`.

autoantwort: In general this is true, but I think that this is not possible at the moment. A SourceControlFile is afaik only created by parsing a file currently. To set `default-features` to something different than `depend-defaults` you have to set `default-feautes`. This is saved in `Optional<bool> dep.default_features` and a dependency is non trivial when the optional has a value.  
But I can change this. This would then remove unnecessary `"default-features": true` when `depends-default` is true (the default)

autoantwort: But this is then wrong when using CONTROL files because there the `features` field has no entry `default` iiuc. 

Imho we should get rid off "default" and "core" in the features vector and make default_features a bool again or remove the default_features field completely and always add add "default" and "core" to the features vector. But currently this is a mess.

autoantwort: I think I will leave it as it is in this PR because it is working right now.

autoantwort: This would be easy to solve when https://github.com/microsoft/vcpkg-tool/pull/813 gets merged

autoantwort:
@ras0219-msft Since https://github.com/microsoft/vcpkg-tool/pull/813 was merged I merged with main. I have changed the implementation so that a `Dependency` is still context free. The downsite is that the parser is now not context free anymore, but we have to decide if a `Dependency` xor the parser should be context free or not (I personally prefer to have a context free description of a `Dependency`). I can change the code so that it is the other way around my making the `default_features` member of the `Dependency` an `Optional<bool>`.

autoantwort:
I have created https://github.com/microsoft/vcpkg-docs/pull/119 for documentation

silverqx:
So much effort invested in this and no reviews or merges, I would like to see this merged if the PR is correct and working with big packages like `qtbase` or `qt5-base`.

src/vcpkg/sourceparagraph.cpp
silverqx: > But currently this is a mess.

If this is true, then nothing to argue. 🙃

Arech:
Dear stakeholders, any progress on that?
This issue with ignoring `default-features:false` is a major source of pain for the development due to frequent unnecessary rebuilds. Also vcpkg CI doesn't cover some platforms at all such as QNX; lots of ports are buggy there and having minimal builds are the must..

mrexodia:
Also just ran into this issue, would be really good to have!

medns:
Also I have the same issue, need this pr!

dg0yt:
> @ras0219-msft This is open since nearly 1.5 years. Do you have an update? :)

Update: More than three years.

PR #544, opened 2022-05-20
Small cleanups for copy_file sendfile fix

hmoffatt:
Cleanups for #543


src/vcpkg/base/files.cpp
```diff
@@ -2898,6 +2898,8 @@ namespace vcpkg
             return true;
 #else // ^^^ (defined(__APPLE__) // !defined(__APPLE__) vvv
 
+            destination_fd.fchmod(source_stat.st_mode, ec);
```
dg0yt, line 2901: This is now generally *before* writing the output and for some paths before reading `ec`, instead of immediately before returning. IMO this is a behavioural change, not deduplication.

hmoffatt: Perhaps, but is there a need to delay the fchmod until the end, requiring the duplication?

dg0yt: Maybe it is really an improvement ;-)

ras0219-msft: I think this change LGTM, but it needs to add a check for `ec` and an early failure if the `fchmod` failed.

ras0219-msft (changes requested):

src/vcpkg/base/files.cpp
hmoffatt: Thanks that's a good point, I pushed a change.

ras0219-msft (approved):

PR #558, opened 2022-05-25
Add https binary cache

autoantwort:
This can be used to use gitlab generic packages as binary cache. 

We could also allow http. 

Maybe we should replace the header field by 'options to curl'. Then the `-u` option or multiple headers could be used.


include/vcpkg/binarycaching.h
```diff
@@ -69,6 +69,15 @@ namespace vcpkg
         virtual void precheck(View<Dependencies::InstallPlanAction> actions, View<CacheStatus*> cache_status) const = 0;
     };
 
+    struct UrlTemplate
+    {
+        std::string url_template;
+        std::vector<std::string> headers;
+
+        LocalizedString valid();
+        std::string instanciate_variables(const Dependencies::InstallPlanAction& action) const;
```
vicroms, line 78:
```suggestion
        std::string instantiate_variables(const Dependencies::InstallPlanAction& action) const;
```

Osyotr:
http support can be useful when pushing/fetching from internal server

BillyONeal (approved):

BillyONeal:
Thanks!

BillyONeal:
This completely broke binary caching for `x-azblob` because it causes curl command lines like:

```
curl --create-dirs --location -w "5ec47b8e-6776-4d70-b9b3-ac2a57bc0a1c %{http_code}\n" "https://vcpkgbinarycacheeastasia.blob.core.windows.net/cache/06db868bb3ae7a8a4d85685d799e0630d91a4594b9c1caa6dac361a42a46c407.zip?SAS" -o "C:\Dev\vcpkg\buildtrees\vcpkg-cmake\x64-windows.zip" -H "x-ms-version: 2020-04-08" -H "x-ms-blob-type: BlockBlob" "https://vcpkgbinarycacheeastasia.blob.core.windows.net/cache/dcc66cf2da4a2a8f4526bd8f779c455afb9eca970749bd0a123f8f1152b18630.zip?SAS" -o "C:\Dev\vcpkg\buildtrees\zlib\x86-windows.zip" -H "x-ms-version: 2020-04-08" -H "x-ms-blob-type: BlockBlob"
```

Note that `-H` is duplicated. https://curl.se/docs/manpage.html#-H `-H` is sent for all requests, not per-url like this appears to be trying to do.

BillyONeal:
image

autoantwort:
This are actually two issues: 
1. Headers are send twice
2. azure blob headers meant for put requests are also send in get requests (I didn't know that and thought it would be ok to also send them in get requests)

I have fixed this two issues in https://github.com/microsoft/vcpkg-tool/pull/592

PR #573, opened 2022-06-04
Optimize messages

Thomas1664:


Thomas1664:
CI fails because of "wrong comments" at stage localization. Why are the comments relevant during normal execution?
Is it possible to just turn them on when they're needed using `#define`?

Thomas1664:
The problem is that from `DECLARE_AND_REGISTER_MESSAGE` are methods called before `main` even starts. This causes hundreds of useless allocations that slow start-up notably down (by a few seconds). IMO the implementation of localization should be reworked to not do anything before `main` and to avoid allocations of comments and examples for localization messages at start-up because they are only needed for one internal command. 
If these allocations can't be avoided, the localization feature should run in another thread to not block everything on the main thread.

Thomas1664:
Marking ready for review to get some feedback by maintainers.

Thomas1664:
Initial example implementation of an alternative better approach to initialize localization:

```cpp
diff --git a/include/vcpkg/base/messages.h b/include/vcpkg/base/messages.h
index ebb3878b..089db3ef 100644
--- a/include/vcpkg/base/messages.h
+++ b/include/vcpkg/base/messages.h
@@ -11,6 +11,7 @@
 #include <string>
 #include <type_traits>
 #include <utility>
+#include <vector>
 
 namespace vcpkg
 {
@@ -121,6 +122,14 @@ VCPKG_FORMAT_WITH_TO_STRING(vcpkg::LocalizedString);
 
 namespace vcpkg::msg
 {
+    struct Msg
+    {
+        StringLiteral name;
+        StringLiteral extra_comment;
+        size_t index;
+        StringLiteral default_format_string;
+    };
+
     namespace detail
     {
         template<class Tag, class Type>
@@ -163,6 +172,20 @@ namespace vcpkg::msg
         ::size_t startup_register_message(StringLiteral name, StringLiteral format_string);
         ::size_t startup_register_message(StringLiteral name, StringLiteral format_string, std::string&& comment);
 
+        static std::vector<Msg*> global_messages;
+
+        template<typename T>
+        ::size_t register_msg(T* msg)
+        {
+            global_messages.emplace_back((Msg*)(msg));
+
+            auto x = msg->name;
+            auto y = msg->extra_comment;
+            auto a = msg->default_format_string;
+
+            return global_messages.size();
+        }
+
         ::size_t number_of_messages();
 
         // REQUIRES: index < last_message_index()
@@ -297,6 +320,7 @@ namespace vcpkg::msg
         using is_message_type = void;                                                                                  \
         static constexpr ::vcpkg::StringLiteral name = #NAME;                                                          \
+        static constexpr ::vcpkg::StringLiteral extra_comment = COMMENT;                                               \
         static const ::size_t index;                                                                                   \
         static constexpr ::vcpkg::StringLiteral default_format_string = __VA_ARGS__;                                   \
     } msg##NAME VCPKG_UNUSED = {}
@@ -304,8 +328,8 @@ namespace vcpkg::msg
 
 #ifndef VCPKG_GENERATE_LOCALIZATION
 #define REGISTER_MESSAGE(NAME)                                                                                         \
-    const ::size_t NAME##_msg_t::index =                                                                               \
-        ::vcpkg::msg::detail::startup_register_message(NAME##_msg_t::name, NAME##_msg_t::default_format_string)
+    const ::size_t NAME##_msg_t::index = ::vcpkg::msg::detail::register_msg(&msg##NAME);                                \
+    //::vcpkg::msg::detail::startup_register_message(NAME##_msg_t::name, NAME##_msg_t::default_format_string)
 #else
 #define REGISTER_MESSAGE(NAME)                                                                                         \
     const ::size_t NAME##_msg_t::index = ::vcpkg::msg::detail::startup_register_message(                               \
```

This saves the memory addresses of all strings to a static vector. The only problem is that I had to "trick" C++ because the message structs don't have a common base class.

BillyONeal:
I would like to see strong argument(s) that the performance change here is meaningful to increase the "configuration state space" by introducing another option like this. How does this affect actual startup time?

Thomas1664:
> 

To be clear: by startup time I mean the time from the first `DECLARE_AND_REGISTER_MESSAGE` to the beginning of `main`.
Without this PR, startup time is 11 ms. With my changes it's just 5 ms. This is a improvement of ~50%.

However, I don't like the solution that much. Therefore I explored another way to decrease startup time. You can see my first results in the diff of my last comment. Basically, `REGISTER_MESSAGE` stores the memory address of its corresponding struct inside a global static vector.
I would prefer the changes in the diff but I had to introduce a "base type" of the types introduced by `DECLARE_MESSAGE`. The problem is that I can't have this class as a real base class of these structs (Do you know a workaround?) therefore, I cast the pointers to type `Msg`. This allows us to do any calculations from within `main` and even sets the baseline to do this on a worker thread.

BillyONeal:
> To be clear: by startup time I mean the time from the first `DECLARE_AND_REGISTER_MESSAGE` to the beginning of `main`.
> Without this PR, startup time is 11 ms. With my changes it's just 5 ms. This is an improvement of ~50%.

Before: 11.52 ms

image

After: 11.41 ms

image

both figures are very noisy and I'm not sure how "real" this result is. I don't know if creating additional "compile state space" is worth a <1% improvement here.

BillyONeal:
> You can see my first results in the diff of my last comment. Basically, `REGISTER_MESSAGE` stores the memory address of its corresponding struct inside a global static vector.

That seems broken because it would create different vectors per TU and thus messages would be duplicated.

Thomas1664:
> both figures are very noisy and I'm not sure how "real" this result is. I don't know if creating additional "compile state space" is worth a <1% improvement here.

Your values are probably more "real" than mine because I just went lucky with 1 iteration and you did 100. On the other hand, execution becomes faster if you let the application run twice. 11 ms can't be measured reliably because if you wait a few seconds between 2 runs and maybe even do something different on your computer, the difference is up to 5 ms. No real user runs vcpkg 100 times in a loop.

Thomas1664:
> > You can see my first results in the diff of my last comment. Basically, `REGISTER_MESSAGE` stores the memory address of its corresponding struct inside a global static vector.
> 
> That seems broken because it would create different vectors per TU and thus messages would be duplicated.

This won't duplicate the messages because I'm just taking in raw pointers. Is there a way to only have 1 vector for the entire application?

Thomas1664:
> Is there a way to only have 1 vector for the entire application?

Rhetorical question: https://stackoverflow.com/questions/14349877/static-global-variables-in-c

ras0219-msft:
What about simply deferring the creation of the examples string until it's requested? Meaning, store some constexpr data structure in place of the examples and only perform the full std::string creation if the subcommand to emit the localization doc is called.

Thomas1664:
> What about simply deferring the creation of the examples string until it's requested? Meaning, store some constexpr data structure in place of the examples and only perform the full std::string creation if the subcommand to emit the localization doc is called.

It's less efficient in comparison to have a global vector with just pointers to the structs. With 3 vectors you still allocate 3 times and probably allocating more space than just for the pointers. I would prefer a combination of @ras0219-msft's idea and my idea using a global vector, if it's ok for you to type-pun the message structs.

Thomas1664:
@BillyONeal Are you ok with my proposed solution after fixing the global vector?

BillyONeal:
> @BillyONeal Are you ok with my proposed solution after fixing the global vector?

I don't understand what your proposed solution is, since it seems to depend on being able to touch a global vector from headers, and I am unaware of how something like that can.

Thomas1664:
> I don't understand what your proposed solution is, since it seems to depend on being able to touch a global vector from headers, and I am unaware of how something like that can.

@BillyONeal My proposed solution is to have a global vector that stores the memory location of every message struct. This can be done as [this stack-overflow post](https://stackoverflow.com/questions/14349877/static-global-variables-in-c) describes by declaring the vector as extern in the header file and as static in the source file:

`messages.h`:
```cpp
extern std::vector<Msg> msg_store;
```

`messages.cpp`:
```cpp
static std::vector<Msg> msg_store;
```

Note that the vector won't be accessed from a header file (just like you can't use `REGISTER_MESSAGE` in a header).

Then I want to do the necessary calculations (i.e. filling in the other vectors as needed) in `threadunsafe_initialize_context()` and in `commands.generate-message-map.cpp`.

BillyONeal:
That seems functionally identical to how this currently works?

Thomas1664:
> That seems functionally identical to how this currently works?

Yes, with the following differences
- Comments/ example string only in `commands.generate-message-map.cpp`
- Don't do anything (except `msg_store`) if localization isn't enabled

BillyONeal:
>Don't do anything (except msg_store) if localization isn't enabled (if that's possible because we have to know we're not running x-generate-default-message-map)

I am opposed to adding "compilation modes" like this without a demonstration of meaningful performance gains by doing so, which I don't see here. I certainly don't see an 11ms -> 5ms reduction as described above.

Thomas1664:
> >Don't do anything (except msg_store) if localization isn't enabled (if that's possible because we have to know we're not running x-generate-default-message-map)
> 
> I am opposed to adding "compilation modes" like this without a demonstration of meaningful performance gains by doing so, which I don't see here. I certainly don't see an 11ms -> 5ms reduction as described above.

This won't be a compilation mode. There are already dedicated ENV vars.

BillyONeal:
In that case I don't have strong opinions. (I still doubt it's making a meaningful perf difference though)

BillyONeal:
I think https://github.com/microsoft/vcpkg-tool/pull/615 probably changes how you would want to approach changes like this; are you still interested in pursuing this? /cc @JavierMatosD

Thomas1664:
> I think #615 probably changes how you would want to approach changes like this; are you still interested in pursuing this? /cc @JavierMatosD

@BillyONeal 
Not really. #615 is not what I would call a "redesign". @JavierMatosD basically just moved all message definitions to `message.h`. The underlying interface stayed the same.

However, as most allocations come from concatenating example strings, I think the best way to tear down allocations is a similar implementation as #622 for `Strings::concat`. The problem is that some types that are fed into this aren't can't be casted to `StringView`. The first step would be to make `char` and `LocalizedString` work with `StringView`.

BillyONeal:
> @BillyONeal Not really. #615 is not what I would call a "redesign".

OK fair.

> @JavierMatosD basically just moved all message definitions to `message.h`. The underlying interface stayed the same.

Yes, but how you may want to approach getting the improvements described herein is likely substantially different now that that stuff need not be part of the header. Yes, the example strings didn’t move to the “REGISTER” layer, but you could do that which would let you make it lazy if you want, etc.

> However, as most allocations come from concatenating example strings, I think the best way to tear down allocations is a similar implementation as #622 for `Strings::concat`. The problem is that some types that are fed into this aren't can't be casted to `StringView`. The first step would be to make `char`, `int`, `size_t`, and `DOWRD` work with `StringView`.

Formatting things that aren’t strings necessarily will involve allocations, as will performing concats, etc.

Thomas1664:
@BillyONeal I think I don't have to change anything except marking `startup_register_message`, `get_examples_for_args`, and `format_examples_for_args` as `constexpr`. Everything these function need (i.e. the localized string itself, examples for arguments) is known at compile time. 

Furthermore, we would need a global message variable: `static constexpr Message message;`

However, that would require C++ 20 because of `constexpr` for `std::string` and `std::vector`. We need gcc 10 for C++ 20.

Thomas1664:
> I don't think that helps you meaningfully because there are no constepxr std::strings.

Turns out that you're right. I just saved 1 allocation out of 14.

The function that causes most of the allocations is `detail::format_examples_for_args`. As I mentioned in previous comments I would like to move calls to this function somewhere where we have control on. The problem I see in this context is how to save the information for the example string that is passed to `get_examples_for_args(StringView extra_comment, const MessageCheckFormatArgs<Arg0, Args...>&)`?
The only idea I have is changing the type of the example string vector to be a `std::vector<std::pair<std::string, std::vector<FormatArgAbi>>>`. However, this still causes 3 allocations.
Ideally I want to store this information somewhere in the message struct itself and not as a template argument where the information is gone at runtime.

PR #604, opened 2022-06-24
[VCPKG-ARTIFACTS] Fixes to make Espressif IDF work (with artifacts for dependencies) 

fearthecowboy:
Includes work from @BillyONeal 's fork for espressif.

- when a registry is updated, the zip file will check to see how deep to strip off folders that should not be there 
- made registry loading async
- fix git clone to be smarter/faster 
- make the 'localname' for a repo a bit cleaner for github and aka.ms registries, otherwise generate a smaller name 

To test this:

``` powershell
# ensure we have vcpkg stable
iex (iwr -useb https://aka.ms/vcpkg-init.ps1)

#clone the blink repo
git clone https://github.com/some-example/blink
cd blink

# run the devel version of vcpkg:
C:\work\2022\vcpkg-tool\out\build\x64-Debug\vcpkg.ps1 activate
```

## output
``` text
warning: vcpkg-ce ('configure environment') is experimental and may change at any time.
Using in-development vcpkg-artifacts built at: C:/work/2022/vcpkg-tool/ce/ce
Updating registry data from https://github.com/fearthecowboy/registries/releases/download/espressif-idf-1.10/Espressif.zip
Updating registry data from https://aka.ms/vcpkg-ce-default
 Artifact                                                                                                             Version                Status        Dependency  Summary
 espressif:idf                                                                                                        1.10.2                 will install  *           Espressif IDF
 microsoft:tools/kitware/cmake                                                                                        3.20.1                 will install  *           Kitware's cmake tool
 microsoft:tools/ninja-build/ninja                                                                                    1.10.2                 will install  *           Ninja is a small build system with a focus on speed.
 [https://github.com/fearthecowboy/registries/releases/download/espressif-idf-1.10/Espressif.zip]:ccache              4.3.0                  will install  *
 [https://github.com/fearthecowboy/registries/releases/download/espressif-idf-1.10/Espressif.zip]:dfu_util            0.9.0                  will install  *
 [https://github.com/fearthecowboy/registries/releases/download/espressif-idf-1.10/Espressif.zip]:riscv32-esp-elf     8.4.0-patch3           will install  *
 [https://github.com/fearthecowboy/registries/releases/download/espressif-idf-1.10/Espressif.zip]:openocd             0.11.0-esp32-20211120  will install  *
 [https://github.com/fearthecowboy/registries/releases/download/espressif-idf-1.10/Espressif.zip]:xtensa-esp32s2-elf  8.4.0-patch3           will install  *
 [https://github.com/fearthecowboy/registries/releases/download/espressif-idf-1.10/Espressif.zip]:xtensa-esp32s3-elf  8.4.0-patch3           will install  *
 [https://github.com/fearthecowboy/registries/releases/download/espressif-idf-1.10/Espressif.zip]:xtensa-esp32-elf    8.4.0-patch3           will install  *
 [https://github.com/fearthecowboy/registries/releases/download/espressif-idf-1.10/Espressif.zip]:esp32s2ulp-elf      2.28.51-esp20191205    will install  *
 [https://github.com/fearthecowboy/registries/releases/download/espressif-idf-1.10/Espressif.zip]:esp32ulp-elf        2.28.51-esp20191205    will install  *

```


BillyONeal:
ROFL it's actually "github.com/some-example" 🤣😂

fearthecowboy:
Yeah, I got tired of not having a place for 1-off examples of stuff so I made a new org for myself.

BillyONeal:
I merged https://github.com/microsoft/vcpkg-tool/commit/426593ed2671a6b766042fc090a231094b47efb5 then merged main into this to get the different-PR content out.

ce/ce/amf/installer.ts
```diff
@@ -220,6 +220,14 @@ class GitCloneNode extends Installer implements GitInstaller {
     this.setMember('subdirectory', value);
   }
 
+  get espidf() {
+    return this.flags.has('espidf');
+  }
+
+  set espidf(value: boolean) {
+    this.flags.set('espidf', value);
```
BillyONeal, line 228: Doesn't this need to unset if false and set if true? The "get" says true on "has".

ce/ce/archivers/ZipUnpacker.ts
```diff
@@ -49,6 +49,19 @@ export class ZipUnpacker extends Unpacker {
 
       const openedFile = await archiveUri.openFile();
       const zipFile = await ZipFile.read(openedFile);
+      if (options.strip === -1) {
```
BillyONeal, line 52: It seems like we should do this for all extractors that support "strip", not just zip.

ce/ce/archivers/ZipUnpacker.ts
```diff
@@ -49,6 +49,19 @@ export class ZipUnpacker extends Unpacker {
...
+        // when strip == -1, strip off all common folders off the front of the file names
+        options.strip = 0;
+        const folders = [...zipFile.folders.keys()].sort((a, b) => a.length - b.length);
+        const files = [...zipFile.files.keys()];
+        for (const folder of folders) {
+          if (files.all((filename) => filename.startsWith(folder))) {
+            options.strip = folder.split('/').length - 1;
```
BillyONeal, line 59: Do we need to worry about \\?

ce/ce/archivers/git.ts
```diff
@@ -105,20 +96,27 @@ export class Git {
...
       '-C',
-      this.#targetFolder,
+      this.#targetFolder.fsPath,
       'checkout',
       options.commit ? options.commit : ''
     ], {
-      env: this.#environment
+      env: this.#environment,
+      cwd: this.#targetFolder.fsPath,
+      onStdErrData: (chunk) => {
```
BillyONeal, line 105: This lambda is repeated at least twice; extract as a function and dedupe?

ce/ce/installers/espidf.ts
```diff
@@ -0,0 +1,114 @@
...
+    'install-python-env'
+  ], {
+    env: extendedEnvironment
+  });
+
+  if (installPythonEnv.code) {
+    return false;
+  }
+
+  return true;
```
BillyONeal, line 65: Elsewhere you made this:

```suggestion
  return installPythonEnv.code === 0;
```

ce/ce/archivers/git.ts
```diff
@@ -59,11 +58,7 @@ export class Git {
       }
     });
 
-    if (result.code) {
-      return false;
-    }
-
-    return true;
+    return result.code === 0 ? true : false;
```
BillyONeal, line 61:
```suggestion
    return result.code === 0;
```

ce/ce/archivers/git.ts
```diff
@@ -74,24 +69,20 @@ export class Git {
...
+      env: this.#environment,
+      cwd: this.#targetFolder.fsPath
     });
 
-    if (result.code) {
-      return false;
-    }
-
-    return true;
+    return result.code === 0 ? true : false;
```
BillyONeal, line 85:
```suggestion
    return result.code === 0;
```

ce/ce/archivers/git.ts
```diff
@@ -129,19 +127,123 @@ export class Git {
...
+        const regex = /\s([0-9]*?)%/;
+        chunk.toString().split(/^/gim).map((x: string) => x.trim()).filter((each: any) => each).forEach((line: string) => {
+          const match_array = line.match(regex);
+          if (match_array !== null) {
+            events.heartbeat?.(line.trim());
+          }
+        });
+      }
     });
+    return result.code === 0 ? true : false;
```
BillyONeal, line 149:
```suggestion
    return result.code === 0;
```

ce/ce/archivers/git.ts
```diff
@@ -129,19 +127,123 @@ export class Git {
...
+      throw new Error(`${this.#targetFolder.fsPath} is not a directory.`);
     }
 
-    return true;
+    const result = await execute(this.#toolPath, ['init'], {
+      env: this.#environment,
+      cwd: this.#targetFolder.fsPath
+    });
+
+    return result.code === 0 ? true : false;
```
BillyONeal, line 171:
```suggestion
    return result.code === 0;
```

ce/ce/archivers/git.ts
```diff
@@ -129,19 +127,123 @@ export class Git {
...
+      'remote',
+      'add',
+      name,
+      location.toString()
+    ], {
+      env: this.#environment,
+      cwd: this.#targetFolder.fsPath
+    });
+
+    return result.code === 0 ? true : false;
```
BillyONeal, line 193:
```suggestion
    return result.code === 0;
```

ce/ce/archivers/git.ts
```diff
@@ -129,19 +127,123 @@ export class Git {
...
+        chunk.toString().split(/^/gim).map((x: string) => x.trim()).filter((each: any) => each).forEach((line: string) => {
+          const match_array = line.match(regex);
+          if (match_array !== null) {
+            events.heartbeat?.(line.trim());
+          }
+        });
+      }
+    });
+
+    return result.code === 0 ? true : false;
```
BillyONeal, line 226:
```suggestion
    return result.code === 0;
```

ce/ce/archivers/git.ts
```diff
@@ -129,19 +127,123 @@ export class Git {
...
+      'config',
+      '-f',
+      this.#targetFolder.join(configFile).fsPath,
+      key,
+      value
+    ], {
+      env: this.#environment,
+      cwd: this.#targetFolder.fsPath
+    });
+    return result.code === 0 ? true : false;
```
BillyONeal, line 247:
```suggestion
    return result.code === 0;
```

ce/ce/artifacts/artifact.ts
```diff
@@ -232,6 +234,17 @@ export class Artifact extends ArtifactBase {
       this.session.activation.addExports(exportsBlock, this.targetLocation);
     }
 
+    // if espressif install
+    if (this.metadata.info.flags.has('espidf')) {
```
BillyONeal, line 238: Is that get/set property earlier in the review used now? Since this is reaching into flags directly.

ce/ce/archivers/git.ts
```diff
@@ -129,19 +127,123 @@ export class Git {
...
+   * Adds a remote location to the git repo.
+   * @param name the name of the remote to add.
+   * @param location the location of the remote to add.
+   * @returns true if the addition was successful, false otherwise.
+   */
+  async addRemote(name: string, location: Uri) {
+    const result = await execute(this.#toolPath, [
+      '-C',
+      this.#targetFolder.fsPath,
+      'remote',
```
BillyONeal, line 184: It may be worth adding a comment here that you're adding a remote and pulling that because you want submodules to work, unlike something like

https://github.com/microsoft/vcpkg-tool/blob/35c8e299dfd7b75c4b458204edc3290cdb325e6b/src/vcpkg/vcpkgpaths.cpp#L1149

BillyONeal (changes requested):

ce/ce/archivers/ZipUnpacker.ts
fearthecowboy: I don't believe so - all zip files folder names I've found had forward slashes - but I can test and see if there is a chance that back slashes are in them.

ce/ce/amf/installer.ts
fearthecowboy: set's implementation does that 

> in flags.ts
```  typescript
set(flag: string, value: boolean) {
    this.assert(true);
    if (value) {
      this.node.add(new Scalar(flag));
    } else {
      this.node.delete(flag);
    }
  }
```

BillyONeal: Sounds good! I thought it was just an array rather than a thing...

ce/ce/archivers/ZipUnpacker.ts
fearthecowboy: I could add that for tar based ones, but it would require going thru the whole file twice - the filenames in a tar archive are spread out the whole way.  (The zip implementation gets them up front regardless).

I guess it only has to do it if `-1 ` is specified. 

So the question is how bad do we want that? Lemme see how fast I can do that.

BillyONeal: Can we just make it something other than "strip" then? Like "remove-prefixes" or something, so that what users see has the same behavior when it's the same name. (Or just make the tar extractor explicitly say "I can't do that" when a user does)

ce/ce/artifacts/artifact.ts
fearthecowboy: Ah! no, it's not -- the espidf flag there was when we had it nested in the `git clone ` ..

I've deleted that code several times, but it keeps creeping back in when I pull from someone else 😢 

lemme remove that

ce/ce/archivers/git.ts
fearthecowboy: ok

ce/ce/archivers/ZipUnpacker.ts
BillyONeal: (That is to say: if we are saying "strip means this" it should mean that everywhere and emit diagnostics if that's impossible)

ce/ce/archivers/ZipUnpacker.ts
fearthecowboy: Forward slashes are required for paths in zip files in the spec. 

image

ce/ce/archivers/ZipUnpacker.ts
fearthecowboy: Updated tar to match zip extractor behavior for guessing the folders to strip

ce/ce/archivers/git.ts
fearthecowboy: added to where it's used (`ce/ce/installers/git.ts`)

ce/ce/artifacts/artifact.ts
BillyONeal: This one doesn't look changed?

BillyONeal (approved):

ce/ce/artifacts/artifact.ts
fearthecowboy: This is the correct usage, the other has been removed.

PR #608, opened 2022-06-28
Fix artifact search to work well with multiple versions of artifacts and registries

fearthecowboy:
Searching artifacts on a specific registry didn't work quite right.

This makes it so if the user prefixes the search criteria with a registry ie `vcpkg find artifact arm:gcc` that it will just search the given registry.
It also prints a message when there are no search results at all.

And when it uses the current internal ce downloader, puts the version and hash in the filename because otherwise it would clobber the download file when it switches between versions.


ce/ce/installers/util.ts
```diff
@@ -16,9 +16,17 @@ export function artifactFileName(name: string, install: Installer, extension: st
     result += '-';
     result += install.lang;
   }
+  // add the version number into the filename too.
+  result += '-' + (version || '');
```
BillyONeal, line 20: Isn't the `||` dead because it isn't `string?`?

ce/ce/cli/commands/find.ts
```diff
@@ -42,14 +43,30 @@ export class FindCommand extends Command {
...
+        }
+        for (const result of artifacts) {
+          if (!result.metadata.info.dependencyOnly) {
+            const name = artifactIdentity(result.registryId, id, result.shortName);
+            table.push(name, result.metadata.info.version, result.metadata.info.summary || '');
+          }
         }
       }
     }
+    if (!table.anyRows) {
```
BillyONeal, line 66: Why is this a member of table rather than just a local here?

(Not asking to change it, asking to understand myself)

ce/ce/cli/commands/find.ts
```diff
@@ -42,14 +43,30 @@ export class FindCommand extends Command {
...
-        const latest = artifacts[0];
-        if (!latest.metadata.info.dependencyOnly) {
-          const name = artifactIdentity(latest.registryId, id, latest.shortName);
-          table.push(name, latest.metadata.info.version, latest.metadata.info.summary || '');
+      // eslint-disable-next-line prefer-const
+      for (let [registry, id, artifacts] of await registries.search({
+        // use keyword search if no registry is specified
+        keyword: each.indexOf(':') === -1 ? each : undefined,
+        // otherwise use the criteria as an id
+        idOrShortName: each.indexOf(':') > -1 ? each : undefined,
```
BillyONeal, line 51: 1. Shouldn't idOrShortName be the part after the : rather than the whole thing?
2. Could you introduce `const registrySpecified = each.indexOf(':') !== -1;` which would avoid the need for some of the comments?

BillyONeal (commented):
Can you show a screenshot of before/after, and/or add a test for the new behavior?

ce/ce/cli/commands/find.ts
fearthecowboy: I figured that it was reusable; but not important to be so.

fearthecowboy:
image

VS

image

ce/ce/installers/util.ts
fearthecowboy: Yeah, that seems a smidge redundant. fixing

ce/ce/cli/commands/find.ts
fearthecowboy: The actual search uses the full value with the colon to scope the search

BillyONeal (approved):

PR #619, opened 2022-07-08
Localization commands integrate

JavierMatosD:
Handles all the localization for commands.integrate.cpp. The only exception is anything that utilizes the HelpTableFormatter. Since the formatter is used in many locations it will be done in a separate PR.


include/vcpkg/base/messages.h
```diff
@@ -991,6 +1006,10 @@ namespace vcpkg
                     "{value} is a localized message name like LocalizedMessageMustNotEndWithNewline",
                     "The message named {value} ends with a newline which should be added by formatting "
                     "rather than by localization.");
+    DECLARE_MESSAGE(MissingCompletionDirectory,
+                    (msg::value, msg::path, msg::error),
+                    "'{value}' is the target completion. Ex: fish, zsg, bash, etc.. '{error}' is the error code.",
```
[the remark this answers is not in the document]
JavierMatosD: Sorry, that is a typo, zsh, and I agree the comment needs more context.

include/vcpkg/base/messages.h
```diff
@@ -409,6 +409,8 @@ namespace vcpkg
                     (msg::command_line),
                     "",
                     "The first argument to '{command_line}' must be 'artifact' or 'port'.");
+    DECLARE_MESSAGE(AddingCompletionEntry, (msg::path), "", "Adding vcpkg completion entry to '{path}'.");
+    DECLARE_MESSAGE(AddingVcpkgCompletion, (msg::path), "", "Adding vcpkg completion entry to '{path}'.");
```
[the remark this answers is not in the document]
vicroms: I usually don't add quotes to paths, although the argument to double quote them for copy-paste into terminal is valid.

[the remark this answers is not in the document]
BillyONeal: Right, I'm saying either it should not have additional single quotes (like the original product code) or double-quotes-and-correct-escaping-for-pasting-into-a-terminal should happen.

BillyONeal, line 413: We shouldn't be adding 2 of the same message?

include/vcpkg/base/messages.h
```diff
@@ -1055,6 +1083,10 @@ namespace vcpkg
         "{value} may be either a 'vendor' like 'Azure' or 'NuGet', or a file path like C:\\example or /usr/example",
         "Restored {count} package(s) from {value} in {elapsed}. Use --debug to see more details.");
     DECLARE_MESSAGE(ResultsHeader, (), "Displayed before a list of installation results.", "RESULTS");
+    DECLARE_MESSAGE(ScriptFailed,
```
BillyONeal, line 1086: I think you can reuse CommandFailed for this one?

src/vcpkg/commands.integrate.cpp
```diff
@@ -362,27 +363,17 @@ namespace vcpkg::Commands::Integrate
...
-All MSBuild C++ projects can now #include any installed libraries.
-Linking will be handled automatically.
-Installing new libraries will make them instantly available.
-
-CMake projects should use: "-DCMAKE_TOOLCHAIN_FILE=%s"
-)",
-            cmake_toolchain.generic_u8string());
+        msg::println(
+            msg::format(msgCMakeToolChainFile, msg::command_name = cmake_toolchain.generic_u8string())
+                .append_raw("\nAll MSBuild C++ projects can now #include any installed libraries. Linking will be "
```
BillyONeal, line 372: This needs to be localized.

src/vcpkg/commands.integrate.cpp
```diff
@@ -396,14 +387,14 @@ CMake projects should use: "-DCMAKE_TOOLCHAIN_FILE=%s"
...
+        auto message = msg::format(msgUserWideIntegration);
         if (was_deleted)
         {
-            print2(Color::success, "User-wide integration was removed\n");
+            msg::println(message.append_raw("was removed."));
         }
         else
         {
-            print2(Color::success, "User-wide integration is not installed\n");
+            msg::println(message.append_raw("is not installed."));
```
BillyONeal, line 397: These need to be localized as separate messages rather than a single message with additional stuff.  (And "was removed" is certainly not localized :D)

Generally speaking, composing messages that would be printed as totally separate lines is OK, but not something you expect to be the same sentence. (Because where you're appending may not be the right place to say that in all languages)

src/vcpkg/commands.integrate.cpp
```diff
@@ -444,27 +435,20 @@ CMake projects should use: "-DCMAKE_TOOLCHAIN_FILE=%s"
             cmd_execute_and_capture_output(cmd_line, default_working_directory, get_clean_environment()), Tools::NUGET);
         if (!maybe_nuget_output)
         {
-            Checks::exit_with_message(
-                VCPKG_LINE_INFO, "Error: NuGet package creation failed: %s\n", maybe_nuget_output.error());
+            msg::println_error(msgNugetPackageCreationFailed, msg::error = maybe_nuget_output.error());
```
BillyONeal, line 438: Ditto can we just reuse `CommandFailed`?

BillyONeal (changes requested):

include/vcpkg/base/messages.h
```diff
@@ -993,9 +1012,17 @@ namespace vcpkg
...
                     (msg::path, msg::exit_code),
                     "",
-                    "msiexec failed while extracting '{path}' with launch or exit code {exit_code} and message:");
+                    "msiexec failed while extracting \"{path}\" with launch or exit code {exit_code} and message:");
     DECLARE_MESSAGE(NoLocalizationForMessages, (), "", "No localized messages for the following: ");
     DECLARE_MESSAGE(NoRegistryForPort, (msg::package_name), "", "no registry configured for port {package_name}");
+    DECLARE_MESSAGE(NugetPackageCreationFailed,
+                    (msg::error),
+                    "'{error}' is the NuGet output message.",
+                    "NuGet package creation failed: '{error}' ");
```
BillyONeal, line 1021:
```suggestion
```

This doesn't appear to be referenced?

include/vcpkg/base/messages.h
```diff
@@ -993,9 +1012,17 @@ namespace vcpkg
...
+    DECLARE_MESSAGE(NugetPackageFileCreationFailed,
+                    (msg::path),
+                    "",
+                    "NuGet package creation failed. No .nupkg was produced. \"{path}\" ");
```
BillyONeal, line 1025:
```suggestion
    DECLARE_MESSAGE(NugetPackageFileSucceededButCreationFailed,
                    (msg::path),
                    "",
                    "NuGet package creation succeeded, but no .nupkg was produced. Expected: \"{path}\"");
```

I think this change lost the impact of the original product code's meaning that we got a zero exit code from running `nuget pack` but there was no `nupkg` for some reason; this may prompt the user to take different troubleshooting steps or that sort of thing.

include/vcpkg/base/messages.h
```diff
@@ -1016,6 +1043,7 @@ namespace vcpkg
                     (msg::package_name),
                     "",
                     "the baseline does not contain an entry for port {package_name}");
+    DECLARE_MESSAGE(PreviousIntegrationFileRemains, (), "", "Previous integration file was not removed.");
```
BillyONeal, line 1046: I like this name!

include/vcpkg/base/messages.h
```diff
@@ -1081,6 +1113,10 @@ namespace vcpkg
                     "unknown binary provider type: valid providers are 'clear', 'default', 'nuget', "
                     "'nugetconfig','nugettimeout', 'interactive', 'x-azblob', 'x-gcs', 'x-aws', "
                     "'x-aws-config', 'http', and 'files'");
+    DECLARE_MESSAGE(UnknownParameterForIntegrate,
+                    (msg::value),
+                    "'{value}' is the parameter",
```
BillyONeal, line 1118:
```suggestion
                    "'{value}' is a user-supplied command line option. For example, given vcpkg integrate frobinate, {value} would be frobinate.",
```

Feel free to use another nonsense word than frobinate :)

include/vcpkg/base/messages.h
```diff
@@ -1152,6 +1190,11 @@ namespace vcpkg
                     (),
                     "",
                     "vcpkg-ce ('configure environment') is experimental and may change at any time.");
+    DECLARE_MESSAGE(
+        VcpkgCompletion,
+        (msg::value, msg::path),
+        "'{value}' is the subject for completion. i.e. bash, zsh, etc.",
+        "vcpkg '{value}' completion is already imported to your \"{path}\" file.\nThe following entries were found: ");
```
BillyONeal, line 1197:
```suggestion
        "vcpkg {value} completion is already imported to your \"{path}\" file.\nThe following entries were found:");
```

I'm surprised that our validator let you get away with the trailing whitespace ....

BillyONeal (changes requested):
Strongly request-changes for the CMakeSettings.json change, the other things are nitpicks.

BillyONeal (approved):

PR #631, opened 2022-07-15
Localize binaryparagraph.cpp and sourceparagraph.cpp

JavierMatosD:
Localizing:
- binaryparagraph.cpp
- sourceparagraph.cpp


include/vcpkg/base/messages.h
```diff
@@ -510,6 +510,24 @@ namespace vcpkg
                     "An example of env_var is \"HTTP(S)_PROXY\""
                     "'--' at the beginning must be preserved",
                     "-- Automatically setting {env_var} environment variables to \"{url}\".");
+    DECLARE_MESSAGE(
+        BinaryParagraphDescription,
+        (msg::spec,
+         msg::version,
+         msg::version,
+         msg::value,
```
BillyONeal, line 518: Ditto from somewhere else I don't think you can reuse replacements like this. Between this and this PR being draft not going to review further for now...

include/vcpkg/base/messages.h
```diff
@@ -510,6 +510,33 @@ namespace vcpkg
...
+                    (msg::value),
+                    "'{value}' is the value of feature in the binary paragraph.",
+                    "\nfeature: '{value}'");
+    DECLARE_MESSAGE(BinParagraphMaintainers,
+                    (msg::value),
+                    "'{value}' is the binary paragraph's maintainers.",
+                    "\nmaintainers: ['{value}']");
+    DECLARE_MESSAGE(BinParagraphPortVersion, (msg::version), "", "\nport_version: {version}");
+    DECLARE_MESSAGE(BinParagraphSpecAndVersion, (msg::spec, msg::version), "", "\nspec: '{spec}'\nversion: {version}");
+    DECLARE_MESSAGE(BinParagraphType, (msg::value), "'{value}' is the binary paragraph type", "\ntype: '{value}'");
```
BillyONeal, line 539: These are intended to be a dump of an on disk file for us to review so I don't think we should localize them.

include/vcpkg/base/messages.h
```diff
@@ -741,6 +769,10 @@ namespace vcpkg
                     "",
                     "Visual Studio Code was not found and the environment variable {env_var} is not set or invalid.");
     DECLARE_MESSAGE(ErrorVsCodeNotFoundPathExamined, (), "", "The following paths were examined:");
+    DECLARE_MESSAGE(ErrorWhileParsing,
+                    (msg::value),
+                    "'{value}' is the file name.",
```
BillyONeal, line 774: If it's the file name, should it not be `msg::path`? Alternately this should just say `vcpkg-configuration.json` directly in the message text.

include/vcpkg/base/messages.h
```diff
@@ -751,6 +783,12 @@ namespace vcpkg
     DECLARE_MESSAGE(ExpectedPortName, (), "", "expected a port name here");
     DECLARE_MESSAGE(ExpectedTripletName, (), "", "expected a triplet name here");
     DECLARE_MESSAGE(ExtendedDocumentationAtUrl, (msg::url), "", "Extended documentation available at '{url}'.");
+    DECLARE_MESSAGE(FailedToParseBinParagraph,
+                    (msg::url, msg::error_msg, msg::value),
+                    "'{error_msg}' is the error message for failing to parse the Binary Paragraph, '{value}' is the "
+                    "Binary Paragraph",
+                    "[sanity check] Failed to parse a serialized binary paragraph.\nPlease open an issue at {url}, "
```
BillyONeal, line 790: Is there a different URL we would ever use here or should it just be part of the message?

include/vcpkg/base/messages.h
```diff
@@ -751,6 +783,12 @@ namespace vcpkg
...
+                    "with the following output:\n{error_msg}\n=== Serialized BinaryParagraph ===\n{value}");
```
BillyONeal, line 791: This ==== looks like attempting to format in the localized text which I think we should avoid; note the other comment about the serialized thing probably not being a thing that should be localized.

include/vcpkg/base/messages.h
JavierMatosD: What if the url changes in the future?

include/vcpkg/base/messages.h
```diff
@@ -764,8 +766,15 @@ namespace vcpkg
     DECLARE_MESSAGE(ExpectedFailOrSkip, (), "", "expected 'fail', 'skip', or 'pass' here");
     DECLARE_MESSAGE(ExpectedPortName, (), "", "expected a port name here");
     DECLARE_MESSAGE(ExpectedTripletName, (), "", "expected a triplet name here");
-    DECLARE_MESSAGE(ExtendedDocumenationAtUrl, (msg::url), "", "Extended documentation available at '{url}'.");
+    DECLARE_MESSAGE(ExtendedDocumentationAtUrl, (msg::url), "", "Extended documentation available at '{url}'.");
```
BillyONeal, line 769: Good catch!

include/vcpkg/base/messages.h
```diff
@@ -764,8 +766,15 @@ namespace vcpkg
...
     DECLARE_MESSAGE(FailedToExtract, (msg::path), "", "Failed to extract \"{path}\":");
+    DECLARE_MESSAGE(FailedToParseBinParagraph,
+                    (msg::error_msg, msg::value),
+                    "'{error_msg}' is the error message for failing to parse the Binary Paragraph, '{value}' is the "
+                    "Binary Paragraph",
+                    "[sanity check] Failed to parse a serialized binary paragraph.\nPlease open an issue at "
+                    "https://github.com/microsoft/vcpkg, "
+                    "with the following output:\n{error_msg}\nSerialized Binary Paragraph: \n{value}");
```
BillyONeal, line 777:
```suggestion
                    "with the following output:\n{error_msg}\nSerialized Binary Paragraph:");
```

(and just `append_raw` at the call site)

src/vcpkg/sourceparagraph.cpp
```diff
@@ -1383,17 +1383,15 @@ namespace vcpkg
                     [](const std::unique_ptr<ParseControlErrorInfo>& ppcei) { return !ppcei->extra_fields.empty(); }))
             {
                 Strings::append(message,
-                                "This is the list of valid fields for CONTROL files (case-sensitive): \n\n    ",
+                                msg::format(msgListOfValidFieldsForControlFiles).data(),
```
BillyONeal, line 1386:
```suggestion
                                msg::format(msgListOfValidFieldsForControlFiles).extract_data(),
```

BillyONeal (changes requested):

include/vcpkg/base/messages.h
JavierMatosD: Why for this instance of a newline and not for others?

src/vcpkg/sourceparagraph.cpp
```diff
@@ -1441,11 +1439,10 @@ namespace vcpkg
             }
             if (start_of_chunk != end_of_chunk)
             {
-                print2(StringView{start_of_chunk, end_of_chunk});
+                msg::println_error(LocalizedString::from_raw(StringView{start_of_chunk, end_of_chunk}));
```
BillyONeal, line 1442: This should be:

```
msg::write_unlocalized_text_to_stdout(StringView{start_of_chunk, end_of_chunk});
```

src/vcpkg/sourceparagraph.cpp
```diff
@@ -1616,11 +1613,12 @@ namespace vcpkg
             auto maybe_configuration = reader.visit(*configuration, get_configuration_deserializer());
             if (!reader.errors().empty())
             {
-                print2(Color::error, "Errors occurred while parsing ", ManifestDeserializer::VCPKG_CONFIGURATION, "\n");
+                msg::println_error(msgErrorWhileParsing, msg::path = ManifestDeserializer::VCPKG_CONFIGURATION);
```
BillyONeal, line 1616: This is wrong, it should be listing the file on each line with an error, but it faithfully preserves the previous behavior. It does add "error: " in front of everything but I think that's OK. (Here's the old behavior:)

BillyONeal: image

src/vcpkg/binaryparagraph.cpp
```diff
@@ -280,16 +280,10 @@ namespace vcpkg
...
-                                       R"([sanity check] Failed to parse a serialized binary paragraph.
-Please open an issue at https://github.com/microsoft/vcpkg, with the following output:
-    Error: %s
-
-=== Serialized BinaryParagraph ===
-%s
-            )",
-                                       parsed_paragraph.error(),
-                                       my_paragraph);
+            Checks::msg_exit_maybe_upgrade(
```
BillyONeal, line 283: `[sanity check]` makes it sound like a debugging message, but it clearly has user instructions, so it makes sense.

That said, it seems like line 293 should get the same treatment?

BillyONeal (approved):

src/vcpkg/binaryparagraph.cpp
JavierMatosD: > `[sanity check]` makes it sound like a debugging message, but it clearly has user instructions, so it makes sense.
> 
> That said, it seems like line 293 should get the same treatment?

Please check out the related change. :)

src/vcpkg/binaryparagraph.cpp
```diff
@@ -289,56 +289,30 @@ namespace vcpkg
...
+                                               .append_raw(format_binary_paragraph(binary_paragraph)));
         }
     }
+
+    std::string format_binary_paragraph(BinaryParagraph paragraph)
+    {
+        const auto& join_str = R"(", ")";
+        return Strings::format(
+            "\nspec: \"%s\"\nversion: \"%s\"\nport_version: %d\ndescription: [\"%s\"]\nmaintainers: [\"%s\"]\nfeature: "
+            "\"%s\"\ndefault_features: [\"%s\"]\ndependencies: [\"%s\"]\nabi: \"%s\"\ntype: %s",
```
BillyONeal, line 306:
```suggestion
        return fmt::format(
            "\nspec: \"{}\"\nversion: \"{}\"\nport_version: {}\ndescription: [\"{}\"]\nmaintainers: [\"{}\"]\nfeature: "
            "\"{}\"\ndefault_features: [\"{}\"]\ndependencies: [\"{}\"]\nabi: \"{}\"\ntype: {}",
```

src/vcpkg/binaryparagraph.cpp
```diff
@@ -289,56 +289,30 @@ namespace vcpkg
...
+                                               .append(msgOriginalBinParagraphHeader)
+                                               .append_raw(format_binary_paragraph(pgh))
+                                               .append(msgSerializedBinParagraphHeader)
+                                               .append_raw(format_binary_paragraph(binary_paragraph)));
         }
     }
+
+    std::string format_binary_paragraph(BinaryParagraph paragraph)
+    {
+        const auto& join_str = R"(", ")";
```
BillyONeal, line 303:
```suggestion
        constexpr StringLiteral join_str = R"(", ")";
```

BillyONeal (approved):

PR #632, opened 2022-07-15
Localize build.cpp

JavierMatosD:


CMakeSettings.json
```diff
@@ -30,6 +30,11 @@
           "name": "VCPKG_BUILD_FUZZING",
           "value": "True",
           "type": "BOOL"
+        },
+        {
+          "name": "CLANG_FORMAT",
+          "value": "C:/Program Files/LLVM/bin/clang-format.exe",
+          "type": "FILEPATH"
         }
```
BillyONeal, line 38:
```suggestion
```

:wink:

include/vcpkg/base/messages.h
```diff
@@ -741,6 +741,7 @@ namespace vcpkg
                     "",
                     "Visual Studio Code was not found and the environment variable {env_var} is not set or invalid.");
     DECLARE_MESSAGE(ErrorVsCodeNotFoundPathExamined, (), "", "The following paths were examined:");
+    DECLARE_MESSAGE(ErrorWhileWriting, (msg::error_msg), "", "Error occured while writing {error_msg}");
```
BillyONeal, line 744:
```suggestion
    DECLARE_MESSAGE(ErrorWhileWriting, (msg::path), "", "Error occurred while writing {path}");
```

In the product code this is the path to the file being written.

include/vcpkg/base/messages.h
```diff
@@ -751,6 +752,10 @@ namespace vcpkg
     DECLARE_MESSAGE(ExpectedPortName, (), "", "expected a port name here");
     DECLARE_MESSAGE(ExpectedTripletName, (), "", "expected a triplet name here");
     DECLARE_MESSAGE(ExtendedDocumentationAtUrl, (msg::url), "", "Extended documentation available at '{url}'.");
+    DECLARE_MESSAGE(FailedToFindDependencyAbi,
```
BillyONeal, line 755: Hmmm not sure how I feel about this one; I think it should be a `Debug::` message.

What's going on here is, when computing the ABI hash for a package, that contains the ABI hashes of the dependencies:

image

To get this error message, vcpkg would have to somehow decide that this dependency needed to be installed, then also somehow be unable to find the dependency when calculating this hash. For that to happen something has seriously gone off the rails and I don't think a message beyond one targeted at vcpkg's developers themselves is going to do much...

include/vcpkg/base/messages.h
```diff
@@ -885,10 +890,15 @@ namespace vcpkg
...
                     "invalid argument: binary config '{binary_source}' requires a SAS token without a "
                     "preceeding '?' as the second argument");
+    DECLARE_MESSAGE(InvalidBuildInfo, (msg::error_msg), "", "Invalid BUILD_INFO file for package: {error_msg}");
     DECLARE_MESSAGE(InvalidFormatString,
                     (msg::actual),
                     "{actual} is the provided format string",
                     "invalid format string: {actual}");
+    DECLARE_MESSAGE(InvalidLinkage,
+                    (msg::system_name, msg::value),
+                    "'{value}' is the linkage type.",
```
BillyONeal, line 900:
```suggestion
                    "'{value}' is the linkage type vcpkg did not understand. (Correct values would be static or dynamic)",
```

include/vcpkg/base/messages.h
```diff
@@ -1081,6 +1097,14 @@ namespace vcpkg
                     "unknown binary provider type: valid providers are 'clear', 'default', 'nuget', "
                     "'nugetconfig','nugettimeout', 'interactive', 'x-azblob', 'x-gcs', 'x-aws', "
                     "'x-aws-config', 'http', and 'files'");
+    DECLARE_MESSAGE(UnknownPolicySetting,
+                    (msg::option, msg::value),
+                    "'{value}' is the policy.",
```
BillyONeal, line 1102:
```suggestion
                    "'{value}' is the policy in question. These are unlocalized names that ports use to control post build checks. Some examples are VCPKG_POLICY_DLLS_WITHOUT_EXPORTS, VCPKG_POLICY_MISMATCHED_NUMBER_OF_BINARIES, or VCPKG_POLICY_ALLOW_OBSOLETE_MSVCRT",
```

BillyONeal (changes requested):
Waiting for the clang-format thing again :)

src/vcpkg/build.cpp
```diff
@@ -914,11 +913,11 @@ namespace vcpkg
             return_code = cmd_execute_and_stream_data(
                 command,
                 [&](StringView sv) {
-                    print2(sv);
-                    Checks::check_exit(VCPKG_LINE_INFO,
-                                       out_file.write(sv.data(), 1, sv.size()) == sv.size(),
-                                       "Error occurred while writing '%s'",
-                                       stdoutlog);
+                    msg::println(LocalizedString::from_raw(sv));
```
BillyONeal, line 916:
```suggestion
                    msg::write_unlocalized_text_to_stdout(sv);
```

This is actually broken because you're adding additional newlines that aren't present in the chunk.

BillyONeal (changes requested):
The extra ln from the println needs to be fixed since it will garble the output.

BillyONeal (approved):

PR #636, opened 2022-07-18
[localization] Create example strings at compile time

Thomas1664:
This PR gets rid of almoast every allocation during initialization of the localization feature. Remaining allocations come from the extension of the vectors.

I also added a new type called `StringArray` which is basically a `std::array<char, N>`. This is needed because the existing `StringLiteral` doesn't provide a way to add string literals using just `constexpr` functions. Furthermore, I think `StringLiteral` should be derived from `std::array` anyway because it literly is a `const char[]`.

My changes don't affect the API used to initialize the localization comments that much as I stated in #573.


Thomas1664:
@BillyONeal Is this a false positive warning?

```
D:\a\1\s\include\vcpkg\base\stringview.h(115): error C2220: the following warning is treated as an error
D:\a\1\s\include\vcpkg\base\stringview.h(120) : warning C28020: The expression '0<=_Param_(1)&&_Param_(1)<=1-1' is not true at this call.: Lines: 116, 118, 120, 118, 120
D:\a\1\s\include\vcpkg\base\stringview.h(123) : warning C6294: Ill-defined for-loop.  Loop body not executed.
D:\a\1\s\include\vcpkg\base\stringview.h(120) : warning C28020: The expression '0<=_Param_(1)&&_Param_(1)<=1-1' is not true at this call.: Lines: 116, 118, 120, 118, 120
D:\a\1\s\include\vcpkg\base\stringview.h(120) : warning C28020: The expression '0<=_Param_(1)&&_Param_(1)<=1-1' is not true at this call.: Lines: 116, 118, 120, 118, 120
D:\a\1\s\include\vcpkg\base\stringview.h(120) : warning C28020: The expression '0<=_Param_(1)&&_Param_(1)<=1-1' is not true at this call.: Lines: 116, 118, 120, 118, 120
```

include/vcpkg/base/messages.h
```diff
@@ -63,7 +63,7 @@ namespace vcpkg
             return *this;
         }
         template<class Message, class... Args>
-        LocalizedString& append(Message m, const Args&... args)
+        constexpr LocalizedString& append(Message m, const Args&... args)
```
BillyONeal, line 66: This requires C++20 (and even then a very recent C++20 compiler since constexpr string is taking a long time to land), so I don't think we can do that. https://en.cppreference.com/w/cpp/compiler_support claims gcc 12 / msvc 2022 / clang 15. (Unavailable for Apple Clang entirely)

include/vcpkg/base/messages.h
```diff
@@ -148,16 +148,38 @@ namespace vcpkg::msg
...
+
+            const auto out = comment + StringArray(" ") + example;
+            return std::string(out.begin(), out.end());
         }
 
+        inline constexpr auto get_examples() { return StringArray{""}; }
+
         template<class Arg0, class... Args>
-        std::string get_examples_for_args(StringView extra_comment, const MessageCheckFormatArgs<Arg0, Args...>&)
+        inline constexpr auto get_examples(const Arg0& arg, Args... args)
```
BillyONeal, line 171: Does this damage the check that the format string is valid? (Note `MessageCheckFormatArgs` in the old code which failed to compile if the replacement strings were wrong)

include/vcpkg/base/stringview.h
```diff
@@ -86,3 +87,47 @@ namespace vcpkg
...
+constexpr std::array<char, N - 1> to_array(const char (&a)[N], std::index_sequence<Is...>)
+{
+    return {{a[Is]...}};
+}
+
+template<std::size_t N>
+constexpr std::array<char, N - 1> to_array(const char (&a)[N])
+{
+    return to_array(a, std::make_index_sequence<N - 1>());
+}
```
BillyONeal, line 101: I don't understand what this is trying to do. If I'm reading it correctly, it always gives you a std::array substantially the same as the input, so you could have said `std::array{a}` for the same effect. Or alternately just left it as an array. Did you mean to do null termination or something in here or...?

include/vcpkg/base/stringview.h
```diff
@@ -86,3 +87,47 @@ namespace vcpkg
...
+{
+    if constexpr (lhs.empty())
+        return rhs;
+    else
+    {
+        StringArray<L + R - 1> out;
+
+        for (size_t i = 0; i < lhs.size(); i++)
+        {
+            out.at(i) = lhs[i];
```
BillyONeal, line 124:
```suggestion
            out[i] = lhs[i];
```

Ditto below. It's strange to use `at()` and `operator[]` in the same line and I can't imagine why we wouldn't just always use `[]`.

include/vcpkg/base/stringview.h
```diff
@@ -86,3 +87,47 @@ namespace vcpkg
...
+        }
+
+        for (size_t i = L; i < L + R - 1; i++)
+        {
+            out.at(i - 1) = rhs[i - L];
```
BillyONeal, line 129: Where is the -1 coming from?

BillyONeal (changes requested):
TL;DR: I like this direction, but I think we can't do it until C++20 is more widely deployed.

> @BillyONeal Is this a false positive warning?
> 
> ```
> D:\a\1\s\include\vcpkg\base\stringview.h(115): error C2220: the following warning is treated as an error
> D:\a\1\s\include\vcpkg\base\stringview.h(120) : warning C28020: The expression '0<=_Param_(1)&&_Param_(1)<=1-1' is not true at this call.: Lines: 116, 118, 120, 118, 120
> D:\a\1\s\include\vcpkg\base\stringview.h(123) : warning C6294: Ill-defined for-loop.  Loop body not executed.
> D:\a\1\s\include\vcpkg\base\stringview.h(120) : warning C28020: The expression '0<=_Param_(1)&&_Param_(1)<=1-1' is not true at this call.: Lines: 116, 118, 120, 118, 120
> D:\a\1\s\include\vcpkg\base\stringview.h(120) : warning C28020: The expression '0<=_Param_(1)&&_Param_(1)<=1-1' is not true at this call.: Lines: 116, 118, 120, 118, 120
> D:\a\1\s\include\vcpkg\base\stringview.h(120) : warning C28020: The expression '0<=_Param_(1)&&_Param_(1)<=1-1' is not true at this call.: Lines: 116, 118, 120, 118, 120
> ```
> 
> If not: What did I do wrong?

I'm not sure. See my comment about "where did the -1 come from", I observe the analyzer is correct if lhs.size() == 1 but rhs.size() == 0. If the invariant is that StringArray ends with a null, then the empty test doesn't make sense and trying to form StringArray<0> should be ill formed.

It's also possible that the analyzer just doesn't understand what you're doing here. Perhaps something like:

```
template<::size_t L, ::size_t R>
constexpr StringArray<L + R - 1> operator+(const StringArray<L> lhs, const StringArray<R> rhs) noexcept
{
    // -1 because lhs's null terminator is removed
    StringArray<L + R - 1> out;
    auto after_left = std::copy(lhs.begin(), lhs.end() - 1, out.begin());
    std::copy(rhs.begin(), rhs.end(), after_left)
    return out;
}
```

would make more sense?

Thomas1664:
> It's also possible that the analyzer just doesn't understand what you're doing here. Perhaps something like:
> 
> ```
> template<::size_t L, ::size_t R>
> constexpr StringArray<L + R - 1> operator+(const StringArray<L> lhs, const StringArray<R> rhs) noexcept
> {
>     // -1 because lhs's null terminator is removed
>     StringArray<L + R - 1> out;
>     auto after_left = std::copy(lhs.begin(), lhs.end() - 1, out.begin());
>     std::copy(rhs.begin(), rhs.end(), after_left)
>     return out;
> }
> ```
> 
> would make more sense?

Yes but I can't use `std::copy` because it's not `constexpr` in 17

include/vcpkg/base/messages.h
Thomas1664: Yes, I tried to run the CMake target locally and it turned out that not all strings are valid

Thomas1664: @BillyONeal Do you have any suggestions on how to enable this check again?

BillyONeal:
> Yes, I tried to use `std::copy` in the first place but I can't use it because it's not `constexpr` in 17. Functions not being `constexpr` really sucks.

You are using constexpr string which is a C++20 feature though

Thomas1664:
> > Yes, I tried to use `std::copy` in the first place but I can't use it because it's not `constexpr` in 17. Functions not being `constexpr` really sucks.
> 
> You are using constexpr string which is a C++20 feature though

I'm not using constexpr string, I'm using std::array<char, n> as a base type, which is in 17 constexpr. I had to to this trick to avoid std::string. If I would have used constexpr string, this wouldn't have compiled. Note that I didn't upgrade the CXX standard - it's still at 17.

If you think I'm using some features from 20, please tell me the exact lines. Can you explain why my code compiles with 17?

include/vcpkg/base/messages.h
Thomas1664: Removed constexpr here.

Thomas1664: @BillyONeal Now I see where confusion about C++ 20 came from: I already removed constexpr here. Note that this change had no impact on my approach.

include/vcpkg/base/messages.h
```diff
@@ -148,16 +148,66 @@ namespace vcpkg::msg
 
         std::string format_examples_for_args(StringView extra_comment, const FormatArgAbi* args, std::size_t arg_count);
 
-        inline std::string get_examples_for_args(StringView extra_comment, const MessageCheckFormatArgs<>&)
+        template<::size_t M, ::size_t N>
+        inline std::string get_examples_for_args(const StringArray<M>& comment, const StringArray<N>& example)
```
Thomas1664, line 152: This function combines the comment and the already concatenated example strings.

include/vcpkg/base/messages.h
```diff
@@ -148,16 +148,66 @@ namespace vcpkg::msg
...
+        {
+            if constexpr (Arg0::real_example().length() == 0)
+                return StringArray{""};
+            else
+                return StringArray{" "} + arg.real_example();
         }
 
         template<class Arg0, class... Args>
-        std::string get_examples_for_args(StringView extra_comment, const MessageCheckFormatArgs<Arg0, Args...>&)
+        constexpr auto example_piece(const Arg0& arg, Args... args)
```
Thomas1664, line 188: This is basically the same as `Strings::concat`: It concatenates the example strings and inserts spaces where necessary.

include/vcpkg/base/messages.h
```diff
@@ -148,16 +148,66 @@ namespace vcpkg::msg
...
+            else
+                return StringArray{" "} + arg.real_example() + example_piece(args...);
+        }
+
+        inline constexpr auto get_examples() { return StringArray{""}; }
+
+        template<class Arg0, class... Args>
+        inline constexpr auto get_examples(const Arg0& arg, Args... args)
+        {
+            if constexpr (sizeof...(args) == 0)
```
Thomas1664, line 201: Do not enter `example_piece` if there is only one example (that lives inside `arg`)

include/vcpkg/base/messages.h
```diff
@@ -232,7 +282,15 @@ namespace vcpkg::msg
     constexpr static struct NAME##_t                                                                                   \
     {                                                                                                                  \
         constexpr static const char* name = #NAME;                                                                     \
-        constexpr static const char* example = EXAMPLE;                                                                \
+        constexpr static StringArray example = EXAMPLE;                                                                \
+        constexpr static StringArray example_str = "An example of {" #NAME "} is " EXAMPLE ".";                        \
+        constexpr static auto real_example()                                                                           \
```
Thomas1664, line 287: If no example is provided, return the empty example, otherwise return the already formatted example string. I don't think we need the member `example` of this struct anymore.

include/vcpkg/base/messages.h
```diff
@@ -288,16 +346,17 @@ namespace vcpkg::msg
     {                                                                                                                  \
         using is_message_type = void;                                                                                  \
         static constexpr ::vcpkg::StringLiteral name = #NAME;                                                          \
-        static constexpr ::vcpkg::StringLiteral extra_comment = COMMENT;                                               \
+        static constexpr StringArray extra_comment = COMMENT;                                                          \
         static constexpr ::vcpkg::StringLiteral default_format_string = __VA_ARGS__;                                   \
         static const ::size_t index;                                                                                   \
+        static constexpr StringArray example_str = vcpkg::msg::detail::get_examples ARGS;                              \
```
Thomas1664, line 352: This calls the function `get_examples`. The `()` part is hidden inside the macro parameter.

include/vcpkg/base/stringview.h
```diff
@@ -86,3 +88,45 @@ namespace vcpkg
 
 VCPKG_FORMAT_AS(vcpkg::ZStringView, vcpkg::StringView);
 VCPKG_FORMAT_AS(vcpkg::StringLiteral, vcpkg::StringView);
+
+template<::size_t N>
+struct StringArray : public std::array<char, N>
```
Thomas1664, line 93: This is the struct that makes the concatenations possible: It inherits from `std::array` (maybe I should make the array a class member rather than inheriting from it [react with 👍if you want me to do that]).

I'm unsure where (in terms of namespace and header file) to put this because there might be other cases where this struct is useful so I didn't want to put it into `messages.h`.

include/vcpkg/base/stringview.h
```diff
@@ -86,3 +88,45 @@ namespace vcpkg
...
+    constexpr StringArray() noexcept : std::array<char, N>() { }
+    constexpr StringArray(const char (&str)[N]) noexcept : std::array<char, N>()
+    {
+        for (size_t i = 0; i < N; i++)
+        {
+            this->at(i) = str[i];
+        }
+    }
+
+    constexpr ::size_t length() const noexcept { return this->size() - 1; }
```
Thomas1664, line 104: -1 because of null termination. I had to implement it to prevent `std::array<char, 0>`.

include/vcpkg/base/stringview.h
```diff
@@ -86,3 +88,45 @@ namespace vcpkg
...
+        return rhs;
+    else if constexpr (rhs.length() == 0)
+        return lhs;
+    else
+    {
+        StringArray<L + R - 1> out;
+
+        for (size_t i = 0; i < lhs.length(); i++)
+        {
+            out[i] = lhs[i];
```
Thomas1664, line 123: Can't use `std::copy` because it's not `constexpr` in C++17

src/vcpkg/base/messages.cpp
```diff
@@ -196,12 +196,12 @@ namespace vcpkg::msg
             {
                 for (size_t j = i + 1; j < m.names.size(); ++j)
                 {
-                    if (m.names[i] == m.names[j])
+                    if (msg::detail::get_message_name(i) == msg::detail::get_message_name(j))
```
Thomas1664, line 199: Changes in this file are not needed by the thing I tried to implement

include/vcpkg/base/stringview.h
BillyONeal: Add:

```
template<class InIt, class OutIt>
constexpr OutIt constexpr_copy(InIt first, InIt last, OutIt dest) {
  while (first != last) {
    *dest = *first;
    ++dest;
    ++first;
  }

  return dest;
}
```

and use that?

include/vcpkg/base/stringview.h
BillyONeal, line 104:
```suggestion
    constexpr ::size_t size() const noexcept { return this->size() - 1; }
```

The STL convention for containers is that they have `size`, not `length`.

include/vcpkg/base/stringview.h
BillyONeal: I prefer to see the std::array be a member rather than use inheritance, since several of the std::array members, including size, do the wrong thing.

include/vcpkg/base/stringview.h
```diff
@@ -86,3 +88,45 @@ namespace vcpkg
 
 VCPKG_FORMAT_AS(vcpkg::ZStringView, vcpkg::StringView);
 VCPKG_FORMAT_AS(vcpkg::StringLiteral, vcpkg::StringView);
+
+template<::size_t N>
+struct StringArray : public std::array<char, N>
+{
```
BillyONeal, line 94:
```suggestion
{
    static_assert(N != 0, "Space for the null terminator is necessary.");
```

include/vcpkg/base/stringview.h
```diff
@@ -86,3 +88,45 @@ namespace vcpkg
...
+    constexpr StringArray() noexcept : std::array<char, N>() { }
+    constexpr StringArray(const char (&str)[N]) noexcept : std::array<char, N>()
+    {
+        for (size_t i = 0; i < N; i++)
```
BillyONeal, line 98: Please add the copy-alike from https://github.com/microsoft/vcpkg-tool/pull/636/files#r927156486 and implement this with that.

include/vcpkg/base/stringview.h
```diff
@@ -86,3 +88,45 @@ namespace vcpkg
...
+
+template<::size_t L, ::size_t R>
+constexpr StringArray<L + R - 1> operator+(const StringArray<L> lhs, const StringArray<R> rhs) noexcept
+{
+    if constexpr (lhs.length() == 0)
+        return rhs;
+    else if constexpr (rhs.length() == 0)
+        return lhs;
+    else
+    {
```
BillyONeal, line 118:
```suggestion
```

I don't think those do anything.

BillyONeal (commented):
I like this direction and you resolved my C++20 concerns. I would like to see the addition of `constexpr_copy` and the couple of loops replaced with that, then I think this is OK.

@JavierMatosD please review this since it interacts with the localization work you're currently doing.

include/vcpkg/base/messages.h
BillyONeal: I think you've done what can be done already.

include/vcpkg/base/stringview.h
Thomas1664: Yes, but there is `std::string.length()`

BillyONeal: Right, but that's a legacy holdover from before `std::string` gained an STL container interface.

include/vcpkg/base/stringview.h
Thomas1664: No joke, `static_assert` is not evaluated to constexpr😂. Tgerefore, I have to use `enable_if`.

Thomas1664: I ended up using `std::enable_if`

BillyONeal: That doesn't have the desired behavior though; we want attempting to form StringArray<0> to be a hard error with a message describing why, not SFINAE out.

Thomas1664:
I profiled the optimizations in this PR. The result was that this PR resulted in a ~6,75 times performance improvement!

(measured on Windows 10 using `.\vcpkg contact --no-sendmetrics`; before: ~27000 ms; after: ~4000 ms)

dg0yt:
> 



> I profiled the optimizations in this PR. The result was that this PR resulted in a ~6,75 times performance improvement!
> 
> (measured on Windows 10 using `.\vcpkg contact --no-sendmetrics`; before: ~27000 ms; after: ~4000 ms)

Is this a representative workload? How about `./vcpkg depend-info qt5-base:x64-windows`? Or at least a frequently used command like `x-add-version` or `format-manifest`?

Thomas1664:
> Is this a representative workload? How about `./vcpkg depend-info qt5-base:x64-windows`? Or at least a frequently used command like `x-add-version` or `format-manifest`?

Yes because the affected code happens even before `main` is called. When I run commands that do even more than just printing something out, the variance increases.

Thomas1664:
On GCC 9.4 Debug I get about 5% improvement.

BillyONeal:
> ~~(measured on Windows 10 using `.\vcpkg contact --no-sendmetrics`; before: ~27000 ms; after: ~4000 ms)~~

I don't understand this situation where you're waiting 4 seconds to run a vcpkg command; I'm seeing 2 orders of magnitude less.

image

Is there something funny going on with this machine? I would understand some variance in how fast machines are but not 100 times differences.

> On GCC 9.4 Debug I get about 5% improvement.

That's closer to what I would expect although still not 4000ms.

Thomas1664:
> I don't understand this situation where you're waiting 4 seconds to run a vcpkg command; I'm seeing 2 orders of magnitude less.

> Is there something funny going on with this machine? I would understand some variance in how fast machines are but not 100 times differences.

Maybe it came to the attention of my antivirus that an unsigned executable creates and deletes several files?
The reason for the huge performance improvement on Windows was that I measured an executable that I built the previous day. However, everything that I built on that day had the same bad performance.

dg0yt:
> I don't understand this situation where you're waiting 4 seconds to run a vcpkg command

I thought this was refering to 10000 runs... 
A single execution on linux:
~~~
$ time ./vcpkg contact --no-sendmetrics
Send an email to vcpkg@microsoft.com with any feedback.

real    0m0,001s
user    0m0,001s
sys     0m0,000s
~~~

The same on windows, MSYS2 bash:
~~~
$ time ./vcpkg contact  --no-sendmetrics
Send an email to vcpkg@microsoft.com with any feedback.

real    0m0.118s
user    0m0.000s
sys     0m0.015s
~~~

BillyONeal:
> > I don't understand this situation where you're waiting 4 seconds to run a vcpkg command
> 
> I thought this was refering to 10000 runs... A single execution on linux:

That would explain it :D

Thomas1664:
> I like this direction and you resolved my C++20 concerns. I would like to see the addition of `constexpr_copy` and the couple of loops replaced with that, then I think this is OK.
> 
> @JavierMatosD please review this since it interacts with the localization work you're currently doing.

@JavierMatosD Ready for review

JavierMatosD (approved):

BillyONeal (approved):

Thomas1664:
@BillyONeal Looks like you messed it up. The CI failure doesn't come from merging main because `LicenseExpressionContainsUnicode` was not changed.

```
"C:\Windows\system32\cmd.exe" /D /E:ON /V:OFF /S /C "CALL "D:\a\_temp\b0229f60-ea9f-4a31-beb3-3206a7923b36.cmd""
error: message LicenseExpressionContainsUnicode has an incorrect comment:
{value} was specified in a comment, but was not used in the message.
error: message LicenseExpressionContainsUnicode has an incorrect comment:
{value} was specified in a comment, but was not used in the message.
```

BillyONeal:
Megaderp, sorry

PR #646, opened 2022-07-21
[Localization] Ursa Minor 🐻

JavierMatosD:
Localized messages in the following files: 
- export.chocolatey.cpp
- dependencies.cpp
- configuration.cpp

It also adds a new message argument: supports_expression


BillyONeal:
https://ryanhenwood.bandcamp.com/track/ursa-major

include/vcpkg/base/messages.h
```diff
@@ -767,8 +776,21 @@ namespace vcpkg
     DECLARE_MESSAGE(ExpectedFailOrSkip, (), "", "expected 'fail', 'skip', or 'pass' here");
     DECLARE_MESSAGE(ExpectedPortName, (), "", "expected a port name here");
     DECLARE_MESSAGE(ExpectedTripletName, (), "", "expected a triplet name here");
+    DECLARE_MESSAGE(ExportingPackage, (msg::package_name), "", "Exporting {package_name}...");
     DECLARE_MESSAGE(ExtendedDocumenationAtUrl, (msg::url), "", "Extended documentation available at '{url}'.");
     DECLARE_MESSAGE(FailedToExtract, (msg::path), "", "Failed to extract \"{path}\":");
+    DECLARE_MESSAGE(FailedToFindPortFeature,
+                    (msg::value, msg::spec),
+                    "'{value}' is a port feature name.",
+                    "Could not find feature {value} in port {spec}.");
```
BillyONeal, line 785:
```suggestion
                    "Could not find {value} in {spec}.");
```

I know the original code said feature and port here but I don't think it's necessary in this context.

Also consider reusing this message for https://github.com/microsoft/vcpkg-tool/blob/7d292feb55656816952b68efd4a1a738839d6c78/src/vcpkg/dependencies.cpp#L254-L259 (or using the language from there here too).

If there are more of these we could add `{feature}` or similar. (I wasn't opposed to adding `feature_spec` in our chat; I was only commenting that we didn't need it for that example)

include/vcpkg/base/messages.h
```diff
@@ -919,6 +943,10 @@ namespace vcpkg
                     (msg::actual),
                     "{actual} is the provided format string",
                     "invalid format string: {actual}");
+    DECLARE_MESSAGE(InvalidPackageFeature,
+                    (msg::package_name, msg::value),
+                    "'{value}' is a port feature.",
+                    "Port {package_name} does not have a {value} feature");
```
BillyONeal, line 949: As I said in the other one, I think this should be merged with `FailedToFindPortFeature`.

include/vcpkg/base/messages.h
```diff
@@ -1054,10 +1086,21 @@ namespace vcpkg
                     (msg::path),
                     "",
                     "NuGet package creation succeeded, but no .nupkg was produced. Expected: \"{path}\"");
+    DECLARE_MESSAGE(OptionRequired, (msg::option), "", "--{option} option is required.");
     DECLARE_MESSAGE(PackageFailedtWhileExtracting,
                     (msg::value, msg::path),
                     "'{value}' is either a tool name or a package name.",
                     "'{value}' failed while extracting {path}.");
+    DECLARE_MESSAGE(PackagesToInstall, (), "", "The following packages will be built and installed:");
+    DECLARE_MESSAGE(PackagesToInstallDirectly, (), "", "The following packages will be directlt installed:");
```
BillyONeal, line 1095:
```suggestion
    DECLARE_MESSAGE(PackagesToInstallDirectly, (), "", "The following packages will be directly installed:");
```

src/vcpkg/dependencies.cpp
```diff
@@ -829,17 +826,19 @@ namespace vcpkg
...
                                 m_var_provider.get_dep_info_vars(spec.spec()).value_or_exit(VCPKG_LINE_INFO)))
                         {
-                            const auto msg = Strings::format("%s[%s] is only supported on '%s'",
-                                                             spec.port(),
-                                                             spec.feature(),
-                                                             to_string(*supports_expression.get()));
+                            auto localized_msg =
+                                msg::format(msgUnsupportedPortFeature,
+                                            msg::package_name = spec.port(),
+                                            msg::value = spec.feature(),
```
BillyONeal, line 832: I think you should change `UnsupportedPortFeature` to be:

```
"{spec} is only supported on '{supports_expression}'"
```

and then

```suggestion
                                            msg::spec = spec,
```

See https://github.com/microsoft/vcpkg-tool/blob/7d292feb55656816952b68efd4a1a738839d6c78/src/vcpkg/packagespec.cpp#L17-L21 and https://github.com/microsoft/vcpkg-tool/blob/7d292feb55656816952b68efd4a1a738839d6c78/include/vcpkg/packagespec.h#L222 which should have already taught fmt how to write these.

src/vcpkg/dependencies.cpp
```diff
@@ -829,17 +826,19 @@ namespace vcpkg
...
+
                             if (unsupported_port_action == UnsupportedPortAction::Error)
                             {
-                                Checks::exit_with_message(VCPKG_LINE_INFO, "Error: " + msg);
+                                Checks::msg_exit_with_message(VCPKG_LINE_INFO, localized_msg);
                             }
                             else
                             {
-                                m_warnings.push_back("Warning: " + msg);
+                                m_warnings.push_back("Warning: " + localized_msg.extract_data());
```
BillyONeal, line 841:
```suggestion              m_warnings.push_back(msg::format(msg::msgWarningMessage).append(std::move(localized_msg)).extract_data());
```

BillyONeal, line 841:
```suggestion
                                m_warnings.push_back(msg::format(msg::msgWarningMessage).append(std::move(localized_msg)).extract_data());
```

src/vcpkg/dependencies.cpp
```diff
@@ -1177,51 +1181,47 @@ namespace vcpkg
...
-                   actions_to_output_string(already_installed_plans),
-                   '\n');
+            msg::println(
+                msg::format(msgInstalledPackages).append_raw("\n" + actions_to_output_string(already_installed_plans)));
         }
 
         if (!remove_specs.empty())
         {
-            std::string msg = "The following packages will be removed:\n";
+            auto message = msg::format(msgPackagesToRemove).append_raw("\n");
```
BillyONeal, line 1195:
```suggestion
            auto message = msg::format(msgPackagesToRemove);
```

This `\n` is already added by either `append_raw` or `println`. Please check that the output here makes sense by building vcpkg and running with various install and remove commands to trigger the conditions, and compare with the 'before' state. Example:

image

(I think this change will insert an extra newline before the actual list...)

src/vcpkg/dependencies.cpp
```diff
@@ -1177,51 +1181,47 @@ namespace vcpkg
 
         if (!excluded.empty())
         {
-            print2("The following packages are excluded:\n", actions_to_output_string(excluded), '\n');
+            msg::println(msg::format(msgExcludedPackages).append_raw("\n" + actions_to_output_string(excluded)));
```
BillyONeal, line 1184:
```suggestion
            msg::println(msg::format(msgExcludedPackages).append_raw('\n').append_raw(actions_to_output_string(excluded)));
```

It should be a tiny bit more efficient to always do appends like this rather than forming a temporary concatenated string.

Yeah, it's a perf nitpick that will never matter but it's good habits :)

BillyONeal: Ditto many times below.

src/vcpkg/dependencies.cpp
```diff
@@ -1177,51 +1181,47 @@ namespace vcpkg
...
-            print2("The following packages will be directly installed:\n",
-                   actions_to_output_string(only_install_plans),
-                   '\n');
+            msg::println(msg::format(msgPackagesToInstallDirectly)
+                             .append_raw("\n" + actions_to_output_string(only_install_plans)));
         }
 
-        if (has_non_user_requested_packages)
-            print2("Additional packages (*) will be modified to complete this operation.\n");
+        if (has_non_user_requested_packages) msg::println(msgPackagesToModify);
```
BillyONeal, line 1219:
```suggestion
        if (has_non_user_requested_packages) { msg::println(msgPackagesToModify); }
```

BillyONeal (changes requested):

include/vcpkg/base/messages.h
```diff
@@ -1147,12 +1172,23 @@ namespace vcpkg
...
+    DECLARE_MESSAGE(OptionRequired, (msg::option), "", "--{option} option is required.");
     DECLARE_MESSAGE(OriginalBinParagraphHeader, (), "", "\nOriginal Binary Paragraph");
     DECLARE_MESSAGE(PackageFailedtWhileExtracting,
                     (msg::value, msg::path),
                     "'{value}' is either a tool name or a package name.",
                     "'{value}' failed while extracting {path}.");
     DECLARE_MESSAGE(PackageRootDir, (), "", "(Experimental) Specify the packages root directory.");
+    DECLARE_MESSAGE(PackagesToInstall, (), "", "The following packages will be built and installed:");
+    DECLARE_MESSAGE(PackagesToInstallDirectly, (), "", "The following packages will be directly installed:");
+    DECLARE_MESSAGE(PackagesToModify, (), "", "Additional packages (*) will be modified to comeplete this operation.");
```
vicroms, line 1184:
```suggestion
    DECLARE_MESSAGE(PackagesToModify, (), "", "Additional packages (*) will be modified to complete this operation.");
```

vicroms (changes requested):

ras0219-msft (commented):
~In general, start with messages in the (single) cpp file that uses them. Only move them to the common area when they're used at least twice.~ I see that this is now intended based on https://github.com/microsoft/vcpkg-tool/pull/615

vicroms (changes requested):

include/vcpkg/base/messages.h
```diff
@@ -1147,12 +1172,23 @@ namespace vcpkg
...
+    DECLARE_MESSAGE(PackagesToReBuild, (), "", "The following packages will be rebuilt:");
```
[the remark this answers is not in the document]
JavierMatosD: What change are you suggesting here?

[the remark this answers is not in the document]
vicroms: ReBuild -> Rebuild, not strongly against just leaving as-is.

BillyONeal:
Would you be willing to take the changes in https://github.com/BillyONeal/vcpkg-tool/tree/ursa-major-working ?

BillyONeal (approved):

BillyONeal:
Added those changes as Javier confirmed they were OK in person

vicroms (approved):

PR #688, opened 2022-09-01
Optimize `x-history` Using Parallel Algorithms

Thomas1664:
Vcpkg needs to run 2 git processes (one for `CONTROL`, the other one for `vcpkg.json`) for each entry in the version database. This results in over 100 executions of git for the `openssl` port

Because those processes don't depend on each other, it is possible to loop over every port version in parallel using `std::for_each`. This results in a ~1.5x performance improvement [Windows, Debug, command: `vcpkg x-history openssl`] (before: 27.4 s; after: 18.5 s).

### DIFF
--- a/src/vcpkg/commands.porthistory.cpp
+++ b/src/vcpkg/commands.porthistory.cpp
@@ -12,6 +12,9 @@
 #include <vcpkg/versiondeserializers.h>
 #include <vcpkg/versions.h>
 
+#include <algorithm>
+#include <execution>
+
 namespace vcpkg::Commands::PortHistory
 {
     namespace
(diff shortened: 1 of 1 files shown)


ras0219-msft:
Thanks for the PR!

Definitely agree this will improve the performance of `x-history`, however our intent is to remove this command entirely. The command was created before the existence of first-class versioning and no longer serves a purpose.

Do you use this command on a regular basis?

Thomas1664:
> Do you use this command on a regular basis?

No, I just stepped over it and saw that it was slow. I can open a PR to remove this command entirely if you wish.

PR #694, opened 2022-09-03
Use parallel for_each to load all ports

Thomas1664:
This PR adds the ability to use parallel algorithms to `CMakeLists.txt`. ~~GCC needs OpenMP to use this feature ([docs](https://gcc.gnu.org/onlinedocs/libstdc++/manual/parallel_mode_using.html)). Fortunately, OpenMP is preinstalled and we can easily find it with cmake.~~

I implemented the parallel version of `std::for_each` for UNIX by myself because Clang doesn't support parallel algorithms at all and GCC requires run-time dependencies.

### `x-add-version --all`

I was able to get 2x performance improvement [12 s before vs. 6 s after; GCC Debug]. I had similar results on Windows release and debug builds. I also tested if there are any downsides when using this command without `--all` and found out that `for_each` is a little bit faster than the range based for loop that was used previously.


include/vcpkg/base/parallel-algorithms.h
```diff
@@ -0,0 +1,25 @@
...
+#include <execution>
+#elif defined(GNU_USE_PARALLEL_ALG)
+#include <parallel/algorithm>
+#else
+#include <algorithm>
+#endif
+
+
+#if defined(USE_PARALLEL_ALG) && !defined(GNU_USE_PARALLEL_ALG)
+#define vcpkg_parallel_for_each std::for_each(std::execution::par,
```
Thomas1664, line 14: This is significantly faster than using a real function because it doesn't get inlined.

include/vcpkg/base/cache.h
```diff
@@ -49,4 +50,32 @@ namespace vcpkg
...
+                return it->second;
+            }
+
+            std::lock_guard lock(m_mtx);
+            return m_cache.emplace_hint(it, k, static_cast<F&&>(f)())->second;
+        }
+
+    private:
+        mutable std::map<Key, Value, Compare> m_cache;
+        mutable std::mutex m_mtx;
```
Thomas1664, line 79: I can't add a mutex to the normal cache because this apparently needs to be copied somewhere.

include/vcpkg/base/parallel-algorithms.h
```diff
@@ -0,0 +1,24 @@
...
+#elif defined(GNU_USE_PARALLEL_ALG)
+#include <parallel/algorithm>
+#else
+#include <algorithm>
+#endif
+
+#if defined(USE_PARALLEL_ALG) && !defined(GNU_USE_PARALLEL_ALG)
+#define vcpkg_parallel_for_each std::for_each(std::execution::par,
+#elif defined(GNU_USE_PARALLEL_ALG)
+#define vcpkg_parallel_for_each __gnu_parallel::for_each(
```
dg0yt, line 15: https://gcc.gnu.org/onlinedocs/libstdc++/manual/parallel_mode.html:
> For C++17 and above there are new overloads of the standard algorithms which take an execution policy argument. You should consider using those instead of the non-standard parallel mode extensions. 

Do I understand correctly that `__gnu_parallel::for_each` shouldn't be used? 

I dislike the macro stuff here. And the unbalanced `(` in definition and `)` when used are not acceptable IMO.

Thomas1664: > Do I understand correctly that `__gnu_parallel::for_each` shouldn't be used?

> For C++17 and above there are new overloads of the standard algorithms which take an execution policy argument. You should consider using those instead of the non-standard parallel mode extensions.

I considered using STL parallel algorithms but they need tbb as a dependence which is not available in CI and on people's devices. I already discussed this with BillyONeal on Discord: We can't depend on tbb. Therefore, I think I did the right think here since `__gnu_parallel` is not deprecated.

> I dislike the macro stuff here. And the unbalanced ( in definition and ) when used are not acceptable IMO.

As stated above, macros are the fastest solution here als regular functions can't be inlined. I definately agree  on the braces with you. I will work out a better solution here.

dg0yt: > As stated above, macros are the fastest solution here als regular functions can't be inlined.

The compiler might even inline a regular function if optimization is enabled. But we are talking about template functions here, not regular functions. Macros may be faster at compile time, but I don't think there is an advantage at run time.

Thomas1664: > The compiler might even inline a regular function if optimization is enabled. But we are talking about template functions here, not regular functions. Macros may be faster at compile time, but I don't think there is an advantage at run time.

I have already benchmarked that using the template function is slower than using `std:.for_each` directly with optimization turned on. To repeat: The templated function doesn't get inlined, which makes it slower. I already have benchmarked this. If you don't believe me, feel free to benchmark this on your own. If you are not able to provide a suggestion that is faster than my code, I won't change this.

IMO there is nothing wrong with using macros as functions.

include/vcpkg/base/cache.h
```diff
@@ -49,4 +50,32 @@ namespace vcpkg
...
+    struct AtomicCache
+    {
+        template<class KeyIsh,
+                 class F,
+                 std::enable_if_t<std::is_constructible<Key, const KeyIsh&>::value &&
+                                      detail::is_callable<Compare&, const Key&, const KeyIsh&>::value,
+                                  int> = 0>
+        const Value& get_lazy(const KeyIsh& k, F&& f) const
+        {
+            auto it = m_cache.lower_bound(k);
```
BillyONeal, line 64: This is a data race, since nothing stops `emplace_hint` from being called by another thread at the same time.

BillyONeal (changes requested):
In addition to the data race I made a comment about, I highly doubt that `update_version_db_file` and `update_baseline_version` are safe to call concurrently.

I am additionally concerned about adding additional runtime dependencies on Linux (OpenMP). We are a bootstrapping tool in a lot of situations, so we have to be careful there. If that dependency would ever be a separate apt package, for example, we probably do not want to do that.

That isn't saying that I dislike the parallel algorithms library, or that I dislike what this is doing. On the contrary, I think this concept in general is a good improvement, and certainly on Windows where it doesn't introduce any additional deployment problems, I'm all for it. (If anything I'm biased to say use more parallel algos as the original author of most of MSVC++'s parallel algorithms implementation :)) It's just that I think we need to value broad compatibility over performance improvements like this given the position that we are in. We ripped out the std::filesystem dependency for similar reasons.

Paths forward here if you want to keep contributing in a similar area:
* Parallel for_each isn't that complex an algorithm to do for this, particularly when we know the 'elements' are going to be relatively expensive / do disk I/O like this. You could build such a thing on top of pthreads which would eliminate the extra dependencies concern on other platforms.
* Our installation process currently could benefit a lot from being pipelined. For example, building the next thing doesn't have to wait for building binary cache packages and/or copying installed bits for the current thing.

Thanks for your contribution and I hope it works out!

ras0219-msft:
> Our installation process currently could benefit a lot from being pipelined. For example, building the next thing doesn't have to wait for building binary cache packages and/or copying installed bits for the current thing.

A quick note to make the implicit explicit: we _do_ have potential race conditions between multiple installs (manipulations within the downloads folder). The obvious improvement here is to start the next build while {packing up, uploading to binary caches, installing files}, but we currently _cannot_ start another build while the current build is in progress.

Thomas1664:
> A quick note to make the implicit explicit: we do have potential race conditions between multiple installs (manipulations within the downloads folder). The obvious improvement here is to start the next build while {packing up, uploading to binary caches, installing files}, but we currently cannot start another build while the current build is in progress.

We can only do binary caching in parallel because post build lint needs to happen before the next build starts. Parallel downloads are impossible because downloads are triggered from inside cmake and there might be file conflicts.

If we were to install ports in parallel we have to keep the console output consistent maybe by adding `[port-name]` in front of it.

In general, this is does not require parallel algorithms but is more a problem of `std::async`.

Thomas1664:
> * Parallel for_each isn't that complex an algorithm to do for this, particularly when we know the 'elements' are going to be relatively expensive / do disk I/O like this. You could build such a thing on top of pthreads which would eliminate the extra dependencies concern on other platforms.

I'd prefer `std::async` because it's C++ and doesn't require additional dependencies.

Thomas1664:
> Parallel for_each isn't that complex an algorithm to do for this, particularly when we know the 'elements' are going to be relatively expensive / do disk I/O like this. You could build such a thing on top of pthreads which would eliminate the extra dependencies concern on other platforms.

@BillyONeal I'm unsure how to determine the number of threads to use. I noticed that Windows uses way more threads than specified in `std::thread::hardware_concurrency()`. What do you consider a sufficient limit without slowing things down?

Another option to fix the compatibility problem might be to use tbb as a static library. Is there anything that prevents us from doing so?

BillyONeal:
> I'm unsure how to determine the number of threads to use. I noticed that Windows uses way more threads than specified in `std::thread::hardware_concurrency()`.

The Windows implementation calls the system thread pool and it decides how many to use based on system wide operating conditions. Baring that, `std::thread::hardware_concurrency()` seems a reasonable starting point.

We aren't trying to replace a production OpenMP implementation here.

include/vcpkg/base/parallel-algorithms.h
```diff
@@ -0,0 +1,69 @@
...
+                {
+                    cb(*(begin + i + j));
+                }
+            }));
+        }
+
+        for (It start = begin + other_threads * quot; start != end; ++start)
+        {
+            cb(*start);
+        }
```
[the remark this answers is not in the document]
Thomas1664: This is way more complicated than it needs to be...

src/vcpkg/commands.add-version.cpp
```diff
@@ -393,15 +394,16 @@ namespace vcpkg::Commands::AddVersion
             msg::println_warning(msgAddVersionDetectLocalChangesError);
         }
 
-        for (auto&& port_name : port_names)
-        {
+        std::mutex mtx;
+
+        auto work = [&](const std::string& port_name) {
```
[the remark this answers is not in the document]
Thomas1664: @ras0219-msft The thing is that `try_load_port()` itself may call `Checks::check_exit()`. Given the fact that we sprinkle `Checks::check_exit` everywhere in our code, there is nearly nothing that we can parallelize!

IMO we should replace this by throwing an exception instead and print the error message properly by wrapping everything in `main()` inside a `try ... catch` and call `Checks::check_exit` from `catch`.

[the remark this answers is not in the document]
ras0219-msft: Then as part of this, `try_load_port()` should be fixed to never call `Checks::check_exit()` and always return its errors via the existing return type of `Expected<>`.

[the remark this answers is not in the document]
Thomas1664: @ras0219-msft The formatting check can't be parallelized because it calls too many functions that may exit. Please also have a look at https://github.com/microsoft/vcpkg/issues/27152 .

Thomas1664:
Hi @BillyONeal, @ras0219-msft,

As mentioned above, we can't use parallel algorithms if the functions that are called from different threads may exit. Unfortunately, exiting from a function is widespread across the codebase and even happens in base things like `Filesystem` or `Json`. Even if I replaced all of those exits with `ExpectedS`, there might be future changes down in the call stack that add calls to exit. Therefore, I don't think it's safe to use any kind of parallelization until we have a policy on when it is ok to exit the program.

BillyONeal:
It seems unlikely to me that we are going to invest in this direction anytime soon.

PR #695, opened 2022-09-04
Work around popen not being thread safe on MacOS

Thomas1664:
Related https://github.com/microsoft/vcpkg/issues/26651

~~vcpkg occasionally crashed at `decompress_in_parallel` > `cmd_execute_and_capture_output_parallel` because threads that are launched by `cmd_execute_and_capture_output_parallel` were writing to the result vector without locking the vector.~~

Not a problem because they access different indexes of the vector and don't cause a resize.

@omartijn Thanks for providing the debug information.

### `max_concurrency()`

`get_concurrency()` returned `int` which didn't make any sense because there can't be `-1` threads.


src/vcpkg/archives.cpp
```diff
@@ -333,7 +333,7 @@ namespace vcpkg
         auto results =
             cmd_execute_and_capture_output_parallel(jobs, default_working_directory, get_clean_environment());
 #ifdef __APPLE__
-        int i = 0;
+        size_t i = 0;
```
Thomas1664, line 336: Integer underflow when `results` is lager than `INT_MAX`

src/vcpkg/base/system.process.cpp
```diff
@@ -476,18 +476,24 @@ namespace vcpkg
         }
 
         std::atomic<size_t> work_item{0};
+        std::mutex mtx;
         const auto num_threads =
             static_cast<size_t>(std::max(1, std::min(get_concurrency(), static_cast<int>(cmd_lines.size()))));
 
+        vcpkg::Checks::check_exit(VCPKG_LINE_INFO, num_threads > 1, "num_threads is %d", num_threads);
```
Thomas1664, line 483: This is weird: `num_threads` should be an `unsigned int`. Negative number of threads doesn't make any sense. Interestingly, `get_concurrency()` also returns `int` instead which doesn't make any sense either. If `-1` actually is a valid state, I would return an `Optional<unsigned int>` instead.

src/vcpkg/base/system.process.cpp
```diff
@@ -476,18 +476,24 @@ namespace vcpkg
...
+
         auto work = [&]() {
             std::size_t item;
             while (item = work_item.fetch_add(1), item < cmd_lines.size())
             {
-                res[item] = cmd_execute_and_capture_output(cmd_lines[item], wd, env);
+                auto cmd_output = cmd_execute_and_capture_output(cmd_lines[item], wd, env);
+                std::lock_guard<std::mutex> guard(mtx);
+                res[item] = std::move(cmd_output);
```
Thomas1664, line 491: Lock when writing to shared memory

src/vcpkg/commands.dependinfo.cpp
```diff
@@ -120,7 +120,7 @@ namespace vcpkg::Commands::DependInfo
                 std::string value = iter->second;
                 try
                 {
-                    return std::stoi(value);
+                    return std::max(std::stoi(value), NO_RECURSE_LIMIT_VALUE);
```
Thomas1664, line 123: Who says a user puts only numbers >= -1 in here?

src/vcpkg/base/system.cpp
```diff
@@ -417,17 +417,26 @@ namespace vcpkg
...
 
-    int get_concurrency()
+    unsigned int get_concurrency()
     {
-        static int concurrency = [] {
+        static unsigned int concurrency = [] {
             auto user_defined_concurrency = get_environment_variable("VCPKG_MAX_CONCURRENCY");
             if (user_defined_concurrency)
             {
-                return std::stoi(user_defined_concurrency.value_or_exit(VCPKG_LINE_INFO));
```
Thomas1664, line 426: Crash if `VCPKG_MAX_CONCURRENCY` was NaN

src/vcpkg/base/system.process.cpp
```diff
@@ -476,18 +476,21 @@ namespace vcpkg
...
-        const auto num_threads =
-            static_cast<size_t>(std::max(1, std::min(get_concurrency(), static_cast<int>(cmd_lines.size()))));
+        std::mutex mtx;
+        const auto num_threads = std::max(static_cast<size_t>(1), std::min(static_cast<size_t>(get_concurrency()), cmd_lines.size()));
 
         auto work = [&]() {
             std::size_t item;
             while (item = work_item.fetch_add(1), item < cmd_lines.size())
             {
-                res[item] = cmd_execute_and_capture_output(cmd_lines[item], wd, env);
```
Thomas1664, line 486: Crash because of race condition

src/vcpkg/base/system.cpp
```diff
@@ -417,17 +417,26 @@ namespace vcpkg
...
+                try
+                {
+                    int res = std::stoi(user_defined_concurrency.value_or_exit(VCPKG_LINE_INFO));
+                    vcpkg::Checks::check_exit(VCPKG_LINE_INFO, res > 0, "VCPKG_MAX_CONCURRENCY is %d, must be > 0", res);
+                    return static_cast<unsigned int>(res);
+                }
+                catch (std::exception&)
+                {
+                    Checks::msg_exit_with_message(VCPKG_LINE_INFO, msgOptionMustBeInteger, msg::option = "VCPKG_MAX_CONCURRENCY");
```
Thomas1664, line 434: @BillyONeal Should we throw if `VCPKG_MAX_CONCURRENCY` is invalid or should we simply continue with `std::thread::hardware_concurrency()`

src/vcpkg/base/system.process.cpp
omartijn: I'm really wondering if this is it, though. Maybe I'm wrong, but I've always been under the impression that updating _distinct keys_ from an `std::vector` from multiple threads should be safe, provided that you're not resizing the `std::vector`.

Once it's allocated, it's just a wrapper around a chunk of memory, as long as writes touch different memory addresses I'd imagine that should be safe to happen simultaneously.

Thomas1664: > Once it's allocated, it's just a wrapper around a chunk of memory, as long as writes touch different memory addresses I'd imagine that should be safe to happen simultaneously.

Technically, I would agree with you. However, I don't think that the standard specifies that there shall be no resizing going on between vector allocation and between insertions at defined indexes.

The other possibility is that multiple threads try to write to the same index but this is permitted by the atomic add.

I think this is a use case for parallel `for_each` but unfortunately, Clang doesn't support this.

Thomas1664: Additionally, I don't think that the lock will hurt performance that much because `operator[]` should be fast in comparison to executing the command.

Thomas1664: > I'm really wondering if this is it, though. Maybe I'm wrong, but I've always been under the impression that updating _distinct keys_ from an `std::vector` from multiple threads should be safe, provided that you're not resizing the `std::vector`.
> 
> Once it's allocated, it's just a wrapper around a chunk of memory, as long as writes touch different memory addresses I'd imagine that should be safe to happen simultaneously.

@omartijn I think you are actually right there. I did some tests and it seems like concurrent writes to different indexes of the vector aren't a problem at all here and I can safely remove the lock. I did some research on the error code `EXC_BAD_ACCESS` and `KERN_INVALID_ADDRESS` and found issues that point to bugs in Xcode. Since the crash occurs in `popen` I think that this is not an issue with vcpkg but related to Xcode. The correct thing would be to compile vcpkg with a newer version of Xcode and hope that the bug is fixed. I will also add unit tests to make sure we don't mess up concurrency here.

omartijn: If it is indeed an xcode bug, that would explain why I was unable to reproduce it with a local debug build. I'm interested which issues you found that it's an issue in xcode. Could you share links?

The fact that it crashes inside `popen` itself is not an indication that it's a bug in xcode. In fact, `popen` is part of the system library and is not updated by an xcode update. Usually a `popen` crash means that one of the arguments is garbage (i.e. `nullptr` or already freed, not correctly null-terminated).

Thomas1664:
@omartijn Gotcha!

I created a unit test that executes the command `echo a` where a is a string of the char a that matches the current index, i.e. `` for index 0, `a` for index 1, `aa` for index 2 and so on.

Here are the results:

```
...
Index: 36, length: 36
Index: 37, length: 37
Index: 38, length: 18446744073709551615
```

The problem is that in random positions the string returned from the child process is missing a '\0' which leads to a string length of `SIZE_T_MAX`. I subtracted 1 in my test to align with the current index.

The segfault comes into place when someone tries to read at index > 38 in this case.

src/vcpkg/base/system.process.cpp
```diff
@@ -895,18 +896,21 @@ namespace vcpkg
...
         {
             return format_system_error_message("popen", errno);
         }
 
-        char buf[1024];
+        char buf[1024] = {0};
         // Use fgets because fread will block until the entire buffer is filled.
-        while (fgets(buf, 1024, pipe))
+        // 1 less than buffer size because OSX writes \0 1 byte after the given char count
+        while (fgets(buf, 1023, pipe))
```
Thomas1664, line 913: Although this is apparently not causing the bug, this was still broken:

On OSX `fgets` takes the maximum count of characters to read, inserting `\0` after the last character:

>Security Note for fgets: Although the [fgets](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/fgets.3.html#//apple_ref/doc/man/3/fgets) function provides the ability to read a limited amount of data, you must be careful when using it. Like the other functions in the “safer” column, fgets always terminates the string. However, unlike the other functions in that column, it takes a maximum number of bytes to read, not a buffer size.

In practical terms, this means that you must always pass a size value that is one fewer than the size of the buffer to leave room for the null termination. If you do not, the fgets function will dutifully terminate the string past the end of your buffer, potentially overwriting whatever byte of data follows it.

(Source: https://developer.apple.com/library/archive/documentation/Security/Conceptual/SecureCodingGuide/Articles/BufferOverflows.html)

In contrast, Linux takes in the size of the buffer and reads 1 less than size:

> fgets() reads in at most one less than size characters from stream and stores them into the buffer pointed to by s. Reading stops after an EOF or a newline. If a newline is read, it is stored into the buffer. A terminating null byte (aq\0aq) is stored after the last character in the buffer.

Source: https://linux.die.net/man/3/fgets

This caused a buffer overflow on OSX.
This probably never happened in reality because the output of all commands that vcpkg is running should be less than 1024 characters per line.

Thomas1664:
My findings so far:
sometimes the input command is missing something:

`[DEBUG] command line:  echo  2>&1`

This results in an error:

```
/Users/runner/work/1/s/src/vcpkg-test/system.cpp:180: FAILED:
  REQUIRE( out->exit_code == 0 )
with expansion:
  127 == 0
```

There is no console output because error messages go to `stderr` which is not captured by `popen`. Therefore, the string containing the console output remains uninitialized:

`Index: 5, length: 18446744073709551615, content: `

omartijn:
For me, the sanitizer errors go away when I make the following change:

```
diff --git a/src/vcpkg/binarycaching.cpp b/src/vcpkg/binarycaching.cpp
index 5161c79cb..7596df357 100644
--- a/src/vcpkg/binarycaching.cpp
+++ b/src/vcpkg/binarycaching.cpp
@@ -2134,8 +2134,10 @@ ExpectedS<DownloadManagerConfig> vcpkg::parse_download_configuration(const Optio
 
     LockGuardPtr<Metrics>(g_metrics)->track_property("asset-source", "defined");
 
+    std::string asset_sources{Strings::concat("$", VcpkgCmdArguments::ASSET_SOURCES_ENV)};
+
     AssetSourcesState s;
-    AssetSourcesParser parser(*arg.get(), Strings::concat("$", VcpkgCmdArguments::ASSET_SOURCES_ENV), &s);
+    AssetSourcesParser parser(*arg.get(), asset_sources, &s);
     parser.parse();
     if (auto err = parser.get_error())
     {

```

Thomas1664:
> For me, the sanitizer errors go away when I make the following change:

This might be the case but does not explain why my test case fails. I was able to show that although I get error code 127 the inputs are always valid. The setup seems also correct: We flush `stdout` before `popen`. The only remaining possibilities are that either that debug messages are displayed between flush and popen which seems to be possible after looking at the logs or that popen has an internal error

src/vcpkg/base/system.process.cpp
```diff
@@ -895,18 +896,21 @@ namespace vcpkg
...
 
-        const auto pipe = popen(actual_cmd_line.c_str(), "r");
+        const auto pipe = popen(cmd_str, "r");
         if (pipe == nullptr)
         {
             return format_system_error_message("popen", errno);
         }
 
-        char buf[1024];
+        char buf[1024] = {0};
```
BillyONeal, line 910:
```suggestion
        char buf[1024];
```

I don't think zeroing this is useful given that we immediately overwrite it with `fgets`.

src/vcpkg-test/system.cpp
```diff
@@ -5,9 +5,11 @@
 #include <vcpkg/base/optional.h>
 #include <vcpkg/base/strings.h>
 #include <vcpkg/base/stringview.h>
+#include <vcpkg/base/system.debug.h>
 #include <vcpkg/base/system.h>
 #include <vcpkg/base/system.process.h>
 
+#include <iostream>
```
BillyONeal, line 12: Do we use iostream for anything else? I thought we burninated all use of it.

BillyONeal (approved):

src/vcpkg-test/system.cpp
Thomas1664: No, I used this for debugging

src/vcpkg/base/system.process.cpp
Thomas1664: > Huh? I don't see that _at all_. If I look in the manpage, or look it up online, I find the following at the link you provided:

This is the correct link: https://developer.apple.com/library/archive/documentation/Security/Conceptual/SecureCodingGuide/Articles/BufferOverflows.html

I recommend searching for `fgets` to find the paragraph about `fgets`. Both pages are contradicting. It's not clear to me which one is correct although I would assume the `fgets` man page to be correct.

Thomas1664:
@BillyONeal are you sure you want to approve this one?

Unit tests are still failing. I'd have to remove the new test case to make CI succeed. Are you sure that my changes in `system.debug.h` are really necessary?

BillyONeal:
Approval means 'I don't see any major flags here and would be OK with merging it'. Of course, tests need to be made working before the merge can actually happen.

src/vcpkg/base/system.process.cpp
```diff
@@ -501,6 +502,23 @@ namespace vcpkg
         {
             w.get();
         }
+
+#if defined(__APPLE__)
```
Thomas1664, line 506: Note that @ras0219-msft Didn't want to have this inside `cmd_execute_and_stream_data`: https://github.com/microsoft/vcpkg-tool/pull/306#issuecomment-1007031965

Because this is an issue for every command that is executed in parallel, I think it is more reasonable to put it here than just putting it in `decompress_in_parallel`

ras0219-msft: Why is this an issue for every command executed in parallel? I am strongly against this change without a thorough understanding of why it happens and why it's safe to rerun any arbitrary process with a 127 exit code.

ras0219-msft (changes requested):

src/vcpkg/base/system.process.cpp
Thomas1664: > Why is this an issue for every command executed in parallel?

I really don't know. I guess this is something you have to ask the folks at Apple since this is most likely not our fault. I read the documentation for every POSIX function that is involved and couldn't find anything. Therefore, I think this is an internal bug of `popen` / `pclose`.

> without a thorough understanding of why it happens

The issue seems to occur only when many processes are launched at the same time with `popen` but this is just guessing. You better ask the Apple folks as we can't step into this.

> why it's safe to rerun any arbitrary process with a 127 exit code.

We're not running every process again that exits with 127 - only if there is no error message (console output). I'm pretty sure there is an error message if you did `eho` instead of `echo` and thus was a real 127. 127 may also be returned if the system is unable to execute the command although it exists: https://www.ibm.com/support/pages/exit-code-127-means-jobs-command-can-not-be-found-or-executed

What else do you expect to happen when you rerun a program that returned 127?

ras0219-msft: Doing further research, I think it's likely due to opening and closing files internally as part of `popen()`. The full process is perhaps not thread safe because it relies on file descriptor inheritance / numbering.

Instead of continuing to try to patch over race conditions, I think we should fix this problem correctly in one of two ways:
1. (cheap and easy) Add a mutex around the `popen()` call on MacOS.
2. (more involved) Switch to using `posix_spawn()` for MacOS.

Thomas1664: I will go for a static mutex inside the function that uses `popen` instead of passing it around as a pointer because this seems to be the least messy way.

Thomas1664:
Ping @ras0219-msft for review

src/vcpkg/base/system.cpp
```diff
@@ -417,17 +417,28 @@ namespace vcpkg
...
+        static unsigned int concurrency = [] {
             auto user_defined_concurrency = get_environment_variable("VCPKG_MAX_CONCURRENCY");
             if (user_defined_concurrency)
             {
-                return std::stoi(user_defined_concurrency.value_or_exit(VCPKG_LINE_INFO));
+                try
+                {
+                    int res = std::stoi(user_defined_concurrency.value_or_exit(VCPKG_LINE_INFO));
+                    vcpkg::Checks::check_exit(
+                        VCPKG_LINE_INFO, res > 0, "VCPKG_MAX_CONCURRENCY is %d, must be > 0", res);
```
ras0219-msft, line 430: This is in response to user input and so must be localized.

src/vcpkg/base/system.process.cpp
ras0219-msft: Could you add a comment with that link?

src/vcpkg-test/system.cpp
```diff
@@ -142,3 +142,42 @@ TEST_CASE ("cmdlinebuilder", "[system]")
...
+        REQUIRE(out->exit_code == 0);
+
+#if defined(_WIN32)
+        REQUIRE(out->output == (std::to_string(i) + "\r\n"));
+#else
+        if (i != 0)
+        {
+            REQUIRE(out->output[0] == 'a');
+            REQUIRE(out->output[i - 1] == 'a');
+        }
```
ras0219-msft, line 179: I don't understand what this is checking separately from the next line.
```suggestion
```

ras0219-msft (changes requested):

src/vcpkg/base/system.process.cpp
autoantwort: address sanitizer also found this:
```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==7417==ERROR: AddressSanitizer: BUS on unknown address (pc 0x00018ad7c2bc bp 0x00016d02d0d0 sp 0x00016d02d050 T407)
==7417==The signal is caused by a UNKNOWN memory access.
==7417==Hint: this fault was caused by a dereference of a high value address (see register values below).  Disassemble the provided pc to learn which register was used.
AddressSanitizer:DEADLYSIGNAL
    #0 0x18ad7c2bc in popen+0x230 (libsystem_c.dylib:arm64e+0x522bc)
    #1 0x107dcfe4c in wrap_popen+0x2ac (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x37e4c)
    #2 0x103308a94 in vcpkg::cmd_execute_and_stream_data(vcpkg::Command const&, std::__1::function<void (vcpkg::StringView)>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&, vcpkg::Encoding) system.process.cpp:927
    #3 0x1032fffe8 in vcpkg::cmd_execute_and_capture_output(vcpkg::Command const&, vcpkg::WorkingDirectory const&, vcpkg::Environment const&, vcpkg::Encoding, vcpkg::EchoInDebug) system.process.cpp:984
    #4 0x103301c44 in vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2::operator()() const system.process.cpp:489
    #5 0x103327f18 in decltype(static_cast<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2>(fp)()) std::__1::__invoke<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2>(vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2&&) type_traits:3918
    #6 0x103327ebc in void std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2>::__execute<>(std::__1::__tuple_indices<>) future:2190
    #7 0x103327db4 in std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2>::operator()() future:2183
    #8 0x1033274e8 in std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >::__execute() future:1001
    #9 0x10332c678 in decltype(*(static_cast<std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >*>(fp0)).*fp()) std::__1::__invoke<void (std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >::*)(), std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >*, void>(void (std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >::*&&)(), std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >*&&) type_traits:3859
    #10 0x10332c3f0 in void std::__1::__thread_execute<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void (std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >::*)(), std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >*, 2ul>(std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void (std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >::*)(), std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >*>&, std::__1::__tuple_indices<2ul>) thread:287
    #11 0x103329908 in void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void (std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >::*)(), std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >*> >(void*) thread:298
    #12 0x18ae6a068 in _pthread_start+0x90 (libsystem_pthread.dylib:arm64e+0x7068)
    #13 0x18ae64e28 in thread_start+0x4 (libsystem_pthread.dylib:arm64e+0x1e28)

==7417==Register values:
 x[0] = 0x0000000000000000   x[1] = 0x000000010fff00e0   x[2] = 0x000000016d02c7b8   x[3] = 0x0000000000000021  
 x[4] = 0x0000000000002088   x[5] = 0x0000000000000001   x[6] = 0x000000016cfac000   x[7] = 0x0000000000000001  
 x[8] = 0x00000001389d9280   x[9] = 0x0000000000000001  x[10] = 0x0000000000000410  x[11] = 0x0000000114ac0180  
x[12] = 0x47ae41c86816a996  x[13] = 0x00000000ffffffff  x[14] = 0x0000000000007e01  x[15] = 0x0000000000000006  
x[16] = 0x000000018ae64e9c  x[17] = 0x0000000107e24720  x[18] = 0x0000000000000000  x[19] = 0x00000001e6536dd0  
x[20] = 0x0000000105577cc0  x[21] = 0x000000010ec23a40  x[22] = 0x00000001105152e0  x[23] = 0x000000000000000f  
x[24] = 0x00000001e4025000  x[25] = 0x0000019d0000387f  x[26] = 0x0000000000000000  x[27] = 0x0000000000000000  
x[28] = 0x0000000000000000     fp = 0x000000016d02d0d0     lr = 0x000000018ad7c2c8     sp = 0x000000016d02d050  
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: BUS (libsystem_c.dylib:arm64e+0x522bc) in popen+0x230
Thread T407 created by T0 here:
    #0 0x107dd0c5c in wrap_pthread_create+0x54 (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x38c5c)
    #1 0x1033296cc in std::__1::__libcpp_thread_create(_opaque_pthread_t**, void* (*)(void*), void*) __threading_support:421
    #2 0x103328dd0 in std::__1::thread::thread<void (std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >::*)(), std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >*, void>(void (std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >::*&&)(), std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >*&&) thread:314
    #3 0x103326b48 in std::__1::thread::thread<void (std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >::*)(), std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >*, void>(void (std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >::*&&)(), std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >*&&) thread:306
    #4 0x103326090 in std::__1::future<void> std::__1::__make_async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >(std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2>&&) future:2159
    #5 0x10330144c in std::__1::future<std::__1::__invoke_of<std::__1::decay<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2&>::type>::type> std::__1::async<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2&>(std::__1::launch, vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2&) future:2210
    #6 0x1032ff664 in vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&) system.process.cpp:496
    #7 0x103377618 in vcpkg::decompress_in_parallel(vcpkg::Span<vcpkg::Command const>) archives.cpp:334
    #8 0x103468f8c in (anonymous namespace)::ArchivesBinaryProvider::try_restore_n(vcpkg::Span<vcpkg::InstallPlanAction const* const>, vcpkg::Path const&) const binarycaching.cpp:258
    #9 0x103465d00 in (anonymous namespace)::ArchivesBinaryProvider::prefetch(vcpkg::Span<vcpkg::InstallPlanAction const>, vcpkg::Span<vcpkg::CacheStatus* const>) const binarycaching.cpp:215
    #10 0x1033977cc in vcpkg::BinaryCache::prefetch(vcpkg::Span<vcpkg::InstallPlanAction const>) binarycaching.cpp:1421
    #11 0x104202300 in vcpkg::Install::perform(vcpkg::VcpkgCmdArguments const&, vcpkg::ActionPlan&, vcpkg::KeepGoing, vcpkg::VcpkgPaths const&, vcpkg::StatusParagraphs&, vcpkg::BinaryCache&, vcpkg::IBuildLogsRecorder const&, vcpkg::CMakeVars::CMakeVarProvider const&) install.cpp:536
    #12 0x1039399a0 in vcpkg::Commands::CI::perform_and_exit(vcpkg::VcpkgCmdArguments const&, vcpkg::VcpkgPaths const&, vcpkg::Triplet, vcpkg::Triplet) commands.ci.cpp:591
    #13 0x103950f00 in vcpkg::Commands::CI::CICommand::perform_and_exit(vcpkg::VcpkgCmdArguments const&, vcpkg::VcpkgPaths const&, vcpkg::Triplet, vcpkg::Triplet) const commands.ci.cpp:670
    #14 0x102eed49c in (anonymous namespace)::inner(vcpkg::Filesystem&, vcpkg::VcpkgCmdArguments const&) vcpkg.cpp:143
    #15 0x102ee9fa0 in main vcpkg.cpp:350
    #16 0x18ab3fe4c  (<unknown module>)

==7417==ABORTING
[1]    7417 abort      vcpkg ci --x-abi-tools-use-exact-versions
```

src/vcpkg/base/system.process.cpp
autoantwort: According to https://opensource.apple.com/source/Libc/Libc-1439.141.1/stdio/FreeBSD/fgets.c.auto.html
The function reads n - 1 characters.

autoantwort:
@Thomas1664 Can the change for the mutex around popen be extracted to another PR? Its very annoying and I don't want to wait longer :)

src/vcpkg/base/system.cpp
```diff
@@ -417,17 +417,33 @@ namespace vcpkg
...
+                                                      msg::env_var = "VCPKG_MAX_CONCURRENCY",
+                                                      msg::value = res);
+                    }
+                    return static_cast<unsigned int>(res);
+                }
+                catch (std::exception&)
+                {
+                    Checks::msg_exit_with_message(
+                        VCPKG_LINE_INFO, msgOptionMustBeInteger, msg::option = "VCPKG_MAX_CONCURRENCY");
+                }
```
ras0219-msft, line 493: We should minimize the coverage of the `try` to ensure we don't capture any unexpected exceptions.

```suggestion
                int res;
                try
                {
                    res = std::stoi(user_defined_concurrency.value_or_exit(VCPKG_LINE_INFO));
                }
                catch (std::exception&)
                {
                    Checks::msg_exit_with_message(
                        VCPKG_LINE_INFO, msgOptionMustBeInteger, msg::option = "VCPKG_MAX_CONCURRENCY");
                }
                if (!(res > 0))
                {
                    Checks::msg_exit_with_message(VCPKG_LINE_INFO,
                                                  msgEnvInvalidMaxConcurrency,
                                                  msg::env_var = "VCPKG_MAX_CONCURRENCY",
                                                  msg::value = res);
                }
                return static_cast<unsigned int>(res);
            }
```

Thomas1664:
> @Thomas1664 Can the change for the mutex around popen be extracted to another PR? Its very annoying when vcpkg crashes and I don't want to wait longer :)

I think the issue you found with asan already fixes the crash?

autoantwort:
> > @Thomas1664 Can the change for the mutex around popen be extracted to another PR? Its very annoying when vcpkg crashes and I don't want to wait longer :)
> 
> I think the issue you found with asan already fixes the crash that I tried to fix with this PR?

I fixed another crash, but there are multiple ones: https://github.com/microsoft/vcpkg-tool/pull/695#discussion_r1018006232

Thomas1664:
> According to https://opensource.apple.com/source/Libc/Libc-1439.141.1/stdio/FreeBSD/fgets.c.auto.html
The function reads n - 1 characters.

@autoantwort This is apparently not true; Apple's documentation seems to be wrong. See https://dev.azure.com/vcpkg/public/_build/results?buildId=81166&view=logs&jobId=cf796865-97b7-5cd1-be8e-6e00ce4fd8cf&j=cf796865-97b7-5cd1-be8e-6e00ce4fd8cf&t=6d56bf67-9f63-54b3-7bd2-1eeca5f88481

It crashed when I changed it back to 1024 bytes.

autoantwort:
I only see this:
image

Thomas1664:
Never mind, the crash isn't fixed

Thomas1664:
@autoantwort If unit tests succeed on OSX in CI and the message `Mutex enabled` is printed to the console, could you please run the unit tests from this branch on your Mac (IIRC you do have one) 100-1000 times and if it crashes try setting the buffer to 1023 and try again? Seems like CI can't reliably test this.

autoantwort:
> **EDIT**: Never mind, I noticed that the debug message I added wasn't printed.

Because you have to pass `--debug` but you can't so that for the vcpkg-test executable

autoantwort:
Hm I still get 
```
Mutex enabled

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
vcpkg-test is a Catch v2.13.10 host application.
Run with -? for options

-------------------------------------------------------------------------------
cmd_execute_and_capture_output_parallel
-------------------------------------------------------------------------------
/Users/leanderSchulten/git_projekte/vcpkg-tool/src/vcpkg-test/system.cpp:146
...............................................................................

/Users/leanderSchulten/git_projekte/vcpkg-tool/src/vcpkg-test/system.cpp:146: FAILED:
due to a fatal error condition:
  SIGSEGV - Segmentation violation signal

Mutex enabled
===============================================================================
test cases:  348 |  347 passed | 1 failed
assertions: 5632 | 5631 passed | 1 failed

AddressSanitizer:DEADLYSIGNAL
=================================================================
==96972==ERROR: AddressSanitizer: SEGV on unknown address 0x00010a627c71 (pc 0x00019d3382bc bp 0x00016b951910 sp 0x00016b951890 T6)
==96972==The signal is caused by a UNKNOWN memory access.
Mutex enabled
Mutex enabled
Mutex enabled
    #0 0x19d3382bc in popen+0x230 (libsystem_c.dylib:arm64e+0x522bc)
    #1 0x1074b7e4c in wrap_popen+0x2ac (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x37e4c)
    #2 0x104fbc868 in vcpkg::cmd_execute_and_stream_data(vcpkg::Command const&, std::__1::function<void (vcpkg::StringView)>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&, vcpkg::Encoding) system.process.cpp:945
    #3 0x104fb86dc in vcpkg::cmd_execute_and_capture_output(vcpkg::Command const&, vcpkg::WorkingDirectory const&, vcpkg::Environment const&, vcpkg::Encoding, vcpkg::EchoInDebug) system.process.cpp:1004
    #4 0x104fb9434 in vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2::operator()() const system.process.cpp:491
    #5 0x104fc6afc in std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >::__execute() future:1001
    #6 0x104fc703c in void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void (std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >::*)(), std::__1::__async_assoc_state<void, std::__1::__async_func<vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&)::$_2> >*> >(void*) thread:298
    #7 0x19d426068 in _pthread_start+0x90 (libsystem_pthread.dylib:arm64e+0x7068)
    #8 0x19d420e28 in thread_start+0x4 (libsystem_pthread.dylib:arm64e+0x1e28)

==96972==Register values:
 x[0] = 0x0000000000000000   x[1] = 0x000000010c0700e0   x[2] = 0x000000016b950ff8   x[3] = 0x000000000000000f  
 x[4] = 0x0000000000002088   x[5] = 0x0000000000000001   x[6] = 0x000000016b8d0000   x[7] = 0x0000000000000001  
 x[8] = 0x000000010c607280   x[9] = 0x0000000000000001  x[10] = 0x0000000000000410  x[11] = 0x000000010c518980  
x[12] = 0x3afe852f934fa070  x[13] = 0x00000000ffffffff  x[14] = 0x0000000000007e01  x[15] = 0x0000000000000006  
x[16] = 0x000000019d420e9c  x[17] = 0x000000010750c720  x[18] = 0x0000000000000000  x[19] = 0x00000001f8af30c8  
x[20] = 0x0000000105e35c20  x[21] = 0x00000001095898d0  x[22] = 0x00000001093d1320  x[23] = 0x000000000000000d  
x[24] = 0x00000001f65e1000  x[25] = 0x000000000000bd52  x[26] = 0x000000016b952bc0  x[27] = 0x000000016b952180  
x[28] = 0x000000016b9522c0     fp = 0x000000016b951910     lr = 0x000000019d3382c8     sp = 0x000000016b951890  
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (libsystem_c.dylib:arm64e+0x522bc) in popen+0x230
Thread T6 created by T0 here:
    #0 0x1074b8c5c in wrap_pthread_create+0x54 (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x38c5c)
    #1 0x104fb75f4 in vcpkg::cmd_execute_and_capture_output_parallel(vcpkg::Span<vcpkg::Command const>, vcpkg::WorkingDirectory const&, vcpkg::Environment const&) system.process.cpp:499
    #2 0x104e103ac in C_A_T_C_H_T_E_S_T_8() system.cpp:164
    #3 0x1048b8178 in Catch::RunContext::invokeActiveTestCase() catch.hpp:13025
    #4 0x1048ae434 in Catch::RunContext::runCurrentTest(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&) catch.hpp:12998
    #5 0x1048ab944 in Catch::RunContext::runTest(Catch::TestCase const&) catch.hpp:12759
    #6 0x1048c2ebc in Catch::Session::runInternal() catch.hpp:13562
    #7 0x1048c0e60 in Catch::Session::run() catch.hpp:13518
    #8 0x10491bf94 in main catch.cpp:21
    #9 0x19d0fbe4c  (<unknown module>)

==96972==ABORTING
```

autoantwort:
Seems that it fails with std::async because that also calls `popen`

Thomas1664:
> Seems that it fails with std::async because that also calls `popen` (or `pthread_create`)

@autoantwort Although it shouldn't matter but did you also try to set the buffer back to 1023 at the line you commented about https://github.com/microsoft/vcpkg-tool/pull/695#discussion_r1018021986 ?

autoantwort:
Finally found it. The pclose call also has to be protected by the mutex.

autoantwort:
All the errors I get happens before the fgets line. This has nothing to do with the buffer size. All got all errors while the buffer size was 1023.

autoantwort:
Apply this patch to fix the pipeline:
```patch
diff --git a/src/vcpkg/base/system.process.cpp b/src/vcpkg/base/system.process.cpp
index c6eb7820..c00713c8 100644
--- a/src/vcpkg/base/system.process.cpp
+++ b/src/vcpkg/base/system.process.cpp
@@ -960,7 +960,15 @@ namespace vcpkg
             return format_system_error_message("feof", errno);
         }
 
-        int ec = pclose(pipe);
+        int ec;
+        {
+#if defined(__APPLE__)
+            // `pclose` sometimes returns 127 on OSX when executed in parallel.
+            // Related: https://github.com/microsoft/vcpkg-tool/pull/695#discussion_r973364608
+            std::lock_guard guard(mtx);
+#endif
+            ec = pclose(pipe);
+        }
         if (WIFEXITED(ec))
         {
             ec = WEXITSTATUS(ec);
```

autoantwort (approved):
I approve the macOS specific changes

Thomas1664:
> Finally found it. The pclose call also has to be protected by the mutex.

@autoantwort Thanks a lot for your help!

Thomas1664:
It is even documented that our use of `popen` was wrong: Although popen is thread safe, "It is even possible that calling MT-Safe functions in sequence does not yield an MT-Safe combination." (source: https://www.man7.org/linux/man-pages/man7/attributes.7.html ). Note that we might call this function in sequence across multiple threads: https://github.com/microsoft/vcpkg-tool/blob/c0e76df6a63b49b87266b3f7f195d631c671abf2/src/vcpkg/base/system.process.cpp#L485-L491

ras0219-msft:
> It is even documented that our use of `popen` was wrong: Although popen is thread safe, "It is even possible that calling MT-Safe functions in sequence does not yield an MT-Safe combination." (source: https://www.man7.org/linux/man-pages/man7/attributes.7.html ).

I think adding more context to the quote makes more sense:

> For example, having a thread call two MT-
              Safe functions one right after the other does not
              guarantee behavior equivalent to atomic execution of a
              combination of both functions, since concurrent calls in
              other threads may interfere in a destructive way.

My interpretation is that this is restating how
```
int a = x.fetch_add(1);
int b = x.fetch_add(1);
```
doesn't guarantee `b - a == 1` -- another thread could have modified `x` between. Thus, our use should still be permissible.

But all of this is largely philosophical. Given the evidence in this thread, I think adding a lock around `popen` and `pclose` has presented as a 100% fix to the issue (with a small sample size).

BillyONeal:
> But all of this is largely philosophical. Given the evidence in this thread, I think adding a lock around `popen` and `pclose` has presented as a 100% fix to the issue (with a small sample size).

Does that mean you are OK with this change? Note that it is currently unable to merge because you are marked 'request changes'

ras0219-msft (approved):

BillyONeal:
Thanks :)

PR #706, opened 2022-09-16
Optimize the message of integrate install

Cheney-W:
After the PR https://github.com/microsoft/vcpkg-tool/pull/619, the prompt information after executing `./vcpkg integrate install` under windows becomes the following:
image
The information here is all crowded together and the user experience will be bad.

For fixing this issue, I've added two newlines here, and the modified result looks like this:
image


ras0219-msft (approved):

ras0219-msft:
Thanks!

PR #720, opened 2022-09-28
Add command `lint-port`

autoantwort:
Adds command `x-lint-port` that checks for common port issues and the usage of deprecated functions. The command is also able to fix the most of these issues. 

Tests probably needed.


Thomas1664 (commented):
I know that this is still a draft.

In general you should be careful with adding const to local variables because this might avoid moving in some cases

src/vcpkg/portlint.cpp
```diff
@@ -0,0 +1,268 @@
...
+            {"Nunit",
+             "DEPRECATED: This license is based on the MIT license, except with an \"acknowledgement\" clause. That "
+             "clause makes it functionally equivalent to MIT with advertising (Free, but GPLv2/v3 incompatible)"},
+            {"StandardML-NJ", "SMLNJ"},
+            {"wxWindows",
+             "DEPRECATED: Use license expression including main license, \"WITH\" operator, and identifier: "
+             "WxWindows-exception-3.1"}};
+        Status status = Status::Ok;
+        auto& license = scf.core_paragraph->license.value_or_exit(VCPKG_LINE_INFO);
+        for (const auto dep_license : deprecated_licenses)
```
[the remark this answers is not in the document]
autoantwort: Hm this thing is only 32 bytes large (or 16 bytes on x86 windows)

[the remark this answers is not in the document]
Thomas1664: The type is `StringLiteral` which is a string view. IMO the best option is actually to remove const and replace `auto` with `StringView`. I updated my suggestion accordingly.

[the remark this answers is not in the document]
autoantwort: It is a pair of StringViews

Thomas1664:
In general, it might be possible to parallelize linting as well if there weren't any calls to `check_exit()` and alike.

src/vcpkg/portlint.cpp
```diff
@@ -0,0 +1,264 @@
...
+        if (scheme == VersionScheme::String)
+        {
+            if (DateVersion::try_parse(scf.core_paragraph->raw_version))
+            {
+                new_scheme = VersionScheme::Date;
+            }
+            else if (DotVersion::try_parse_relaxed(scf.core_paragraph->raw_version))
+            {
+                new_scheme = VersionScheme::Relaxed;
+            }
```
vicroms, line 30: Extract this to a separate function that can be unit tested.

```cpp
VersionScheme get_recommended_version_scheme(StringView raw_version, VersionScheme original_scheme)
{
    if (DateVersion::try_parse(raw_version)) return VersionScheme::Date;
    if (DotVersion::try_parse_relaxed(raw_version)) return VersionScheme::Relaxed;
    return original_scheme;
}
```

src/vcpkg/portlint.cpp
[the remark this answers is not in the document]
vicroms: I think this check can be extracted to its own unit-testable function too.

```cpp
StringView get_recommended_license(StringView original_license);
```

We should also have separate maps for inputs that map directly to a SPDX license expression and another one for those where you give recommendation messages, we should also make those recommendations into localizable strings.

src/vcpkg/portlint.cpp
```diff
@@ -0,0 +1,264 @@
...
+            {
+                license.replace(index, dep_license.first.size(), dep_license.second.c_str());
+                status |= Status::Fixed;
+            }
+        }
+
+        return status;
+    }
+
+    Status check_portfile_deprecated_functions(Filesystem& fs, SourceControlFileAndLocation& scf, Fix fix)
```
vicroms, line 136: I think you can create an override that operates with the contents of `portfile.cmake` and make the function that takes the `Filesystem` and `SCFL` call the new function,

```cpp
struct FixedPortfile {
    Status status,
    Optional<std::string> contents, // empty if Fix::NO
    Optional<std::set<StringLiteral>> added_host_deps // host-dependencies that need to be added, empty if Fix::NO
};
FixedPortfile check_portfile_deprecated_functions(StringView contents, StringVIew origin, Fix fix, MessageSink& warningsSink);

Status check_portfile_deprecated_functions(Filesystem& fs, SourceControlFileAndLocation& scf, Fix fix)
{
    auto contents = fs.read_contents(scf.source_location / "portfile.cmake", VCPKG_LINE_INFO);
    auto lint_result = check_portfile_deprecated_functions(contents, scf.source_control_file->core_paragraph->name, fix, stdout_sink);
    // add host dependencies and write fixed portfile (or do nothing if Fix::NO)
    return lint_result.status;
}
```

vicroms (changes requested):
Some parts of this PR can be made unit-testable with a few tweaks.

src/vcpkg/portlint.cpp
vicroms: (also for the sake of unit tests)

autoantwort:
@vicroms I have added unit tests

src/vcpkg/portlint.cpp
```diff
@@ -0,0 +1,384 @@
...
+                const auto end_param = portfile_content.find_first_of(" \n\t)", start_param);
+                if (end_param != std::string::npos && end_param <= end)
+                {
+                    const auto original_param = portfile_content.substr(start_param, end_param - start_param);
+                    auto param = StringView(original_param);
+                    if (Strings::starts_with(param, "share/"))
+                    {
+                        param = param.substr(StringLiteral("share/").size());
+                    }
+                    if (param == "${PORT}" || param == origin)
```
Neumann-A, line 261: `share/<packname>`. `packname` can be lowercase or `packageCase`. @dg0yt  prefers lowercase to avoid adding extra dirs if portname = packagename. 

However the case for `packageCaseConfig.cmake` needs to be correct. While the `-config.cmake` version should always be lowercase.

autoantwort: Yeah that is right. But it maybe break the portfile when it will replace some stuff in the files manually after they are moved. But I am fine with reverting the change

vicroms:
Needs a documentation draft PR to make it easier to understand what this is supposed to do and how.

autoantwort:
@vicroms I have created https://github.com/microsoft/vcpkg-docs/pull/23

autoantwort:
@vicroms The ci with the changes created by this PR is also successful https://github.com/microsoft/vcpkg/pull/26981

cnSchwarzer:
PS E:\vcpkg-tool> .\build\vcpkg.exe x-lint-port      
x-add-version with no arguments requires passing --all to update all port versions at once

x-add-version not changed

cnSchwarzer:
I find many ports with `usage` provided, but not installed in the share/${PORT} folder, I'm planning to add this check to this lint-port, do you guys agree?

autoantwort:
This sounds like a great idea! But I think it is easier to create a post build check for that.

cnSchwarzer:
I think we can do both, the post build check can warn user, and we can expand this flag to something like `--x-lint-port --fix-usage` to automatically fix all ports usage issue and actually eliminate the issue.

autoantwort:
I think you don't need a new flag for that. Simply use the `--fix` flag. Do you implement this and create a PR against my branch?

cnSchwarzer:
Yes, I will create a PR against this branch once I tested.

cnSchwarzer:
People use different ways to install usage, should vcpkg provide a unified `vcpkg_install_usage`?

autoantwort:
I think the current situation is fine.

autoantwort:
No, I am done and wait for review from @vicroms

talregev:
@vicroms @Neumann-A @Thomas1664 
Any news of review this PR?

include/vcpkg/portlint.h
```diff
@@ -0,0 +1,53 @@
...
+
+namespace vcpkg::Lint
+{
+    enum class Status : int
+    {
+        Ok = 0b0,
+        Problem = 0b01,
+        Fixed = 0b10,
+        PartiallyFixed = 0b11
+    };
```
Thomas1664, line 22: Why didn't you just write 0, 1, 2 and 3?
If you really want to safe some space, you could use `std::byte`.

Thomas1664 (commented):
My earlier comment to add static to some string literals is still unfixed but like my new comment not a big deal. 
I also noticed the use of `std::string::size_type` instead of `std::size_t` or `size_t` which is also not a problem.

src/vcpkg/portlint.cpp
```diff
@@ -0,0 +1,453 @@
...
+        }
+
+        struct ParamAndValue
+        {
+            std::string::size_type start_param = std::string::npos;
+            std::string::size_type start_value = std::string::npos;
+            std::string::size_type end_value = std::string::npos;
+            bool found() const { return start_param != std::string::npos; }
+            bool value_found() const { return start_value != std::string::npos && end_value != std::string::npos; }
+            std::string get_value(const std::string& content) const
```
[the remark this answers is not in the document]
autoantwort: No because the returned value is used longer then the content parameter.

src/vcpkg/portlint.cpp
```diff
@@ -0,0 +1,268 @@
...
+#include "vcpkg/base/system.print.h"
+#include "vcpkg/base/system.process.h"
+#include "vcpkg/base/util.h"
+#include "vcpkg/installedpaths.h"
+#include "vcpkg/vcpkgpaths.h"
+
+namespace vcpkg::Lint
+{
+
+    constexpr StringLiteral VERSION_RELAXED = "version";
```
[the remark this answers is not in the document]
autoantwort: Afaic https://stackoverflow.com/questions/59307723/static-constexpr-global-variable

include/vcpkg/portlint.h
autoantwort: Because I want to write 
```c++
Status status = Status::Ok;
if (... true) status |= Status::Problem;
if (... true) status |= Status::Fixed;
// status is now Status::PartiallyFixed
```

Thomas1664: smart!

talregev:
@autoantwort Are you going to fix the conflicts?

autoantwort:
I can fix them, but they don't prevent this PR from being reviewed.

talregev:
@Thomas1664 this PR is ready for review by @vicroms from your side?

Thomas1664:
> @Thomas1664 this PR is ready for review by @vicroms from your side?

The vcpkg team reviews PRs independently from community feedback such as my review comments.

talregev:
Any news?

BillyONeal:
> Any news?

I think this keeps getting skipped over because we need to sell a concrete proposal of what this command actually does to our PM team and management, and we aren't getting time to reverse engineer what that looks like from this PR in order to do so.

talregev:
> > Any news?
> 
> I think this keeps getting skipped over because we need to sell a concrete proposal of what this command actually does to our PM team and management, and we aren't getting time to reverse engineer what that looks like from this PR in order to do so.

All you need is time?

BillyONeal:
> All you need is time?

We need like a spec of what the command is supposed to do

PR #737, opened 2022-10-09
[versioning] Add message when vcpkg is cloned to a shallow repository

vicroms:
Appends an additional error message when a built-in repository error occurs and vcpkg is cloned into a shallow repository.

Rough implementation. 

Improvements to how we handle versioning errors can be made but I believe is better to wait until #556 is merged to make more extensive changes. 

On baseline not in git history:
image

On version not in git history:
image

### DIFF
--- a/include/vcpkg/vcpkgpaths.h
+++ b/include/vcpkg/vcpkgpaths.h
@@ -136,7 +136,6 @@ namespace vcpkg
         std::string get_current_git_sha_baseline_message() const;
         ExpectedS<Path> git_checkout_port(StringView port_name, StringView git_tree, const Path& dot_git_dir) const;
         ExpectedL<std::string> git_show(StringView treeish, const Path& dot_git_dir) const;
-        bool git_is_shallow_clone(const Path& dot_git_dir) const;
 
         const DownloadManager& get_download_manager() const;
 
--- a/include/vcpkg/vcpkgpaths.h
+++ b/include/vcpkg/vcpkgpaths.h
@@ -136,6 +136,7 @@ namespace vcpkg
         std::string get_current_git_sha_baseline_message() const;
         ExpectedS<Path> git_checkout_port(StringView port_name, StringView git_tree, const Path& dot_git_dir) const;
         ExpectedL<std::string> git_show(StringView treeish, const Path& dot_git_dir) const;
+        bool git_is_shallow_clone(const Path& dot_git_dir) const;
 
         const DownloadManager& get_download_manager() const;
 
--- a/src/vcpkg/base/git.cpp
+++ b/src/vcpkg/base/git.cpp
@@ -6,6 +6,8 @@
 #include <vcpkg/base/stringview.h>
 #include <vcpkg/base/system.process.h>
 
+#include <vcpkg/tools.h>
+
 namespace
 {
     using namespace vcpkg;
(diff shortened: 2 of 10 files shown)


src/vcpkg/vcpkgpaths.cpp
```diff
@@ -993,6 +999,19 @@ namespace vcpkg
...
+    {
+        auto cmd =
+            git_cmd_builder(dot_git_dir, dot_git_dir).string_arg("rev-parse").string_arg("--is-shallow-repository");
+        auto output = flatten_out(cmd_execute_and_capture_output(cmd), Tools::GIT);
+        if (!output)
+        {
+            Debug::println(output.error());
+            return false;
+        }
+        return Strings::case_insensitive_ascii_equals("true", Strings::trim(output.value_or_exit(VCPKG_LINE_INFO)));
```
Thomas1664, line 1012:
```suggestion
        return Strings::case_insensitive_ascii_equals("true", Strings::trim(*output.get()));
```

It will never exit here because errors were caught by the if statement above.

autoantwort:
If not done: We should also check if VCPKG_ROOT is a git repo at all. Some users download the repo instead of clone it.

include/vcpkg/vcpkgpaths.h
```diff
@@ -136,6 +136,7 @@ namespace vcpkg
         std::string get_current_git_sha_baseline_message() const;
         ExpectedS<Path> git_checkout_port(StringView port_name, StringView git_tree, const Path& dot_git_dir) const;
         ExpectedL<std::string> git_show(StringView treeish, const Path& dot_git_dir) const;
+        bool git_is_shallow_clone(const Path& dot_git_dir) const;
```
Thomas1664, line 139: Would `base/git.h` be a better place for this?

src/vcpkg/base/git.cpp
```diff
@@ -187,4 +189,15 @@ namespace vcpkg
...
+
+    ExpectedL<bool> is_shallow_clone(const GitConfig& config)
+    {
+        auto cmd = git_cmd_builder(config).string_arg("rev-parse").string_arg("--is-shallow-repository");
+        auto output = flatten_out(cmd_execute_and_capture_output(cmd), Tools::GIT);
+        if (!output)
+        {
+            return output.error();
+        }
+        return "true" == Strings::trim(*output.get());
```
BillyONeal, line 201:
```suggestion
        return flatten_out(cmd_execute_and_capture_output(cmd), Tools::GIT).map([](std::string&& output) {
            return "true" == Strings::trim(std::move(output));
        });
```

src/vcpkg/vcpkgpaths.cpp
```diff
@@ -971,8 +971,16 @@ namespace vcpkg
         }
         return ret;
     }
+
     std::string VcpkgPaths::get_current_git_sha_baseline_message() const
     {
+        const auto& git_config = git_builtin_config();
+        auto maybe_shallow_clone = is_shallow_clone(git_config).get();
+        if (maybe_shallow_clone && *maybe_shallow_clone)
```
BillyONeal, line 979: This is a use after free; the pointer returned by get() points to the temporary ExpectedL in this expression.

I think you should add:

```
        template<class... Args>
        T value_or(Args&&... or_args) &&
        {
            auto maybe_value = m_t.get();
            if (maybe_value)
            {
                return std::move(*maybe_value);
            }

            return T(std::forward<Args>(or_args)...);
        }

        template<class... Args>
        T value_or(Args&&... or_args) const&
        {
            auto maybe_value = m_t.get();
            if (maybe_value)
            {
                return *maybe_value;
            }

            return T(std::forward<Args>(or_args)...);
        }
```

to `expected.h` right above the `value_or_exit` overloads, then change this to:

```suggestion
        if (is_shallow_clone(git_config).value_or(false))
```

BillyONeal (changes requested):

BillyONeal:
> On version not in git history: image

I think we should get rid of the "tar failed" part of this message since that's already implied by the message?

BillyONeal (approved):

vicroms:
Output cleanup will be made in a separate PR since it will conflict with ongoing localization efforts by Javier

PR #763, opened 2022-10-21
Use brace initializer instead of empty C string to return empty C++ string

AtariDreams:


autoantwort:
Imho with an empty string it is more obvious what the code is doing

AtariDreams:
I mean it also passes in the string to a constructor when 0 initialization is the proper way to do it

JavierMatosD (approved):

PR #772, opened 2022-10-26
Documenting our use of VSLANG

JavierMatosD:
This PR adds documentation for how to change the language in vcpkg ( i.e., documenting our use of the VSLANG environment variable).

### DIFF
--- a/docs/localization.md
+++ b/docs/localization.md
@@ -136,3 +136,21 @@ namespace vcpkg::Commands
     }
 }
 ```
+## Selecting a Language
+
+Vcpkg chooses a language through the VSLANG environment variable, which should be set to a valid LCID (locale identifier, 4-byte value representing a language). Users can set VSLANG to any of the following LCIDs:
+
+1029: Czech                 \
+1031: German                \
+1033: English               \
+3082: Spanish (Spain)       \
+1036: French                \
+1040: Italian               \
+1041: Japanese              \
+1042: Korean                \
+1045: Polish                \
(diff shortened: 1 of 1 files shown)


docs/localization.md
```diff
@@ -136,3 +136,21 @@ namespace vcpkg::Commands
     }
 }
 ```
+## Selecting a Language
+
+Vcpkg chooses a language through the VSLANG environment variable, which should be set to a valid LCID (locale identifier, 4-byte value representing a language). Users can set VSLANG to any of the following LCIDs:
```
vicroms, line 141:
```suggestion
vcpkg chooses a language through the `VSLANG` environment variable, which should be set to a valid LCID (locale identifier, 4-byte value representing a language). Users can set `VSLANG` to any of the following LCIDs:
```
For branding consistency we never capitalize vcpkg.

We should mention that if vcpkg is given an invalid LCID it falls back to English.

vicroms (changes requested):

vicroms:
Should we point out that the Visual Studio setting may override VSLANG from the user's environment? Has that been tested?

BillyONeal (approved):
@vicroms

> Should we point out that the Visual Studio setting may override VSLANG from the user's environment? Has that been tested?

This is vcpkg-team-facing not end-user-facing documentation. We certainly are not suggesting that users set anything like that. I'm not sure why we would be testing entirely-internal-to-VS behavior.

PR #778, opened 2022-10-28
[RFC][registries] Allow pattern matching in packages declarations

vicroms:
This PR makes the following changes to allow limited pattern matching for package declarations in registries:

* Allow prefixes in the form `foo*` in `"packages"` declarations. 
* The `*` character is a wildcard that matches any zero or more characters. 
* The port name resolution priority will change to:
  1. Exact match
  2. Pattern match
      1. Longer prefixes have higher priority, e.g.: when resolving `boost`, `boost*` > `b*` > `*`.
       2. If there is a tie, we use the registry that is declared first in the `"packages"` array. 
          A warning will be emitted to notify the user of possible misconfiguration.
  3. Default registry

Example: expose `"beicode"` and `"beison"` via patterns.

`vcpkg-configuration.json`
```json
{
  "registries": [
    {
      "kind": "git", 
      "repository": "https://github.com/northwindtraders/vcpkg-registry", 
      "baseline": "dacf4de488094a384ca2c202b923ccc097956e0c", 
      "packages": [ "bei*" ] 
    }
  ]
}
```

Checklist:
- [x] Changes to port name resolution logic
  - [X] Implement resolution priority
  - [x] Implement warning on masked registries
- [X] Change configuration parser
- [X] Unit tests
- [x] Documentation (PR https://github.com/microsoft/vcpkg/pull/27686)


Neumann-A:
oh no. If this is implemented please also add an exclusion list ...

BillyONeal:
The "exclusion list" is adding the excluded thing to the registry you actually want used.

src/vcpkg/registries.cpp
```diff
@@ -1147,15 +1147,75 @@ namespace vcpkg
...
+                return (++cur == last) ? static_cast<int>(idx) : -1;
+            }
+
+            if (idx >= name.size()) return -1;
+            if (c != name[idx]) return -1;
+        }
+
+        // exact match, increase priority by 1
+        if (idx == name.size()) return name.size() + 1;
+        return static_cast<int>(idx);
```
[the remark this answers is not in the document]
BillyONeal: How about:

```diff
diff --git a/include/vcpkg/registries.h b/include/vcpkg/registries.h
index 799648a0..e635b1b7 100644
--- a/include/vcpkg/registries.h
+++ b/include/vcpkg/registries.h
@@ -103,13 +103,6 @@ namespace vcpkg
         std::unique_ptr<RegistryImplementation> implementation_;
     };
 
-    struct RegistryCandidate
-    {
-        const RegistryImplementation* implementation;
-        int priority;
-        size_t index;
-    };
-
     // this type implements the registry fall back logic from the registries RFC:
     // A port name maps to one of the non-default registries if that registry declares
     // that it is the registry for that port name, else it maps to the default registry
@@ -132,7 +125,7 @@ namespace vcpkg
         //
         // It is guaranteed to at least contain one element, the default registry.
         // NOTE: the default registry implementation can be null.
-        std::vector<RegistryCandidate> registries_for_port(StringView name) const;
+        std::vector<const RegistryImplementation*> registries_for_port(StringView name) const;
 
         ExpectedL<Version> baseline_for_port(StringView port_name) const;
 
@@ -212,6 +205,5 @@ namespace vcpkg
         VersionDbEntryDeserializer underlying;
     };
 
-    // exposed for testing
-    int compare_package_prefix(StringView name, StringView pattern);
+    size_t package_match_prefix(StringView name, StringView pattern);
 }
diff --git a/src/vcpkg-test/registries.cpp b/src/vcpkg-test/registries.cpp
index ec0c751b..0c1118f7 100644
--- a/src/vcpkg-test/registries.cpp
+++ b/src/vcpkg-test/registries.cpp
@@ -118,19 +118,19 @@ TEST_CASE ("check valid package patterns", "[registries]")
 
 TEST_CASE ("calculate prefix priority", "[registries]")
 {
-    CHECK(compare_package_prefix("boost", "*") == 0);
-    CHECK(compare_package_prefix("boost", "b*") == 1);
-    CHECK(compare_package_prefix("boost", "boost*") == 5);
-    CHECK(compare_package_prefix("boost", "boost") == 6);
-
-    CHECK(compare_package_prefix("", "") == -1);
-    CHECK(compare_package_prefix("", "*") == -1);
-    CHECK(compare_package_prefix("", "a") == -1);
-    CHECK(compare_package_prefix("boost", "") == -1);
-    CHECK(compare_package_prefix("boost", "c*") == -1);
-    CHECK(compare_package_prefix("boost", "*c") == -1);
-    CHECK(compare_package_prefix("boost", "c**") == -1);
-    CHECK(compare_package_prefix("boost", "c*a") == -1);
+    CHECK(package_match_prefix("boost", "*") == 1);
+    CHECK(package_match_prefix("boost", "b*") == 2);
+    CHECK(package_match_prefix("boost", "boost*") == 6);
+    CHECK(package_match_prefix("boost", "boost") == SIZE_MAX);
+
+    CHECK(package_match_prefix("", "") == SIZE_MAX);
+    CHECK(package_match_prefix("", "*") == 1);
+    CHECK(package_match_prefix("", "a") == 0);
+    CHECK(package_match_prefix("boost", "") == 0);
+    CHECK(package_match_prefix("boost", "c*") == 0);
+    CHECK(package_match_prefix("boost", "*c") == 0);
+    CHECK(package_match_prefix("boost", "c**") == 0);
+    CHECK(package_match_prefix("boost", "c*a") == 0);
 }
 
 TEST_CASE ("select highest priority registry", "[registries]")
@@ -183,19 +183,19 @@ TEST_CASE ("sort candidate registries by priority", "[registries]")
         REQUIRE(candidates.size() == 4);
         size_t idx = 0;
 
-        auto reg = candidates[idx++].implementation;
+        auto reg = candidates[idx++];
         REQUIRE(reg);
         CHECK(get_tri_num(*reg) == 4);
 
-        reg = candidates[idx++].implementation;
+        reg = candidates[idx++];
         REQUIRE(reg);
         CHECK(get_tri_num(*reg) == 3);
 
-        reg = candidates[idx++].implementation;
+        reg = candidates[idx++];
         REQUIRE(reg);
         CHECK(get_tri_num(*reg) == 1);
 
-        reg = candidates[idx++].implementation;
+        reg = candidates[idx++];
         REQUIRE(reg);
         CHECK(get_tri_num(*reg) == 2);
     }
@@ -209,9 +209,7 @@ TEST_CASE ("sort candidate registries by priority", "[registries]")
         RegistrySet set(nullptr, std::move(rs));
 
         auto candidates = set.registries_for_port("cpprestsdk");
-        REQUIRE(candidates.size() == 1);
-        auto reg = candidates.front().implementation;
-        CHECK_FALSE(reg);
+        REQUIRE(candidates.empty());
     }
 }
 
diff --git a/src/vcpkg/registries.cpp b/src/vcpkg/registries.cpp
index 820aeb42..2dcb26df 100644
--- a/src/vcpkg/registries.cpp
+++ b/src/vcpkg/registries.cpp
@@ -1148,74 +1148,71 @@ namespace vcpkg
     const RegistryImplementation* RegistrySet::registry_for_port(StringView name) const
     {
         auto candidates = registries_for_port(name);
-        auto registry = candidates.front();
-        return registry.implementation;
+        if (candidates.empty())
+        {
+            return default_registry();
+        }
+
+        return candidates[0];
     }
 
-    int compare_package_prefix(StringView name, StringView prefix)
+    size_t package_match_prefix(StringView name, StringView prefix)
     {
-        if (prefix.empty() || name.empty()) return -1;
+        if (name == prefix)
+        {
+            // exact match is like matching "infinity" prefix
+            return SIZE_MAX;
+        }
 
-        size_t idx = 0;
-        auto cur = prefix.begin();
-        auto last = prefix.end();
-        for (; cur != last; ++idx, ++cur)
+        // Note that the * is included in the match so that 0 means no match
+        const auto prefix_size = prefix.size();
+        if (prefix_size != 0)
         {
-            const auto c = *cur;
-            if (c == '*')
+            const auto star_index = prefix_size - 1;
+            if (prefix[star_index] == '*' && name.size() >= star_index &&
+                name.substr(0, star_index) == prefix.substr(0, star_index))
             {
-                return (++cur == last) ? static_cast<int>(idx) : -1;
+                return prefix_size;
             }
-
-            if (idx >= name.size()) return -1;
-            if (c != name[idx]) return -1;
         }
 
-        // exact match, increase priority by 1
-        if (idx == name.size()) return name.size() + 1;
-        return static_cast<int>(idx);
+        return 0;
     }
 
-    std::vector<RegistryCandidate> RegistrySet::registries_for_port(StringView name) const
+    std::vector<const RegistryImplementation*> RegistrySet::registries_for_port(StringView name) const
     {
-        size_t registry_index = 0;
+        struct RegistryCandidate
+        {
+            const RegistryImplementation* impl;
+            std::size_t matched_prefix;
+        };
+
         std::vector<RegistryCandidate> candidates;
-        for (auto it = registries().begin(); it != registries().end(); ++it, ++registry_index)
+        for (auto&& registry : registries())
         {
-            int priority = -1;
-            for (auto&& pattern : it->packages())
+            std::size_t longest_prefix = 0;
+            for (auto&& package : registry.packages())
             {
-                priority = std::max(priority, compare_package_prefix(name, pattern));
+                longest_prefix = std::max(longest_prefix, package_match_prefix(name, package));
             }
 
-            if (priority >= 0)
+            if (longest_prefix != 0)
             {
-                candidates.emplace_back(RegistryCandidate{
-                    &it->implementation(),
-                    priority,
-                    registry_index,
-                });
+                candidates.push_back({&registry.implementation(), longest_prefix});
             }
         }
 
         if (candidates.empty())
         {
-            candidates.emplace_back(RegistryCandidate{
-                default_registry(),
-                0,
-                0,
-            });
+            return std::vector<const RegistryImplementation*>();
         }
-        else
-        {
-            Util::sort(candidates, [](auto&& lhs, auto&& rhs) {
-                if (lhs.priority > rhs.priority) return true;
-                if (lhs.priority == rhs.priority) return lhs.index < rhs.index;
-                return false;
+
+        std::stable_sort(
+            candidates.begin(), candidates.end(), [](const RegistryCandidate& lhs, const RegistryCandidate& rhs) {
+                return lhs.matched_prefix > rhs.matched_prefix;
             });
-        }
 
-        return candidates;
+        return Util::fmap(std::move(candidates), [](const RegistryCandidate& target) { return target.impl; });
     }
 
     ExpectedL<Version> RegistrySet::baseline_for_port(StringView port_name) const
```

include/vcpkg/registries.h
```diff
@@ -196,4 +211,7 @@ namespace vcpkg
     private:
         VersionDbEntryDeserializer underlying;
     };
+
+    // exposed for testing
```
BillyONeal, line 215: @ras0219-msft This is what I meant about "most details are public to someone" on Thursday. This thing is public, and tested like it is public.

```suggestion
```

src/vcpkg/configuration.cpp
```diff
@@ -679,6 +688,71 @@ namespace vcpkg
 
     Json::IDeserializer<Configuration>& get_configuration_deserializer() { return ConfigurationDeserializer::instance; }
 
+    Optional<Configuration> parse_configuration(const Filesystem& fs, const Path& path, MessageSink& messageSink)
```
BillyONeal, line 691: From an interface perspective, I have a mild preference to require callers to pass in the content rather than a path. Then this thing need not depend on filesystem at all.

src/vcpkg/configuration.cpp
```diff
@@ -679,6 +688,71 @@ namespace vcpkg
...
+    {
+        if (!fs.exists(path, IgnoreErrors{}))
+        {
+            return nullopt;
+        }
+
```
BillyONeal, line 697:
```suggestion
```

The `if (ec)` part below already has this effect.

src/vcpkg/configuration.cpp
```diff
@@ -679,6 +688,71 @@ namespace vcpkg
...
+            }
+            else
+            {
+                messageSink.println(Color::warning, msgWarnOnParseConfig, msg::path = origin);
+            }
+
+            for (auto&& msg : reader.warnings())
+            {
+                auto warning = fmt::format("    {}\n", msg);
+                messageSink.println(Color::warning, LocalizedString::from_raw(warning));
```
BillyONeal, line 742:
```suggestion
                messageSink.println(Color::warning, LocalizedString().append_indent().append_raw(msg));
```

Ditto below for error

BillyONeal (changes requested):
"Request changes" over the threading concerns in making the serializers mutable. Other changes are suggestions.

src/vcpkg/base/json.cpp
```diff
@@ -1502,4 +1502,74 @@ namespace vcpkg::Json
         }
         return sv.to_string();
     }
+
+    bool PackagePatternDeserializer::is_package_pattern(StringView sv)
+    {
+        if (PackageNameDeserializer::is_package_name(sv))
```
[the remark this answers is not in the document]
vicroms: Agree

src/vcpkg/registries.cpp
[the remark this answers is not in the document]
vicroms: These changes seem fair, and we don't expose the prioritization details.

src/vcpkg/base/json.cpp
```diff
@@ -1502,4 +1502,74 @@ namespace vcpkg::Json
...
+    Optional<std::string> PackagePatternDeserializer::visit_string(Json::Reader& r, StringView sv)
+    {
+        if (!is_package_pattern(sv))
+        {
+            return nullopt;
+        }
+
+        if (m_patterns)
+        {
+            auto it = m_patterns->find(sv.to_string());
```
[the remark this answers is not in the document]
vicroms: I'll modify the reader to make this unnecessary.

src/vcpkg-test/registries.cpp
```diff
@@ -225,6 +355,88 @@ TEST_CASE ("registry_parsing", "[registries]")
...
+    $.registries[0].packages[0] (registry: https://github.com/northwindtraders/vcpkg-registry)
+    $.registries[1].packages[0] (registry: git@github.com:northwindtraders/vcpkg-registry)
+Package "beison" is declared in multiple locations:
+    $.registries[0].packages[1] (registry: https://github.com/northwindtraders/vcpkg-registry)
+    $.registries[2].packages[0] (registry: https://github.com/Microsoft/vcpkg)
+Package "fmt" is declared in multiple locations:
+    $.registries[1].packages[2] (registry: git@github.com:northwindtraders/vcpkg-registry)
+    $.registries[2].packages[1] (registry: https://github.com/Microsoft/vcpkg)
+Duplicate entries will be ignored.
+Remove any duplicate entries to dismiss this warning.)");
```
vicroms, line 437: This warning message needs to be improved.

src/vcpkg/configuration.cpp
vicroms: I like not having to repeat the file validation boilerplate, but I'll think about it.

BillyONeal: > I like not having to repeat the file validation boilerplate, but I'll think about it.

Almost all callers can pass `fs.read_content(path, VCPKG_LINE_INFO)` and get identical behavior.

Neumann-A:
> The "exclusion list" is adding the excluded thing to the registry you actually want used.

Should be documented.

src/vcpkg/base/json.cpp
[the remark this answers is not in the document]
vicroms: Unfortunately, it is documented via an error message:

https://github.com/microsoft/vcpkg-tool/blob/6967066fd77b8a4c94e0748ab81a5a967ce6bb4e/src/vcpkg/sourceparagraph.cpp#L455-L459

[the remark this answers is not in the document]
BillyONeal: I understand it is in the error message but I don't think this counts as documentation unless there is a practical use for such a thing. (As in, I think the rest of the product tried to interpret the packages array literally rather than splitting by dots)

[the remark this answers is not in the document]
vicroms: It is gone.

src/vcpkg/base/json.cpp
```diff
@@ -1351,6 +1351,8 @@ namespace vcpkg::Json
     }
     // } auto stringify()
 
+    uint64_t get_json_parsing_stats() { return g_json_parsing_stats.load(); }
```
vicroms, line 1354: Moved without changes.

vicroms:
> > The "exclusion list" is adding the excluded thing to the registry you actually want used.
> 
> Should be documented.

I've added an example in the docs PR.

src/vcpkg/configuration.cpp
```diff
@@ -387,6 +382,81 @@ namespace
...
+    std::vector<LocalizedString> collect_package_pattern_warnings(const std::vector<RegistryConfig>& registries)
+    {
+        struct LocationAndRegistry
+        {
+            StringView location;
+            StringView registry;
+        };
+
+        // handle warnings from package pattern declarations
+        std::map<StringView, std::vector<LocationAndRegistry>> patterns;
```
BillyONeal, line 394: Key of `StringView` is a bit scary in terms of ownership? Might be worth either using `string` instead or commenting on how the lifetime works.

src/vcpkg/configuration.cpp
```diff
@@ -387,6 +382,81 @@ namespace
...
+        for (auto&& reg : registries)
+        {
+            if (auto packages = reg.package_declarations.get())
+            {
+                for (auto&& pkg : *packages)
+                {
+                    auto it = patterns.find(pkg.pattern);
+                    if (it == patterns.end())
+                    {
+                        it = patterns.emplace(pkg.pattern, std::vector<LocationAndRegistry>{}).first;
```
BillyONeal, line 404:
```suggestion
                        it = patterns.emplace(pkg.pattern).first;
```

src/vcpkg/configuration.cpp
```diff
@@ -387,6 +382,81 @@ namespace
...
+
+        // handle warnings from package pattern declarations
+        std::map<StringView, std::vector<LocationAndRegistry>> patterns;
+        for (auto&& reg : registries)
+        {
+            if (auto packages = reg.package_declarations.get())
+            {
+                for (auto&& pkg : *packages)
+                {
+                    auto it = patterns.find(pkg.pattern);
```
BillyONeal, line 401: Technically you could avoid a double lookup here by using `lower_bound` instead but I don't think it really matters.

src/vcpkg/configuration.cpp
```diff
@@ -387,6 +382,81 @@ namespace
...
+                    }
+                    it->second.emplace_back(LocationAndRegistry{
+                        pkg.location,
+                        reg.pretty_location(),
+                    });
+                }
+            }
+        }
+
+        auto append_declaration_warning = [](LocalizedString& msg,
```
BillyONeal, line 414: I think this should be a function rather than a lambda.

src/vcpkg/configuration.cpp
```diff
@@ -387,6 +382,81 @@ namespace
...
+                for (auto cur = first + 1; cur != last; ++cur)
+                {
+                    append_declaration_warning(warning, cur->location, cur->registry, 2);
+                    if (cur + 1 != last)
+                    {
+                        warning.append_raw("\n");
+                    }
+                }
+                warnings.emplace_back(warning);
+            }
```
BillyONeal, line 455:
```suggestion
            const auto& locations = kvpair.second;
            if (locations.size() > 1)
            {
                auto first = locations.begin();
                const auto last = locations.end();
                auto warning = msg::format_warning(msgDuplicatePackagePattern, msg::package_name = pattern)
                                   .append_raw("\n")
                                   .append_indent()
                                   .append(msgDuplicatePackagePatternFirstOcurrence)
                                   .append_raw("\n");
                append_declaration_warning(warning, first->location, first->registry, 2);
                warning.append_raw("\n")
                    .append_indent()
                    .append(msgDuplicatePackagePatternIgnoredLocations)
                    .append_raw("\n");
                    
                ++first;
                append_declaration_warning(warning, first->location, first->registry, 2);
                while (++first != last)
                {
                    warning.append_raw("\n");
                    append_declaration_warning(warning, first->location, first->registry, 2);
                }
                warnings.emplace_back(warning);
            }
```

Suggest eliminating the second iterator and avoiding the first/if antipattern by using the normal "join" pattern.

BillyONeal (approved):

BillyONeal (approved):

PR #802, opened 2022-11-09
Add `x-test-features` and `format-feature-baseline`

autoantwort:
Introduces a new command `x-test-features`. It can be passed a list of port names or `--all` to specify for which ports the tests should be run. 
The following tests are performed:
- The core feature gets installed with default featured disabled (or `--no-core`)
- Every feature gets installed with default featured disabled (or `--no-separated`)
- All features combined gets installed (or `--no-combined`)

Feature which are unsupported or marked as cascade for the target triplet are skipped.  
For every test only the necessary dependencies are installed (like vcpkg install in manifest mode). 

The idea is to run the feature test for all ports that are changed by a PR. 

Ports that cascade must be explicitly marked as cascade. 

`vcpkg format-feature-baseline scripts/ci.feature.baseline.txt` is a command to format the feature baseline to enforce that all entries in a "block" are sorted alphabetically. 

Docs: https://github.com/microsoft/vcpkg-docs/pull/131

Fixes https://github.com/microsoft/vcpkg/issues/13119

https://github.com/microsoft/vcpkg-tool/pull/908 reduce the needed ci time, this PR increases it again. So the actual change is not likely to be very large.


src/vcpkg/commands.ci.cpp
```diff
@@ -101,26 +102,38 @@ namespace vcpkg::Commands::CI
     static constexpr StringLiteral OPTION_OUTPUT_HASHES = "output-hashes";
     static constexpr StringLiteral OPTION_PARENT_HASHES = "parent-hashes";
     static constexpr StringLiteral OPTION_SKIPPED_CASCADE_COUNT = "x-skipped-cascade-count";
+    static constexpr StringLiteral OPTION_TEST_FEATURE_CORE = "test-feature-core";
+    static constexpr StringLiteral OPTION_TEST_FEATURES_COMBINED = "test-features-combined";
+    static constexpr StringLiteral OPTION_TEST_FEATURES_SEPERATLY = "test-features-seperatly";
```
dg0yt, line 107: `...SEPARATELY  ... -separately`

autoantwort:
@vicroms @ras0219-msft Should this be extracted into an own command? The only interaction point between the current ci command and the feature test thing is that ports that failed to build in the feature test setting are not build again by the ci command (we build all packages) because they maybe will then suddenly work because dependencies are installed that are not listed in the port. The ports are then stored in the binary cache and the feature test will then also works because the package is already in the binary cache even through it would normally fail to build.  
But this could also be solved by emitting a `known-failures` file by the new `test-features` command which is simply a list of abi's that are known to fail. This file is then passed to the ci command and the ci command then does not try to build these packages and we don't have the above mentioned problem.

ras0219-msft:
Docs: https://github.com/microsoft/vcpkg-docs/pull/131

vicroms:
@AugP @BillyONeal @dan-shaw @ras0219-msft 

Hi @autoantwort 

We discussed this new feature and have the following asks:

* Make it an experimental command: `x-test-features`
* Option names in vcpkg use the `--no-<toggle>` pattern instead of `--dont-<toggle>`

We haven't made a review of the implementation yet, but will probably have some more comments later.

dg0yt:
The commands removes all subdirs in `buildtrees/<port>/src`. Can this be avoided, or limited to `.clean` dirs? I would like to test ports without loosing temporary git repos, or editable dirs.

dg0yt:
BTW could it test an unregistered modification, maybe via overlay ports?

autoantwort:
So you want a flag like `--editable`?

dg0yt:
I tested this PR while working locally on a vcpkg PR (zeroc-ice). It is quite helpful but at the moment,
- I loose my git dirs for editing patches, without warning.
- I hardly notice that it doesn't test the current state of my ports dir.

> So you want a flag like `--editable`? 

My comments were independent and shouldn't be mixed.  
My first comment was about not touching *unrelated* source dirs in `buildtrees` (first problem).  
My second comment was about the state of the `ports` dir (second problem).

With `vcpkg install <port>`; `--editable` has a different meaning: It is about not cleaning the *actual* source dir in `buildtrees`. This might be another option, but seems less important to me. I can transfer changes from editable state to proper patches before running test-features. And it is fragile: features may cause source dir modifications.

So I don't ask for a flag but for a general modification.

autoantwort:
> I hardly notice that it doesn't test the current state of my ports dir.

It does that. 

> I loose my git dirs for editing patches, without warning.

The behavior is the same as for `x-set-installed`, so `--clean-buildtrees-after-build` is on by default. I can change that. But I would personally prefer an inverted switch like `--keep-buildtrees-after-build`.

dg0yt:
> It does that.

Indeed, you are right.

I stumbled over another behaviour now: Post-build check errors of dependencies cause cascades which prevent the test of the port of interest. Understandable from a CI point of view, but less convenient for interactive testing. (In my case: readline-unix creates empty bin directories for x64-linux-dynamic. Good to have this covered by post-build checks, but I don't want to be stopped by this....)

autoantwort:
Hm should I add a flag like `--ignore-port-checks`? As a workaround you could install the dependency outside of this command. Then it gets restored from the binary cache the next time, if you don't want to fix the check now :)

dg0yt:
> We haven't made a review of the implementation yet, but will probably have some more comments later.

@vicroms This PR is waiting for MS.

JavierMatosD:
Tagging this for internal tracking

src/vcpkg/binarycaching.cpp
```diff
@@ -2166,11 +2166,20 @@ namespace vcpkg
         m_config.read.push_back(std::move(provider));
     }
 
-    std::vector<CacheAvailability> ReadOnlyBinaryCache::precheck(View<InstallPlanAction> actions)
+    void ReadOnlyBinaryCache::clear_cache()
```
BillyONeal, line 2169: It seems very wrong that this is necessary; different sets of enabled features should result in different ABI SHAs. If they are not that is a serious bug that must be addressed.

BillyONeal: Oh, it isn't "clearing" the cache, it's that we blew away 'packages' so it isn't actually restored anymore.

src/vcpkg/commands.test-features.cpp
```diff
@@ -323,7 +323,7 @@ namespace vcpkg
         msg::println(msgPrecheckBinaryCache);
         binary_cache.precheck(actions_to_check);
 
-        Util::stable_sort(install_plans, [](const auto& left, const auto& right) {
+        Util::stable_sort(install_plans, [](const InstallPlanAction& left, const InstallPlanAction& right) noexcept {
```
autoantwort, line 326: So you prefer `(const std::pair<FullPackageSpec, InstallPlanAction>& left, const std::pair<FullPackageSpec, InstallPlanAction>& right)` over `(const auto& left, const auto& right)`?

BillyONeal: I tend to prefer not making lambdas generic when they don't have to be. I think it's both easier to read, because you can look at the body of the lambda and understand what it does without needing to see everywhere it is used, and it confuses debuggers and similar tools less. Contrast this with variables, where I think `auto` is always fine, because variables can never be understood without all the context around them, and that form generally doesn't confuse tools.

Admittedly, it's probably closer to a style choice than a technical one.

I think generic lambdas are fine when something actually uses that genericity. Long enough type names to prefer `auto` probably exist, but  here, `InstallPlanAction` isn't much longer than `auto` and doesn't even introduce an extra newline.

BillyONeal:
RE: Fix build

Sorry, I pushed to this branch when I left for the day, not when I thought it was in a specifically reviewable state. I shouldn't have used someone else's branch for that, but my upstream was still set here.

src/vcpkg/binarycaching.cpp
```diff
@@ -2536,6 +2536,15 @@ namespace vcpkg
                         generate_nuspec(request.package_dir, action, m_config.nuget_prefix, m_config.nuget_repo);
                 }
 
+                // With x-test-features ports gets build multiple times and the packages folder gets deleted
```
BillyONeal, line 2539: Unfortunately I think this doesn't work / races still because a second install wanting packages from a previous install that haven't finished being made yet is likely; I think we have to block between installs.

BillyONeal: How do you feel about https://github.com/microsoft/vcpkg-tool/pull/802/commits/2fbb847622490df6bf9d13402b84cac8752c840b ?

src/vcpkg/commands.format-feature-baseline.cpp
```diff
@@ -0,0 +1,92 @@
...
+        {msgCmdFormatFeatureBaselineExample},
+        Undocumented,
+        AutocompletePriority::Public,
+        1,
+        1,
+        {{}, {}, {}},
+        nullptr,
+    };
+
+    void command_format_feature_baseline_and_exit(const VcpkgCmdArguments& args, const Filesystem& fs)
```
BillyONeal, line 67: 2 problems with this:

1. The assumption that a block starts with a commented line isn't necessarily true. Sometimes, the comment needs to follow the block. I think the existence of this thing is probably worth including even if we don't have a great solution to it though. Clearly this one wasn't in the right place before either:
  missorting example of ms-gltf
2. This doesn't actually lexically sort port names. I think we can fudge this by just sorting : earlier? 
  misssorting example cmake and cmake-user

BillyONeal: I fixed this and applied to vcpkg's current registry here: https://github.com/microsoft/vcpkg/pull/44136/

src/vcpkg/binarycaching.cpp
autoantwort: I think that would be a huge slowdown (at least for me where the binary cache latency is high and with `-all`).
The worst outcome of the "race" is that a package is build twice since it is not uploaded in the binary cache before it is needed a second time. And the way x-test-features works, at least two other ports will be built before the packet with the same ABI is used a second time so that is must be restored. This is enough time to upload it async without blocking the whole process.  
I use the rename thing at lot since more than two years now and never experienced that ports were rebuild.

src/vcpkg/commands.format-feature-baseline.cpp
autoantwort: This command was never meant to format `ci.baseline.txt`. It created it to format https://github.com/autoantwort/vcpkg/blob/test-features/scripts/ci.feature.baseline.txt

autoantwort: > This doesn't actually lexically sort port names. I think we can fudge this by just sorting : earlier?

Yeah the new sorting order is better 👍

src/vcpkg/binarycaching.cpp
BillyONeal: >The worst outcome of the "race" is that a package is build twice since it is not uploaded to the binary cache before it is needed a second time.

I'm worried about things like "see that zip exists, try to unzip, but the zip was actually half written so we proceed with something corrupted".

>And the way x-test-features works, at least two other ports will be built before the package with the same ABI is used a second time so that is must be restored.

I'm not willing to ship known races like that though. All it takes is a target that is sufficiently slow and things are toast.

>I think that would be a huge slowdown (at least for me where the binary cache latency is high and with -all).

OK, I'm going to try defending against this better...

BillyONeal: Another example that is no longer safe:

```
    Path make_temp_archive_path(const Path& buildtrees, const PackageSpec& spec)
    {
        return buildtrees / spec.name() / (spec.triplet().to_string() + ".zip");
    }
```

autoantwort: > I'm worried about things like "see that zip exists, try to unzip, but the zip was actually half written so we proceed with something corrupted".

Ok but this is already the status quo since years and can happen if you run two vcpkg instances at the same time or have a shared binary cache on a network drive. 


> All it takes is a target that is sufficiently slow and things are toast.

A slower target will reduce the likelihood, a port that is build fast will increase it. "things are toast" -> Worst thing that can happen is that a port is build again while it was in the binary cache. This does not break anything. 

> Another example that is no longer safe: ...

This is wrong. `make_temp_archive_path` is only called when a package is restored which happens synchronously and is independent from when push_success is done.

BillyONeal:
Tests are likely to fail without https://github.com/microsoft/vcpkg/pull/44217 , but just pushed changes that remove the blocking.

src/vcpkg/binarycaching.cpp
BillyONeal: OK, I've added defenses for this edge case

BillyONeal (approved):
Please take a look at changes I pushed and let me know if they are OK.

The following are needed before this can merge:

* <s>https://github.com/microsoft/vcpkg-tool/pull/1604</s>
* https://github.com/microsoft/vcpkg/pull/44217
* <s>(Billy to resurrect the docs PR tomorrow)</s>

Thanks for the feature and sorry it took us forever to land it!

BillyONeal:
Thanks again for the great, ahem, feature! Sorry it took us forever :(

BillyONeal:
Docs PR resurrected https://github.com/microsoft/vcpkg-docs/pull/455

autoantwort:
It now fails with 
```
Feature Test [18/4632] boost-uninstall[core]:arm64-osx
Installing 1/1 boost-uninstall:arm64-osx@1.87.0...
Building boost-uninstall:arm64-osx@1.87.0...
-- 
Please use the following command when you need to remove all boost ports/components:
    "./vcpkg remove boost-uninstall:arm64-osx --recurse"

Invalid BUILD_INFO file for package: read_contents("/Users/leanderSchulten/git_projekte/vcpkg/packages/boost-uninstall_arm64-osx_202/BUILD_INFO"): No such file or directory
note: updating vcpkg by rerunning bootstrap-vcpkg may resolve this failure.
```
when I run `vcpkg x-test-features --x-abi-tools-use-exact-versions --failure-logs=logs --ci-feature-baseline=scripts/ci.feature.baseline.txt --all`

BillyONeal:
> Invalid BUILD_INFO file for package: read_contents("/Users/leanderSchulten/git_projekte/vcpkg/packages/boost-uninstall_arm64-osx_202/BUILD_INFO"): No such file or directory

You need https://github.com/microsoft/vcpkg/pull/44217

PR #800, opened 2022-11-08
WIP: Hash variables using trace

Neumann-A:
Still WIP put already open for comments. 

Current status: trace output has been parsed and put into a workable struct.

PR #816, opened 2022-11-19
Add triplet variables to control abi entries for binarycache

Osyotr:
`VCPKG_IGNORED_ABI_KEYS`: list of entries to remove from from `vcpkg_abi_info.txt` (`key1;key2;...;keyN`)
`VCPKG_ADDITIONAL_ABI_ENTRIES`: list of additional entries to add to `vcpkg_abi_info.txt` (`key1 value1;key2 value2;...;keyN valueN`)

It's now possible to completely override triplet hash or remove dependencies hashes from abi entries.
⚠Of course, this must be used with caution. ⚠

```cmake
set(VCPKG_TARGET_ARCHITECTURE x64)
set(VCPKG_CRT_LINKAGE dynamic)
set(VCPKG_LIBRARY_LINKAGE dynamic)

# ⚠ This might break things, but we take the risk
set(VCPKG_ADDITIONAL_ABI_ENTRIES "my_custom_triplet_abi 9C6553BC-E762-458E-BDEF-2FEEBD6388BD")
set(VCPKG_IGNORED_ABI_KEYS "triplet_abi" "powershell")

if("${PORT}" STREQUAL "libpng")
    list(APPEND VCPKG_IGNORED_ABI_KEYS "zlib")
endif()
```


bansan85:
This feature is really nice. Tracking powershell version is annoying because Windows Update breaks the whole cache. And adding new entries may be needed (i.e. to track version of external tools).

Will it be merged ?

There is conflicts for the merge action. Files changed are small. I don't have write permission in this MR. If @Osyotr don't have time to fix it, is it fine that I create another MR that can be merged ?

Osyotr:
> Tracking powershell version is annoying because Windows Update breaks the whole cache.

Use `--x-abi-tools-use-exact-versions`.

> And adding new entries may be needed (i.e. to track version of external tools).

This is now possible using `VCPKG_HASH_ADDITIONAL_FILES` albeit a bit clumsy.

> Will it be merged ?

I don't have high hopes for that, considering there hasn't been any comments from maintainers in three years.
Feel free to resubmit, but you will also need to write tests.

bansan85:
I noticed `VCPKG_HASH_ADDITIONAL_FILES`, but I was afraid it would require keeping the file alive in the build tree (or somewhere else).
I just made a test, and it's fine. `vcpkg` adds in `vcpkg_abi_info.txt`:

`additional_file_0 33307da56af13f791584518fef2e49641180bfbf2ef7b2d256c9ab6fad564f80`

and the cache worked as expected.

I will also try `--x-abi-tools-use-exact-versions` the next time PowerShell will be updated.

Thanks.

BillyONeal:
I think this needs:

1. A better description of what you want to do with this. This may be a much more 'general' hammer than we need here.
2. At least 1 end to end test for each newly added knob.
3. Docs, probably that need to include (1) as well as counterexamples.

`VCPKG_IGNORED_ABI_KEYS` I am weakly against at this time as anything it could be used to exclude is likely to break vcpkg contracts with customers, but I could be convinced given a sufficient value for (1) above.

`VCPKG_ADDITIONAL_ABI_ENTRIES` I am weakly against at this time because I do not understand how it differs from just having the content in the triplet file itself. I become weakly in favor if we also do https://github.com/microsoft/vcpkg-tool/pull/596 or https://github.com/microsoft/vcpkg-tool/pull/800 as that will give it an obvious use case.

Osyotr:
> A better description of what you want to do with this

Every touch to the triplet file is a guaranteed world rebuild and I wanted to have control over that by replacing `triplet_abi` with own abi entry that is updated manually. Cheap alternative to https://github.com/microsoft/vcpkg-tool/pull/596 and https://github.com/microsoft/vcpkg-tool/pull/800. 

Also, at that time I did some patching of Qt library so I wanted to speed up testing of the patches by skipping rebuilds of some heavy-ish dependencies like QtWebKit. The workflow was like this:
- Add `qt5-base` to `VCPKG_IGNORED_ABI_KEYS` for `qt5-webkit`.
- Patch `qt5-base`.
- Modify triplet:
    - If I *know* that `qt5-base` abi has changed (e.g. changed some public header):
        - Add or bump version of `qt5-base_ABI` in `VCPKG_ADDITIONAL_ABI_ENTRIES`.
    - If I didn't change anything meaningful:
        - Do nothing.
- Test changes with my app.

---
Nowadays I just don't need touch the triplet very often ([this](https://github.com/Neumann-A/my-vcpkg-triplets/blob/e5f3f2d06430fa475feccbb34a5d61023807b854/x64-win-llvm.cmake#L23) trick together with `VCPKG_HASH_ADDITIONAL_FILES` helps).
And for quick patching I sometimes use leftovers from vcpkg build and just copy binaries manually.

I will close this PR (at least for now).

BillyONeal:
> Also, at that time I did some patching of Qt library so I wanted to speed up testing of the patches by skipping rebuilds of some heavy-ish dependencies like QtWebKit. The workflow was like this:
> 
> * Add `qt5-base` to `VCPKG_IGNORED_ABI_KEYS` for `qt5-webkit`.
> * Patch `qt5-base`.
> * Modify triplet:
>   
>   * If I _know_ that `qt5-base` abi has changed (e.g. changed some public header):
>     
>     * Add or bump version of `qt5-base_ABI` in `VCPKG_ADDITIONAL_ABI_ENTRIES`.
>   * If I didn't change anything meaningful:
>     
>     * Do nothing.
> * Test changes with my app.

That is a *very* good (1) and flips my vote from weakly against to weakly in favor.

PR #839, opened 2022-12-20
Print proper error messages if there is an error.

autoantwort:
Fixes https://github.com/microsoft/vcpkg-tool/pull/710/files#r1051638592

### DIFF
--- a/src/vcpkg/base/checks.cpp
+++ b/src/vcpkg/base/checks.cpp
@@ -105,7 +105,13 @@ namespace vcpkg
     {
         if (!expression)
         {
-            exit_with_message(line_info, error_message);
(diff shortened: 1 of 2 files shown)


vicroms (approved):
I think these changes are OK. We may want to discuss if it is acceptable to leave fatal errors unlocalized.

ras0219-msft:
Your assumption is correct -- however it turns out that this code is incorrectly using `check_exit()`. Since we are now going to use these functions as post-built lint checks, running on user-generated files, they need to handle invalid data without crashing. That means all the functions need to be changed over to `ExpectedL<>`s with localized errors.

That said this is a fine improvement in the meantime. Thanks!

PR #840, opened 2022-12-21
Correct handling of PATH in VCPKG_KEEP_ENV_VARS

ras0219-msft:
Before this PR, using `set VCPKG_KEEP_ENV_VARS=PATH` would result in two copies of `PATH` in the environment block. This had (mostly) the correct results for windows processes which would take the first value, however msys2's bash takes the last value and thus has an incredibly mysterious behavior of apparently resetting the path.

This is a targeted fix for the PATH variable, however a more comprehensive rework of how the environment is handled is needed to ensure correct handling of all variables -- see https://github.com/microsoft/vcpkg-tool/pull/424.

### DIFF
--- a/src/vcpkg/base/system.process.cpp
+++ b/src/vcpkg/base/system.process.cpp
@@ -388,7 +388,16 @@ namespace vcpkg
 
             for (auto&& var : vars)
             {
-                env_strings.push_back(var);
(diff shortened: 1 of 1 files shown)


valeriaconde (approved):
LGTM

PR #843, opened 2022-12-28
Avoid calling memcmp for empty string views

Thomas1664:
This gives a 3.3x performance improvement: https://quick-bench.com/q/RXyLBXqCsm-QHSHGxBDSou6N_Fs

Also gets rid of `/home/vsts/work/1/s/src/vcpkg/base/stringview.cpp:32:50: runtime error: null pointer passed as argument 2, which is declared to never be null`

Code used for benchmarking:

```c++
#include <cstring>

struct SV {
  const char* data = nullptr;
  std::size_t size = 0;

  bool operator==(const SV& other) const noexcept
  {
    return size == other.size && memcmp(data, other.data, size) == 0;
  }
};

struct SV_2 {
  const char* data = nullptr;
  std::size_t size = 0;

  bool operator==(const SV_2& other) const noexcept
  {
    if (size == 0 && other.size == 0) return true;
    return size == other.size && memcmp(data, other.data, size) == 0;
  }
};


static void StringCreation(benchmark::State& state) {
  for (auto _ : state) {
    SV sv {};
    benchmark::DoNotOptimize(sv);
    SV sv_2{};
    benchmark::DoNotOptimize(sv_2);
    bool res = sv == sv_2;
    benchmark::DoNotOptimize(res);
  }
}
// Register the function as a benchmark
BENCHMARK(StringCreation);

static void StringCopy(benchmark::State& state) {
  for (auto _ : state) {
    SV_2 sv {};
    benchmark::DoNotOptimize(sv);
    SV_2 sv_2{};
    benchmark::DoNotOptimize(sv_2);
    bool res = sv == sv_2;
    benchmark::DoNotOptimize(res);
  }
}
BENCHMARK(StringCopy);
```

Settings: GCC 9.4, c++17, O3

### DIFF
--- a/src/vcpkg/base/stringview.cpp
+++ b/src/vcpkg/base/stringview.cpp
@@ -29,6 +29,10 @@ namespace vcpkg
 
     bool operator==(StringView lhs, StringView rhs) noexcept
     {
+        if (lhs.empty() && rhs.empty())
+        {
+            return false;
+        }
         return lhs.size() == rhs.size() && memcmp(lhs.data(), rhs.data(), lhs.size()) == 0;
     }
 
--- a/src/vcpkg/base/stringview.cpp
+++ b/src/vcpkg/base/stringview.cpp
@@ -31,7 +31,7 @@ namespace vcpkg
     {
         if (lhs.empty() && rhs.empty())
         {
-            return false;
+            return true;
         }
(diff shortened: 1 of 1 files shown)


ras0219-msft (approved):
Thanks for the PR!

The primary motivation to merge this PR is not performance -- I sincerely doubt you can measure any difference in performance for any vcpkg product scenario with this change. However, passing `nullptr` to `memcmp` is illegal and definitely something we need to avoid.

PR #850, opened 2023-01-10
Fix unescaped curly braces in localized string

vicroms:
Fixes https://github.com/microsoft/vcpkg/issues/28845


BillyONeal (approved):

locales/messages.json
```diff
@@ -946,8 +946,8 @@
   "PortBugMissingImportedLibs": "Import libraries were not present in {path}.\nIf this is intended, add the following line in the portfile:\nset(VCPKG_POLICY_DLLS_WITHOUT_LIBS enabled)",
   "_PortBugMissingImportedLibs.comment": "An example of {path} is /foo/bar.",
   "PortBugMissingIncludeDir": "The folder /include is empty or not present. This indicates the library was not correctly installed.",
-  "PortBugMissingLicense": "The software license must be available at ${CURRENT_PACKAGES_DIR}/share/{spec}/copyright",
-  "_PortBugMissingLicense.comment": "'{CURRENT_PACKAGES_DIR}' should not be translated. An example of {spec} is zlib:x64-windows.",
+  "PortBugMissingLicense": "The software license must be available at ${{CURRENT_PACKAGES_DIR}}/share/{spec}/copyright",
+  "_PortBugMissingLicense.comment": "An example of {spec} is zlib:x64-windows.",
```
ras0219-msft, line 950: This comment indicates that `msg::spec` isn't the best argument placeholder to use -- we probably should use `msg::package_name` instead.

vicroms: Noted, will add it as a drive-by fix in my export-port PR.

dg0yt: `${{PORT}}` instead of spec?

PR #852, opened 2023-01-11
Formatted with clang-format@15

autoantwort:


autoantwort:
It is even better when the version changes randomly between every ci run </s>

autoantwort:
Already fixed in main.

PR #862, opened 2023-01-17
fix issue#28700; parens need to be shell-escaped on Linux

markle11m:


vicroms (approved):

BillyONeal (changes requested):
This change seems insufficient:

image

I think you need to add handling of `$` and \` (backtick) on line 86?

BillyONeal:
image

BillyONeal (approved):
LGTM if PR Build comes back OK

PR #873, opened 2023-01-29
vcpkg upgrade: Handle the removal of features in new port versions

autoantwort:
Currently the command simply fails to execute.


ras0219-msft (approved):
This looks good to me, thanks for the improvement!

PR #877, opened 2023-01-31
Post build check for usage

cnSchwarzer:


cnSchwarzer:
What am I missing?

src/vcpkg/postbuildlint.cpp
```diff
@@ -240,6 +240,28 @@ namespace vcpkg::PostBuildLint
...
 
+    static LintStatus check_for_usage_forgot_install(const Filesystem& fs,
+                                                     const VcpkgPaths& paths,
+                                                     const Path& package_dir,
+                                                     const PackageSpec& spec)
+    {
+        static const std::string STANDARD_INSTALL_USAGE =
+            R"###(file(INSTALL "${CMAKE_CURRENT_LIST_DIR}/usage" DESTINATION "${CURRENT_PACKAGES_DIR}/share/${PORT}"))###";
+
+        auto port_path = paths.root / "ports" / spec.name();
```
autoantwort, line 251: That is not true for overlay ports and custom registries or overwritten versions.

cnSchwarzer:
Seems github changed SHA512 back... reverting last commit

src/vcpkg/postbuildlint.cpp
cnSchwarzer: Is there any exist way to get the port folder?

autoantwort: You can pass this variable further down:
https://github.com/microsoft/vcpkg-tool/blob/f39cfece245139aed1dfb4d7d71b9adf61c0018a/src/vcpkg/postbuildlint.cpp#L1210

cnSchwarzer: Got it

src/vcpkg/postbuildlint.cpp
```diff
@@ -240,6 +240,28 @@ namespace vcpkg::PostBuildLint
...
+        auto usage_path_from = port_path / "usage";
```
autoantwort, line 252: And I don't know if we want to handle these cases:
image
So maybe check if there is a file in the port dir that contains usage and don't end with `.patch`

autoantwort: But otherwise the usage is maybe only for msvc and should not be installed for linux and osx. Then this post build check would be wrong.

cnSchwarzer: True, maybe we should fatal to 'usage' and warning non-stop to other files contains 'usage'?

include/vcpkg/base/messages.h
```diff
@@ -2122,6 +2122,12 @@ namespace vcpkg
                     (msg::path),
                     "",
                     "The /{path} file does not exist. This file must exist for CMake helper ports.");
+    DECLARE_MESSAGE(PortBugMissingProvidedUsage,
+                    (msg::spec, msg::value),
+                    "",
+                    "The port provided \"usage\" but forgot to install to\n"
+                    "${{CURRENT_PACKAGES_DIR}}/share/{spec}/usage\n"
```
Thomas1664, line 2129: Don't add the path and \n here. Please handle this in the calling code using `LocalizedString::append_raw()`

src/vcpkg/postbuildlint.cpp
```diff
@@ -240,6 +240,27 @@ namespace vcpkg::PostBuildLint
         return LintStatus::SUCCESS;
     }
 
+    static LintStatus check_for_usage_forgot_install(const Filesystem& fs,
+                                                     const Path& port_dir,
+                                                     const Path& package_dir,
+                                                     const PackageSpec& spec)
+    {
+        static const std::string STANDARD_INSTALL_USAGE =
```
Thomas1664, line 248:
```suggestion
        static constexpr StringLiteral =
```

src/vcpkg/postbuildlint.cpp
```diff
@@ -240,6 +240,27 @@ namespace vcpkg::PostBuildLint
...
+            R"###(file(INSTALL "${CMAKE_CURRENT_LIST_DIR}/usage" DESTINATION "${CURRENT_PACKAGES_DIR}/share/${PORT}"))###";
+
+        auto usage_path_from = port_dir / "usage";
+        auto usage_path_to = package_dir / "share" / spec.name() / "usage";
+
+        if (fs.exists(usage_path_from, IgnoreErrors{}) && !fs.exists(usage_path_to, IgnoreErrors{}))
```
Thomas1664, line 254:
```suggestion
        if (fs.is_regular_file(usage_path_from) && !fs.is_regular_file(usage_path_to))
```
Please check if the file is a regular file. We had issues with copyright files where people abused this to create copyright folders.
Not sure if `is_regular_file` is the correct name but we have a function for this use case that does what the name suggests.

include/vcpkg/base/messages.h
cnSchwarzer: Does the latest commit match your meaning?

Thomas1664: > Does the latest commit match your meaning?

better, but you just removed `${{CURRENT_PACKAGES_DIR}}` from the path. Please set the entire path in the calling code.

BillyONeal:
Thanks for the new and useful lint check!

PR #883, opened 2023-02-05
Add xbox value to supports expressions

walbourn:
Per the work on community triplets for xbox, this adds "xbox" as a known variable to the supports expressions.

### DIFF
--- a/src/vcpkg/platform-expression.cpp
+++ b/src/vcpkg/platform-expression.cpp
@@ -28,6 +28,7 @@ namespace vcpkg::PlatformExpression
         openbsd,
         osx,
         uwp,
+        xbox,
         android,
         emscripten,
         ios,
@@ -54,6 +55,7 @@ namespace vcpkg::PlatformExpression
             {"openbsd", Identifier::openbsd},
             {"osx", Identifier::osx},
             {"uwp", Identifier::uwp},
+            {"xbox", Identifier::xbox},
             {"android", Identifier::android},
             {"emscripten", Identifier::emscripten},
             {"ios", Identifier::ios},
@@ -532,6 +534,7 @@ namespace vcpkg::PlatformExpression
                         case Identifier::arm64: return true_if_exists_and_equal("VCPKG_TARGET_ARCHITECTURE", "arm64");
                         case Identifier::windows:
                             return true_if_exists_and_equal("VCPKG_CMAKE_SYSTEM_NAME", "") ||
+                                   true_if_exists_and_equal("VCPKG_CMAKE_SYSTEM_NAME", "Windows") ||
                                    true_if_exists_and_equal("VCPKG_CMAKE_SYSTEM_NAME", "WindowsStore") ||
                                    true_if_exists_and_equal("VCPKG_CMAKE_SYSTEM_NAME", "MinGW");
                         case Identifier::mingw: return true_if_exists_and_equal("VCPKG_CMAKE_SYSTEM_NAME", "MinGW");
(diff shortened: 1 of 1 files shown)


walbourn:
Related to [this PR](https://github.com/microsoft/vcpkg/pull/29210) for new community triplets.

src/vcpkg/platform-expression.cpp
```diff
@@ -532,6 +534,7 @@ namespace vcpkg::PlatformExpression
                         case Identifier::arm64: return true_if_exists_and_equal("VCPKG_TARGET_ARCHITECTURE", "arm64");
                         case Identifier::windows:
                             return true_if_exists_and_equal("VCPKG_CMAKE_SYSTEM_NAME", "") ||
+                                   true_if_exists_and_equal("VCPKG_CMAKE_SYSTEM_NAME", "Windows") ||
```
walbourn, line 537: In the current implementation if you explicitly set ``VCPKG_CMAKE_SYSTEM_NAME`` to "Windows", it wouldn't return true for the 'windows' expression.

Was this intentional or just an assumption that all Windows system usage would be "default"?

src/vcpkg/platform-expression.cpp
```diff
@@ -541,6 +544,8 @@ namespace vcpkg::PlatformExpression
                         case Identifier::osx: return true_if_exists_and_equal("VCPKG_CMAKE_SYSTEM_NAME", "Darwin");
                         case Identifier::uwp:
                             return true_if_exists_and_equal("VCPKG_CMAKE_SYSTEM_NAME", "WindowsStore");
+                        case Identifier::xbox:
+                            return true_if_exists_and_equal("Z_VCPKG_IS_XBOX", "true");
```
walbourn, line 548: We do not have a specific CMAKE_SYSTEM_NAME for xbox (it's just a variant of "Windows"). As such, there's no clear way to have the expression evaluator know from the existing VCPKG port variables that it should return "1" for the new ``xbox`` variable.

My proposal here is to have my triplets add:

```
set(Z_VCPKG_IS_XBOX "true")
```

```

**Comment by walbourn** [comment]
My initial "seed" list of port updates for the new supports expressions would be:

Add ``| xbox`` to the following for documentation purposes: directxmath, directxtk12, directxtex, directxmesh, uvatlas 

Add `` & !xbox`` to the following:  directx-dxc, directx-headers, directxsdk, dstorage, d3dx12, xaudio2redist, directxtk, dxsdk-d3dx, dxut, effects11, opengl, opengl-registry, angle, ms-angle, mesa, freeglut

**Comment by dg0yt** [comment]
Alternative to extending this C++ function, there is another extension point in triplet files:
https://github.com/microsoft/vcpkg/blob/master/docs/users/triplets.md#vcpkg_dep_info_override_vars

There is probably no example of how to use it ATM, but I guess the syntax is like:
~~~
set(VCPKG_DEP_INFO_OVERRIDE_VARS xbox !ybox)
~~~

**Comment by autoantwort** [comment]
image
It seems to be  

```cmake
set(VCPKG_DEP_INFO_OVERRIDE_VARS xbox)
```

walbourn:
Awesome. This works!

walbourn:
Well, I still have to add 'xbox' as a known identifier to avoid getting:

```
error: Unrecognized identifer name xbox. Add to override list in triplet file.
```

Anytime I use a triplet other than mine.

walbourn:
Looks like I need to create a fresh PR since I got a little too fast deleting my fork.

walbourn:
Replaced by https://github.com/microsoft/vcpkg-tool/pull/884/files

PR #906, opened 2023-02-15
post build check: Always allow .pc files in lib/pkgconfig and debug/lib/pkgconfig

autoantwort:
Fixes https://github.com/microsoft/vcpkg/issues/29661


Neumann-A:
Maybe emit a msg that it could be moved to share/pkgconfig ?

dg0yt:
> Maybe emit a msg that it could be moved to share/pkgconfig ?

Better not. Safe the noise when the file is in a valid location.

Neumann-A:
> Better not. Safe the noise when the file is in a valid location.

Maybe only in `--debug` then ?

BillyONeal:
> > Better not. Safe the noise when the file is in a valid location.
> 
> Maybe only in `--debug` then ?

I don't think there's much value in debug only post build checks. Particularly when you can just find them with grep et al. later if you want.

I'm going to merge this because it fixes tool release blockers. Thanks!

Neumann-A:
> Particularly when you can just find them with grep et al. later if you want.

I cannot grep through files I dont have, however I can make CI run in debug mode ;)

BillyONeal:
>I cannot grep through files I dont have, however I can make CI run in debug mode ;)

We make a txt file with all the files installed for you.
________________________________

PR #908, opened 2023-02-15
Binary cache: async push_success

autoantwort:
This results in ~10-20% faster build times on my machine.

### DIFF
--- a/include/vcpkg/binarycaching.h
+++ b/include/vcpkg/binarycaching.h
@@ -9,10 +9,14 @@
 #include <vcpkg/base/files.h>
 
 #include <vcpkg/packagespec.h>
+#include <vcpkg/sourceparagraph.h>
 
+#include <condition_variable>
 #include <iterator>
+#include <queue>
 #include <set>
 #include <string>
+#include <thread>
 #include <unordered_map>
 #include <vector>
 
@@ -42,6 +46,21 @@ namespace vcpkg
         const IBinaryProvider* m_available_provider = nullptr; // meaningful iff m_status == available
     };
 
+    struct BinaryPackageInformation
+    {
+        explicit BinaryPackageInformation(const InstallPlanAction& action);
+        std::string package_abi;
+        // The following fields are only needed for the nuget binary cache
+        PackageSpec spec;
+        SourceControlFile& source_control_file;
+        std::string& raw_version;
+        std::string compiler_id;
+        std::string compiler_version;
+        std::string triplet_abi;
+        InternalFeatureSet feature_list;
+        std::vector<PackageSpec> package_dependencies;
+    };
+
     struct IBinaryProvider
     {
         virtual ~IBinaryProvider() = default;
@@ -52,7 +71,9 @@ namespace vcpkg
 
         /// Called upon a successful build of `action` to store those contents in the binary cache.
         /// Prerequisite: action has a package_abi()
-        virtual void push_success(const InstallPlanAction& action) const = 0;
+        virtual void push_success(const BinaryPackageInformation& info,
+                                  const Path& packages_dir,
+                                  MessageSink& msg_sink) = 0;
 
         /// Gives the IBinaryProvider an opportunity to batch any downloading or server communication for
         /// executing `actions`.
@@ -77,7 +98,7 @@ namespace vcpkg
         std::vector<std::string> headers_for_get;
 
         LocalizedString valid() const;
-        std::string instantiate_variables(const InstallPlanAction& action) const;
+        std::string instantiate_variables(const BinaryPackageInformation& info) const;
     };
 
     struct BinaryConfigParserState
@@ -121,8 +142,12 @@ namespace vcpkg
 
     struct BinaryCache
     {
-        BinaryCache() = default;
+        static BinaryCache* current_instance;
+        static void wait_for_async_complete();
+        BinaryCache(Filesystem& filesystem);
         explicit BinaryCache(const VcpkgCmdArguments& args, const VcpkgPaths& paths);
+
+        ~BinaryCache();
 
         void install_providers(std::vector<std::unique_ptr<IBinaryProvider>>&& providers);
         void install_providers_for(const VcpkgCmdArguments& args, const VcpkgPaths& paths);
@@ -131,7 +156,7 @@ namespace vcpkg
         RestoreResult try_restore(const InstallPlanAction& action);
 
         /// Called upon a successful build of `action` to store those contents in the binary cache.
-        void push_success(const InstallPlanAction& action);
+        void push_success(const InstallPlanAction& action, Path package_dir);
 
         /// Gives the IBinaryProvider an opportunity to batch any downloading or server communication for
         /// executing `actions`.
@@ -143,8 +168,22 @@ namespace vcpkg
         std::vector<CacheAvailability> precheck(View<InstallPlanAction> actions);
 
     private:
+        struct ActionToPush
+        {
+            BinaryPackageInformation info;
+            bool clean_after_push;
+            Path packages_dir;
+        };
+        void push_thread_main();
+
         std::unordered_map<std::string, CacheStatus> m_status;
         std::vector<std::unique_ptr<IBinaryProvider>> m_providers;
+        std::condition_variable actions_to_push_notifier;
+        std::mutex actions_to_push_mutex;
+        std::queue<ActionToPush> actions_to_push;
+        std::thread push_thread;
+        std::atomic_bool end_push_thread;
+        Filesystem& filesystem;
     };
 
     ExpectedS<DownloadManagerConfig> parse_download_configuration(const Optional<std::string>& arg);
--- a/src/vcpkg.cpp
+++ b/src/vcpkg.cpp
@@ -9,6 +9,7 @@
 #include <vcpkg/base/system.debug.h>
 #include <vcpkg/base/system.process.h>
 
+#include <vcpkg/binarycaching.h>
 #include <vcpkg/cgroup-parser.h>
 #include <vcpkg/commands.contact.h>
 #include <vcpkg/commands.h>
@@ -154,6 +155,7 @@ namespace vcpkg::Checks
     // Implements link seam from basic_checks.h
     void on_final_cleanup_and_exit()
     {
+        BinaryCache::wait_for_async_complete();
         const auto elapsed_us_inner = g_total_time.microseconds();
         bool debugging = Debug::g_debugging;
 
--- a/src/vcpkg/binarycaching.cpp
+++ b/src/vcpkg/binarycaching.cpp
@@ -297,45 +297,41 @@ namespace
             return RestoreResult::unavailable;
         }
 
-        void push_success(const InstallPlanAction& action) const override
+        void push_success(const BinaryPackageInformation& info,
+                          const Path& packages_dir,
+                          MessageSink& msg_sink) override
         {
             if (m_write_dirs.empty() && m_put_url_templates.empty())
             {
                 return;
             }
 
-            const auto& abi_tag = action.package_abi().value_or_exit(VCPKG_LINE_INFO);
-            auto& spec = action.spec;
+            const auto& abi_tag = info.package_abi;
+            auto& spec = info.spec;
             auto& fs = paths.get_filesystem();
             const auto archive_subpath = make_archive_subpath(abi_tag);
             const auto tmp_archive_path = make_temp_archive_path(paths.buildtrees(), spec);
-            auto compress_result = compress_directory_to_zip(
-                fs, paths.get_tool_cache(), stdout_sink, paths.package_dir(spec), tmp_archive_path);
+            auto compress_result =
+                compress_directory_to_zip(fs, paths.get_tool_cache(), stdout_sink, packages_dir, tmp_archive_path);
             if (!compress_result)
             {
-                msg::println(Color::warning,
-                             msg::format_warning(msgCompressFolderFailed, msg::path = paths.package_dir(spec))
-                                 .append_raw(' ')
-                                 .append_raw(compress_result.error()));
+                msg_sink.println(Color::warning,
+                                 msg::format_warning(msgCompressFolderFailed, msg::path = packages_dir)
+                                     .append_raw(' ')
+                                     .append_raw(compress_result.error()));
                 return;
             }
-            size_t http_remotes_pushed = 0;
             for (auto&& put_url_template : m_put_url_templates)
             {
-                auto url = put_url_template.instantiate_variables(action);
+                auto url = put_url_template.instantiate_variables(info);
                 auto maybe_success = put_file(fs, url, m_secrets, put_url_template.headers_for_put, tmp_archive_path);
                 if (maybe_success)
                 {
-                    http_remotes_pushed++;
+                    m_http_remotes_pushed++;
                     continue;
                 }
 
-                msg::println(Color::warning, maybe_success.error());
-            }
-
-            if (!m_put_url_templates.empty())
-            {
-                msg::println(msgUploadedBinaries, msg::count = http_remotes_pushed, msg::vendor = "HTTP remotes");
+                msg_sink.println(Color::warning, maybe_success.error());
             }
 
             for (const auto& archives_root_dir : m_write_dirs)
@@ -354,14 +350,10 @@ namespace
 
                 if (ec)
                 {
-                    msg::println(Color::warning,
-                                 msg::format(msgFailedToStoreBinaryCache, msg::path = archive_path)
-                                     .append_raw('\n')
-                                     .append_raw(ec.message()));
-                }
-                else
-                {
-                    msg::println(msgStoredBinaryCache, msg::path = archive_path);
+                    msg_sink.println(Color::warning,
+                                     msg::format(msgFailedToStoreBinaryCache, msg::path = archive_path)
+                                         .append_raw('\n')
+                                         .append_raw(ec.message()));
                 }
             }
             // In the case of 1 write dir, the file will be moved instead of copied
@@ -369,7 +361,19 @@ namespace
             {
                 fs.remove(tmp_archive_path, IgnoreErrors{});
             }
+            if (!m_put_url_templates.empty())
+            {
+                msg_sink.println(msgUploadedBinaries, msg::count = m_http_remotes_pushed, msg::vendor = "HTTP remotes");
+            }
         }
+
+        /*void print_upload_statistics(MessageSink& msg_sink) override
+        {
+            if (!m_put_url_templates.empty())
+            {
+                msg_sink.println(msgUploadedBinaries, msg::count = m_http_remotes_pushed, msg::vendor = "HTTP remotes");
+            }
+        };*/
 
         void precheck(View<InstallPlanAction> actions, View<CacheStatus*> cache_status) const override
         {
@@ -411,6 +415,7 @@ namespace
         std::vector<Path> m_write_dirs;
         std::vector<UrlTemplate> m_put_url_templates;
         std::vector<std::string> m_secrets;
+        size_t m_http_remotes_pushed = 0;
     };
     struct HttpGetBinaryProvider : IBinaryProvider
     {
@@ -423,7 +428,7 @@ namespace
 
         RestoreResult try_restore(const InstallPlanAction&) const override { return RestoreResult::unavailable; }
 
-        void push_success(const InstallPlanAction&) const override { }
+        void push_success(const BinaryPackageInformation&, const Path&, MessageSink&) override { }
 
         void prefetch(View<InstallPlanAction> actions, View<CacheStatus*> cache_status) const override
         {
@@ -446,7 +451,7 @@ namespace
 
                     auto&& action = actions[idx];
                     clean_prepare_dir(fs, paths.package_dir(action.spec));
-                    auto uri = url_template.instantiate_variables(action);
+                    auto uri = url_template.instantiate_variables(BinaryPackageInformation{action});
                     url_paths.emplace_back(std::move(uri), make_temp_archive_path(paths.buildtrees(), action.spec));
                     url_indices.push_back(idx);
                 }
@@ -510,7 +515,7 @@ namespace
                         continue;
                     }
 
-                    urls.push_back(url_template.instantiate_variables(actions[idx]));
+                    urls.push_back(url_template.instantiate_variables(BinaryPackageInformation{actions[idx]}));
                     url_indices.push_back(idx);
                 }
 
@@ -808,20 +813,21 @@ namespace
 
         RestoreResult try_restore(const InstallPlanAction&) const override { return RestoreResult::unavailable; }
 
-        void push_success(const InstallPlanAction& action) const override
+        void push_success(const BinaryPackageInformation& info,
+                          const Path& packages_dir,
+                          MessageSink& msg_sink) override
         {
             if (m_write_sources.empty() && m_write_configs.empty())
             {
                 return;
             }
 
-            auto& spec = action.spec;
+            auto& spec = info.spec;
 
-            NugetReference nuget_ref = make_nugetref(action, get_nuget_prefix());
+            NugetReference nuget_ref = make_nugetref(info, get_nuget_prefix());
             auto nuspec_path = paths.buildtrees() / spec.name() / (spec.triplet().to_string() + ".nuspec");
             auto& fs = paths.get_filesystem();
-            fs.write_contents(
-                nuspec_path, generate_nuspec(paths.package_dir(spec), action, nuget_ref), VCPKG_LINE_INFO);
+            fs.write_contents(nuspec_path, generate_nuspec(packages_dir, info, nuget_ref), VCPKG_LINE_INFO);
 
             const auto& nuget_exe = paths.get_tool_exe("nuget", stdout_sink);
             Command cmdline;
@@ -843,7 +849,7 @@ namespace
 
             if (!run_nuget_commandline(cmdline))
             {
-                msg::println(Color::error, msgPackingVendorFailed, msg::vendor = "NuGet");
+                msg_sink.println(Color::error, msgPackingVendorFailed, msg::vendor = "NuGet");
                 return;
             }
 
@@ -867,11 +873,12 @@ namespace
                 {
                     cmd.string_arg("-NonInteractive");
                 }
-                msg::println(
+                msg_sink.println(
                     msgUploadingBinariesToVendor, msg::spec = spec, msg::vendor = "NuGet", msg::path = write_src);
                 if (!run_nuget_commandline(cmd))
                 {
-                    msg::println(Color::error, msgPushingVendorFailed, msg::vendor = "NuGet", msg::path = write_src);
+                    msg_sink.println(
+                        Color::error, msgPushingVendorFailed, msg::vendor = "NuGet", msg::path = write_src);
                 }
             }
             for (auto&& write_cfg : m_write_configs)
@@ -892,14 +899,15 @@ namespace
                 {
                     cmd.string_arg("-NonInteractive");
                 }
-                msg::println(Color::error,
-                             msgUploadingBinariesUsingVendor,
-                             msg::spec = spec,
-                             msg::vendor = "NuGet config",
-                             msg::path = write_cfg);
+                msg_sink.println(Color::error,
+                                 msgUploadingBinariesUsingVendor,
+                                 msg::spec = spec,
+                                 msg::vendor = "NuGet config",
+                                 msg::path = write_cfg);
                 if (!run_nuget_commandline(cmd))
                 {
-                    msg::println(Color::error, msgPushingVendorFailed, msg::vendor = "NuGet", msg::path = write_cfg);
+                    msg_sink.println(
+                        Color::error, msgPushingVendorFailed, msg::vendor = "NuGet", msg::path = write_cfg);
                 }
             }
 
@@ -1007,21 +1015,23 @@ namespace
 
         RestoreResult try_restore(const InstallPlanAction&) const override { return RestoreResult::unavailable; }
 
-        void push_success(const InstallPlanAction& action) const override
+        void push_success(const BinaryPackageInformation& info,
+                          const Path& packages_dir,
+                          MessageSink& msg_sink) override
         {
             if (m_write_prefixes.empty()) return;
             const ElapsedTimer timer;
-            const auto& abi = action.package_abi().value_or_exit(VCPKG_LINE_INFO);
-            auto& spec = action.spec;
+            const auto& abi = info.package_abi;
+            auto& spec = info.spec;
             const auto tmp_archive_path = make_temp_archive_path(paths.buildtrees(), spec);
             auto compression_result = compress_directory_to_zip(
-                paths.get_filesystem(), paths.get_tool_cache(), stdout_sink, paths.package_dir(spec), tmp_archive_path);
+                paths.get_filesystem(), paths.get_tool_cache(), stdout_sink, packages_dir, tmp_archive_path);
             if (!compression_result)
             {
-                vcpkg::msg::println(Color::warning,
-                                    msg::format_warning(msgCompressFolderFailed, msg::path = paths.package_dir(spec))
-                                        .append_raw(' ')
-                                        .append_raw(compression_result.error()));
+                msg_sink.println(Color::warning,
+                                 msg::format_warning(msgCompressFolderFailed, msg::path = packages_dir)
+                                     .append_raw(' ')
+                                     .append_raw(compression_result.error()));
                 return;
             }
 
@@ -1034,10 +1044,10 @@ namespace
                 }
             }
 
-            msg::println(msgUploadedPackagesToVendor,
-                         msg::count = upload_count,
-                         msg::elapsed = timer.elapsed(),
-                         msg::vendor = vendor());
+            msg_sink.println(msgUploadedPackagesToVendor,
+                             msg::count = upload_count,
+                             msg::elapsed = timer.elapsed(),
+                             msg::vendor = vendor());
         }
 
         void precheck(View<InstallPlanAction> actions, View<CacheStatus*> cache_status) const override
@@ -1273,26 +1283,25 @@ namespace vcpkg
         return {};
     }
 
-    std::string UrlTemplate::instantiate_variables(const InstallPlanAction& action) const
+    std::string UrlTemplate::instantiate_variables(const BinaryPackageInformation& info) const
     {
         return api_stable_format(url_template,
                                  [&](std::string& out, StringView key) {
                                      if (key == "version")
                                      {
-                                         out += action.source_control_file_and_location.value_or_exit(VCPKG_LINE_INFO)
-                                                    .source_control_file->core_paragraph->raw_version;
+                                         out += info.raw_version;
                                      }
                                      else if (key == "name")
                                      {
-                                         out += action.spec.name();
+                                         out += info.spec.name();
                                      }
                                      else if (key == "triplet")
                                      {
-                                         out += action.spec.triplet().canonical_name();
+                                         out += info.spec.triplet().canonical_name();
                                      }
                                      else if (key == "sha")
                                      {
-                                         out += action.abi_info.value_or_exit(VCPKG_LINE_INFO).package_abi;
+                                         out += info.package_abi;
                                      }
                                      else
                                      {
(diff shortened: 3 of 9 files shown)


autoantwort:
How or when should "upload messages" (like `Uploaded binaries to {count} {vendor}.`) be printed?

Thomas1664:
Doesn't this have the same problem as #694 that the working thread might exit due to calls to `check_exit` or `value_or_exit`?

src/vcpkg/binarycaching.cpp
```diff
@@ -1455,6 +1496,34 @@ namespace vcpkg
...
+            actions_to_push_notifier.wait(lock, [this]() { return !actions_to_push.empty() || end_push_thread; });
+            if (actions_to_push.empty())
+            {
+                if (end_push_thread) break;
+                continue;
+            }
+
+            auto entry = std::move(actions_to_push.front());
+            actions_to_push.pop();
+            lock.unlock(); // we don't touch actions_to_push anymore
```
Thomas1664, line 1513: Use a scope to release the lock?

autoantwort:
> Doesn't this have the same problem as #694 that the working thread might exit due to calls to `check_exit` or `value_or_exit`?

Kind of. In general we need an option to decide if a binary cache failure should be a hard error or only a warning

src/vcpkg/binarycaching.cpp
autoantwort: But the entry is not copyable and I would have to declare it outside of the scope and assign it inside of the scope

Thomas1664:
> Kind of. In general we need an option to decide if a binary cache failure should be a hard error or only a warning

The problem is that we almost never can be sure that there isn't some nested API call that exits on failure. But it seems like #909 at least partially addresses this issue.

autoantwort:
Yeah but in the binary cache are nearly no hard exists. It currently also only prints warnings.

src/vcpkg/binarycaching.cpp
```diff
@@ -1449,6 +1490,34 @@ namespace vcpkg
...
+
+            for (auto&& provider : m_providers)
+            {
+                provider->push_success(entry.info, entry.packages_dir, stdout_sink);
```
ras0219-msft, line 1511: We can't use stdout_sink for this -- it would mean tearing between threads. This requires a novel `MessageSink` that's a SPSC queue of lines. Roughly:

```cpp
struct BGMessageSink : MessageSink
{
    std::mutex m_lock;
    // guarded by m_lock
    std::vector<std::pair<Color, string>> m_published;
    // unguarded, buffers messages until newline is reached
    std::vector<std::pair<Color, string>> m_unpublished;

    // must be called from producer
    void print(Color c, string s) override
    {
        auto pos = s.find_last('\n');
        if (pos != string::npos)
        {
            std::lock_guard<std::mutex> lk(m_lock);
            m_published.insert(m_published.end(),
                               std::make_move_iterator(m_unpublished.begin()),
                               std::make_move_iterator(m_unpublished.end()));
            m_published.emplace_back(c, s.substr(0, pos + 1));
        }
        m_unpublished.clear();
        if (s.size() > pos + 1)
        {
            m_unpublished.emplace_back(c, s.substr(pos + 1));
        }
    }

    // must be called from consumer (synchronizer of out)
    void print_published(MessageSink& out)
    {
        std::lock_guard<std::mutex> lk(m_lock);
        for (auto&& m : m_published)
        {
            out.print(m.first, m.second);
        }
        m_published.clear();
    }

    // must be called from producer / after producer finishes
    void flushall()
    {
        std::lock_guard<std::mutex> lk(m_lock);
        m_published.insert(m_published.end(),
                           std::make_move_iterator(m_unpublished.begin()),
                           std::make_move_iterator(m_unpublished.end()));
    }
};
```

Then, the outer owner of binary caching can call `.print_published(stdout_sink)` between packages to emit every complete line produced by the binary caching machinery. This at least limits tearing to line-based.

It would also be reasonable to not flush on lines and instead require manual flushing from the consumer thread; this would entirely prevent tearing and ensure we emit all the messages associated with a specific binary upload all at once.

Before destroying this sink, the owner should call `.flushall()` and `.print_published(stdout_sink)` a final time to make sure everything was properly printed.

There's also a _slightly_ higher perf option since this is a SPSC queue -- there could be a third `m_post_published` vector which is swapped with `m_published` under the lock. This would get the actual prints out from under the lock which means any blocking on stdout would not prevent more binary packages from being uploaded in the background. The `swap` would ensure that at least the vector's memory gets reused without constant reallocation -- though since most of the allocations here are the individual strings I'm not sure it makes a difference.

ras0219-msft (changes requested):
I like this direction; unblocking I/O work has great potential for making vcpkg much faster.

However we need to be very careful about the impacts of concurrency -- deadlocks suck :(

src/vcpkg.cpp
```diff
@@ -156,6 +157,7 @@ namespace vcpkg::Checks
     // Implements link seam from basic_checks.h
     void on_final_cleanup_and_exit()
     {
+        BinaryCache::wait_for_async_complete();
```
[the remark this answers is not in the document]
autoantwort: I have changed the `BinaryCache::wait_for_async_complete()` implementation so it does not deadlock anymore. 

I also moved the call to `Checks::exit_with_code` which is not called when crtl+c is handled. (I personally would like to have a way to terminate vcpkg but wait until the binary cache is done so that I don't lose progress.) 

And I prefer it when build packages are uploaded to the binary caches before vcpkg exits because of an error, otherwise I have to build the already build packages again at a later point when there is no cache entry.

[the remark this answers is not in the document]
ras0219-msft: Agreed that it is, desirable to finish uploading on "understood" errors. For example, if a package failed to build or failed to be installed.

I was also wrong about my original assessment of a deadlock. My concern was the call path of the binary upload thread calling `Checks::unreachable()` or `.value_or_exit()`, but it seems that `std::thread::join()` does have a carve-out to handle this specific case: it will throw a `resource_deadlock_would_occur` if you try to join yourself.

I've put some other concerns below, but I don't want those to distract from my main point: I think the best approach for vcpkg right now is to add calls from `Install::perform()` etc to `BinaryCache::wait_for_async_complete()` before any "user-facing" error, such as the exit guarded by `result.code != BuildResult::SUCCEEDED && keep_going == KeepGoing::NO`. This is motivated by the perspective that it's always safer to terminate than to join and possibly deadlock / race condition / etc. We must make it as trivial / correct-by-construction as possible to guarantee that the binary cache thread NEVER attempts to wait on itself.

---

There's still a UB data race if the main thread and binary upload thread attempt to exit at the same time:
> Concurrently calling join() on the same thread object from multiple threads constitutes a data race that results in undefined behavior.
> -- https://en.cppreference.com/w/cpp/thread/thread/join

There's also a serious "scalability" problem if we ever want a second background thread for whatever reason, because BGThread A would join on BGThread B, while BGThread B tries to join on BGThread A. This might be solvable with ever more complex structures, such as a thread ownership DAG that gets threads to join only on their direct children, but I don't think the benefit is worth the cost.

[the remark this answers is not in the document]
autoantwort: The UB and the joining itself could simply be prevented by doing a `if (std::this_thread::get_id() == instance->push_thread.get_id())`. My concern with the explicit approach is that it is easy to forget to call the waiting function of the BinaryCache and every time you want to exit you have to remember to call it. This seems to me to be very prone to human error.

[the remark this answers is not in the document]
autoantwort: I have now implemented that (I hope I have not missed a location, I am not 100% sure)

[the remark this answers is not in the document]
autoantwort: @ras0219-msft Is there anything left that is preventing this PR from being merged?

src/vcpkg/binarycaching.cpp
```diff
@@ -1716,6 +1768,43 @@ namespace vcpkg
...
+                std::swap(my_tasks, actions_to_push);
+            }
+            // Now, consume all of `my_tasks` before taking the lock again.
+            for (auto& action_to_push : my_tasks)
+            {
+                if (end_push_thread)
+                {
+                    msg::println(msgUploadRemainingPackages, msg::count = remaining_packages_to_push);
+                }
+                for (auto&& provider : m_providers)
```
ras0219-msft, line 1794: We already emit messages for each push from the providers, right? I think this message is unneeded verbosity.

ras0219-msft (changes requested):

src/vcpkg/binarycaching.cpp
```diff
@@ -1647,20 +1681,38 @@ namespace vcpkg
...
         {
-            for (auto&& provider : m_providers)
+            const auto clean_packages = action.build_options.clean_packages == CleanPackages::YES;
+            if (clean_packages)
             {
-                provider->push_success(action);
+                static int counter = 0;
+                Path new_packaged_dir = package_dir + "_push_" + std::to_string(++counter);
+                filesystem.remove_all(new_packaged_dir, VCPKG_LINE_INFO);
+                filesystem.rename(package_dir, new_packaged_dir, VCPKG_LINE_INFO);
```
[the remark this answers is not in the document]
autoantwort: I have written the code with https://github.com/microsoft/vcpkg-tool/pull/802 in mind (I already use it there locally) where a package is build multiple times.

include/vcpkg/base/message_sinks.h
```diff
@@ -87,4 +87,29 @@ namespace vcpkg
         CombiningSink(MessageSink& first, MessageSink& second) : m_first(first), m_second(second) { }
         void print(Color c, StringView sv) override;
     };
+
+    struct BGMessageSink : MessageSink
+    {
+        MessageSink& out_sink;
```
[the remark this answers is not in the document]
autoantwort: Yeah, does not really make sense to use it outside of the class.

src/vcpkg/binarycaching.cpp
```diff
@@ -459,7 +454,7 @@ namespace
 
         RestoreResult try_restore(const InstallPlanAction&) const override { return RestoreResult::unavailable; }
 
-        void push_success(const InstallPlanAction&) const override { }
+        void push_success(const BinaryProviderPushRequest&, MessageSink&) override { }
 
         void prefetch(View<InstallPlanAction> actions, View<CacheStatus*> cache_status) const override
```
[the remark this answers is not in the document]
autoantwort: I also thought about that and it would have really helped me in another PR, but since this is not necessary here I would do that in another PR to keep this PR small.

src/vcpkg/binarycaching.cpp
```diff
@@ -1564,11 +1567,41 @@ namespace vcpkg
...
+        }
+        bg_msg_sink.publish_directly_to_out_sink();
+        end_push_thread = true;
+        actions_to_push_notifier.notify_all();
+        push_thread.join();
+        if (have_remaining_packages)
+        {
+            msg::println(msgAllPackagesUploaded);
+        }
+    }
```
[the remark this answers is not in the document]
autoantwort: Ok I can remove it. I like it if a long running process has some kind of finishing message.

src/vcpkg/binarycaching.cpp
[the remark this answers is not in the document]
ras0219-msft: I think it'd be better to leave the design concerns of #802 in that PR and keep this one focused -- could you please remove the rename dance for now?

src/vcpkg/binarycaching.cpp
autoantwort: What should we print if a provider has two sources and one fails and one is successful?

autoantwort: We could also print
```
Stored binaries in {} destinations. 
```
because the files provider do not really upload binaries.

autoantwort: It is now 
```
Wait until the remaining packages (9) are uploaded
Stored binaries in 2 destinations.  (1/9)
Stored binaries in 2 destinations.  (2/9)
Stored binaries in 2 destinations.  (3/9)
Stored binaries in 2 destinations.  (4/9)
Stored binaries in 2 destinations.  (5/9)
Stored binaries in 2 destinations.  (6/9)
Stored binaries in 2 destinations.  (7/9)
Stored binaries in 2 destinations.  (8/9)
Stored binaries in 2 destinations.  (9/9)
➜  vcpkg git:(qtbase-framework) ✗ 
```

ras0219-msft (approved):
Thank you so much, this looks great!

The remaining items on our side:
1. I want to do a final review with the team of the technique here and ensure we're all aware. As you know, we don't use that much concurrency in vcpkg today and it does introduce cross-cutting concerns.
2. We need to determine how much risk this carries and whether we need a feature flag to turn it on/off (e.g. switch it back to synchronous).

include/vcpkg/base/message_sinks.h
```diff
@@ -87,4 +89,31 @@ namespace vcpkg
...
+        // must be called from consumer (synchronizer of out)
+        void print_published();
+
+        void publish_directly_to_out_sink();
+
+    private:
+        MessageSink& out_sink;
+
+        std::mutex m_lock;
+        // guarded by m_lock
```
BillyONeal, line 110: How about `m_published_lock` and `m_unpublished_lock`?

include/vcpkg/binarycaching.h
```diff
@@ -145,9 +176,27 @@ namespace vcpkg
...
+        BGMessageSink bg_msg_sink;
         std::unordered_map<std::string, CacheStatus> m_status;
         std::vector<std::unique_ptr<IBinaryProvider>> m_providers;
+        bool needs_nuspec_data = false;
+        std::condition_variable actions_to_push_notifier;
+        std::mutex actions_to_push_mutex;
+        std::vector<ActionToPush> actions_to_push;
+        std::thread push_thread;
+        std::atomic_bool end_push_thread;
+        std::atomic_int remaining_packages_to_push = 0;
```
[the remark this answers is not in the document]
BillyONeal: It seems like `remaining_packages_to_push` should be the sum of the vector sizes rather than tracked as a separate atomic var?

src/vcpkg/binarycaching.cpp
```diff
@@ -1647,20 +1678,29 @@ namespace vcpkg
...
     }
 
-    void BinaryCache::push_success(const InstallPlanAction& action)
+    void BinaryCache::push_success(const InstallPlanAction& action, Path package_dir)
     {
         const auto abi = action.package_abi().get();
         if (abi)
         {
-            for (auto&& provider : m_providers)
+            const auto clean_packages = action.build_options.clean_packages == CleanPackages::YES;
```
BillyONeal, line 1686: I think this should be left as an enum

src/vcpkg/binarycaching.cpp
```diff
@@ -1647,20 +1678,29 @@ namespace vcpkg
...
+            std::string nuspec;
+            if (needs_nuspec_data)
             {
-                provider->push_success(action);
+                NugetReference nuget_ref = make_nugetref(action, get_nuget_prefix());
+                nuspec = generate_nuspec(package_dir, action, nuget_ref);
             }
-
-            m_status[*abi].mark_restored();
+            std::unique_lock<std::mutex> lock(actions_to_push_mutex);
```
BillyONeal, line 1693:
```suggestion
            std::lock_guard<std::mutex> lock(actions_to_push_mutex);
```

include/vcpkg/binarycaching.h
```diff
@@ -42,6 +47,25 @@ namespace vcpkg
         const IBinaryProvider* m_available_provider = nullptr; // meaningful iff m_status == available
     };
 
+    struct BinaryPackageInformation
+    {
+        explicit BinaryPackageInformation(const InstallPlanAction& action, std::string&& nuspec = "");
+        std::string package_abi;
+        PackageSpec spec;
+        std::string raw_version;
+        std::string nuspec; // only filled if BinaryCache has a provider that returns true for needs_nuspec_data()
```
BillyONeal, line 56:
```suggestion
        Optional<std::string> nuspec; // only filled if BinaryCache has a provider that returns true for needs_nuspec_data()
```

BillyONeal (changes requested):
Request changes due to the CV deadlock.

src/vcpkg/binarycaching.cpp
```diff
@@ -1564,11 +1567,41 @@ namespace vcpkg
...
+        push_thread.join();
+        if (have_remaining_packages)
+        {
+            msg::println(msgAllPackagesUploaded);
+        }
+    }
+
+    BinaryCache::BinaryCache(Filesystem& filesystem)
+        : bg_msg_sink(stdout_sink)
+        , push_thread([this]() { push_thread_main(); })
```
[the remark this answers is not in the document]
BillyONeal: An advantage of starting the thread on push is it allows reuse after a wait_for_end step. No change requested.

src/vcpkg/binarycaching.cpp
```diff
@@ -1716,6 +1756,49 @@ namespace vcpkg
...
 
+    void BinaryCache::push_thread_main()
+    {
+        decltype(actions_to_push) my_tasks;
+        int count_pushed = 0;
+        while (true)
+        {
+            {
+                std::unique_lock<std::mutex> lock(actions_to_push_mutex);
+                actions_to_push_notifier.wait(lock, [this]() { return !actions_to_push.empty() || end_push_thread; });
```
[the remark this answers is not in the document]
BillyONeal: (Note that this proposal uses wait() rather than wait(pred) because the response is different depending on which part of the predicate is true)

autoantwort:
> I think our "independent interfaces / objects" model for binary caching is starting to leak under the strain of 6+ different backends.

I tried to solve this in #911, but this is currently also affected by the problem that the `ToolCache` is not thread-safe (and I am not totally happy with the interface of the IObjectProvider. It should be possible the report the success immediately after a download is done and not after all downloads are done so we can start unzipping while the rest is still downloaded.)

ras0219-msft:
I think that all the major refactoring I've planned has completed (https://github.com/microsoft/vcpkg-tool/pull/1022 shouldn't affect this PR too much).

We should ensure that no tools are downloaded concurrently.

autoantwort:
I have now rebased with main. 

> We should ensure that no tools are downloaded concurrently.

I think this is already the case because of all your preliminary work ❤️

src/vcpkg/binarycaching.cpp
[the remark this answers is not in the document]
autoantwort: I have implemented something similar

autoantwort:
@ras0219-msft @BillyONeal This is ready for review :)

include/vcpkg/base/batch-quere.h
```diff
@@ -0,0 +1,61 @@
...
+    void wait_for_items(std::vector<WorkItem>& out)
+    {
+        std::unique_lock<std::mutex> lock(m_mtx);
+        m_cv.wait(lock, [this]() { return !m_tasks.empty() || !m_running; });
+        m_tasks.pop(out);
+    }
+
+    void stop()
+    {
+        std::lock_guard<std ::mutex> lock(m_mtx);
```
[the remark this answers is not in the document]
autoantwort: Interesting that this was not done by clang-format

include/vcpkg/base/batch-quere.h
```diff
@@ -0,0 +1,61 @@
...
+        m_running = false;
+        m_cv.notify_all();
+    }
+
+    bool stopped() const { return !m_running; }
```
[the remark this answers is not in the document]
autoantwort: Is there any benefit?

src/vcpkg/binarycaching.cpp
```diff
@@ -2165,20 +2173,65 @@ namespace vcpkg
...
+        }
+        if (action.build_options.clean_packages == CleanPackages::YES)
+        {
+            m_fs.remove_all(action.package_dir.value_or_exit(VCPKG_LINE_INFO), VCPKG_LINE_INFO);
+        }
+    }
+
+    void BinaryCache::print_push_success_messages() { m_bg_msg_sink.print_published(); }
+
+    void BinaryCache::wait_for_async_complete()
```
Thomas1664, line 2190: I think the function name is a bit misleading: If I understand this function correctly, it kills the background thread?

src/vcpkg/base/message_sinks.cpp
```diff
@@ -58,4 +59,74 @@ namespace vcpkg
...
+    {
+        std::lock_guard<std::mutex> print_lk(m_print_directly_lock);
+        std::lock_guard<std::mutex> lk(m_published_lock);
+
+        m_print_directly_to_out_sink = true;
+        for (auto& messages : {&m_published, &m_unpublished})
+        {
+            for (auto&& m : *messages)
+            {
+                out_sink.print(m.first, m.second);
```
Thomas1664, line 127:
```suggestion
            for (auto&& msg : *messages)
            {
                out_sink.print(msg.first, msg.second);
```

Neumann-A: could also be replaced with a `std::for_each()` instead. I also feel like there should be an overload in the message sinks to automatically unpack `std::pair<Color, std::string>` or similar

src/vcpkg/binarycaching.cpp
```diff
@@ -2165,20 +2173,65 @@ namespace vcpkg
...
+        if (m_push_thread.joinable())
+        {
+            m_push_thread.join();
+        }
+    }
+
+    void BinaryCache::push_thread_main()
+    {
+        std::vector<ActionToPush> my_tasks;
+        int count_pushed = 0;
```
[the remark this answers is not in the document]
autoantwort: In theory. I prefer to use signed integers https://youtu.be/Puio5dly9N8?t=2560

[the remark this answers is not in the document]
Thomas1664: Given that `count_pushed` does nothing important I think it's ok (for now). On the other hand, you don't know how code might change in the future.
I don't want this to be the source of a hard-to-find bug.

> I prefer to use signed integers

What about `ssize_t` then?

I think it's not about which type you prefer - it's about which type is the correct one. Personally, I tend to use the correct type and avoid implicit conversions between integer types. Therefore, I'd even use `size_t` when looping over `std::array<std::string, 10>`.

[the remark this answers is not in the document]
Neumann-A: the correct type would by `std::vector<ActionToPush>::size_type` (or `difference_type`). Knowing that it boils down to `size_t` is knowing an implementation detail of `std::vector`. Just saying :P

include/vcpkg/binarycaching.h
```diff
@@ -196,23 +200,41 @@ namespace vcpkg
 
     struct BinaryCache : ReadOnlyBinaryCache
```
BillyONeal, line 201: When adding threading functionality to the codebase that is not effectively 'do a single function but faster', I think there needs to be a discussion in a comment of 'this is how the threads and communication between them work'.

For instance:

```
// compression and upload of binary cache entries happens on a single 'background' thread, `m_push_thread`
// Thread safety is achieved within the binary cache providers by:
//   1. Only using one thread in the background for this work.
//   2. Forming a queue of work for that thread to consume in `m_actions_to_push`, which maintains its own thread safety
//   3. Sending any replies from the background thread through `m_bg_msg_sink`
//   4. Ensuring any supporting data, such as tool exes, is provided before the background thread is started.
//   5. Ensuring that work is not submitted to the background thread until the corresponding `packages` directory to upload is no longer being actively touched by the foreground thread.
```

include/vcpkg/base/message_sinks.h
```diff
@@ -87,4 +89,31 @@ namespace vcpkg
...
+        std::vector<std::pair<Color, std::string>> m_published;
+        // buffers messages until newline is reached
+        // guarded by m_print_directly_lock
+        std::vector<std::pair<Color, std::string>> m_unpublished;
```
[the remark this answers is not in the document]
BillyONeal: Something like that? Maybe? I think what this really argues is that we really need a semi-standardized 'document' error type @ras0219-msft has been asking for ages... we're almost done getting rid of `ParseControlErrorInfo` which means we will finally have a unified way of handling errors and can finally look at messing with that.

include/vcpkg/binarycaching.h
BillyONeal: Does your approach here survive https://github.com/microsoft/vcpkg-tool/pull/1076 ? Hard links mean that we have a lot less certainty on the packages directory being 'hermetic'.

include/vcpkg/binarycaching.h
```diff
@@ -196,23 +200,41 @@ namespace vcpkg
...
+        {
+            BinaryPackageWriteInfo request;
+            CleanPackages clean_after_push;
+        };
+
+        void push_thread_main();
+
+        BGMessageSink m_bg_msg_sink;
+        BGThreadBatchQueue<ActionToPush> m_actions_to_push;
+        std::atomic_int m_remaining_packages_to_push = 0;
```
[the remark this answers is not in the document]
BillyONeal: This member existing at all is a smell to me. The number of items in the queue should be a part of the queue, not something tracked externally. Moreover, the queue already contains locks and stuff so I'm not sure why we need an atomic here.

include/vcpkg/base/batch-queue.h
```diff
@@ -0,0 +1,61 @@
...
+public:
+    template<class... Args>
+    void push(Args&&... args)
+    {
+        forward.emplace_back(std::forward<Args>(args)...);
+    }
+
+    bool empty() const { return forward.empty(); }
+
+    void pop(std::vector<T>& out)
```
BillyONeal, line 19: I don't like this thing being called queue given that this is how it works. Given that we expect this to be a multi producer single consumer queue, can we instead put the vector inside and note that only one thread may call pop but any number of threads may call push? That would also resolve the criticism over separate tracking atomics below.

src/vcpkg/base/message_sinks.cpp
```diff
@@ -58,4 +59,74 @@ namespace vcpkg
...
+            }
+
+            tmp.clear();
+        }
+    }
+
+    void BGMessageSink::publish_directly_to_out_sink()
+    {
+        std::lock_guard<std::mutex> print_lk(m_print_directly_lock);
+        std::lock_guard<std::mutex> lk(m_published_lock);
```
BillyONeal, line 120:
```suggestion
        std::lock_guard<std::mutex, std::mutex> lk(m_print_directly_lock, m_published_lock);
```

if this survives

BillyONeal (commented):
1. I'm going to tackle the structural error reporting etc. thing now that I think this change is really exposing / crying out for. In particular, I think we should consider a design which more formally associates chunks of console output with the package in question rather than a free for all on the message sink type requiring a lot of 'reverse engineer from stuff the caller already knew' kind of behavior.
2. I think the discussion of how we are achieving thread safety across several large components like this needs the comment like I described in a comment.
3. The comment about '`BinaryCache` is supposed to hide `unique_ptr`' should get fixed, which will also eliminate a lot of the changes in this PR attempting to adapt to that change.

When I have a solution for (1) are you interested in me just pushing changes into this PR or do you want a PR for your PR?

(By and large I think this change is structurally good)

autoantwort:
> When I have a solution for (1) are you interested in me just pushing changes into this PR or do you want a PR for your PR?

You can just push to this PR.  I will look through the other comments in the following days :)

include/vcpkg/base/batch-queue.h
autoantwort: `BatchQueue` alone is not thread safe. 

> Given that we expect this to be a multi producer single consumer queue, can we instead put the vector inside and note that only one thread may call pop but any number of threads may call push?

I don't get what you have in mind here 😅 

> I don't like this thing being called queue given that this is how it works.

Do you have an idea for a better name? :)

src/vcpkg/base/message_sinks.cpp
Thomas1664: I think you actually mean `std::scoped_lock`.

include/vcpkg/binarycaching.h
autoantwort: https://github.com/microsoft/vcpkg-tool/pull/1076 does not have an impact here. The files "copied" to the installed dir are not later changed by another package. Even if they are overwritten, the hard link in the packages folder would still link to the same original file.

include/vcpkg/binarycaching.h
```diff
@@ -196,23 +200,41 @@ namespace vcpkg
...
     {
-        static ExpectedL<BinaryCache> make(const VcpkgCmdArguments& args, const VcpkgPaths& paths, MessageSink& sink);
+        static ExpectedL<std::unique_ptr<BinaryCache>> make(const VcpkgCmdArguments& args,
+                                                            const VcpkgPaths& paths,
+                                                            MessageSink& sink);
 
         BinaryCache(const Filesystem& fs);
         BinaryCache(const BinaryCache&) = delete;
-        BinaryCache(BinaryCache&&) = default;
+        BinaryCache(BinaryCache&&) = delete;
```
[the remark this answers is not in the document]
autoantwort: I have now put all data in an extra struct that is hold via a std::unique_ptr by the BinaryCache class.

include/vcpkg/binarycaching.h
autoantwort: It has an impact on windows if https://github.com/microsoft/vcpkg-tool/pull/802 gets merged before this. Because when the packages folder gets compressed by 7z, the next feature gets tested and stuff gets removed from the installed dir and on windows you cant remove a hard link even if the linked file is only opened via another hard link -.-

autoantwort: I was finally able to catch this situation that happens in combination of this PR with #802:
image

autoantwort: Interestingly the catch seems to only happen with this specific rocksdb.lib file and error nearly only happens with lib files. 

And in general it only happens on the dev drive and not on a normal NTFS drive. Strage 😕

include/vcpkg/base/message_sinks.h
```diff
@@ -75,4 +77,30 @@ namespace vcpkg
...
+
+    private:
+        MessageSink& out_sink;
+
+        std::mutex m_published_lock;
+        std::vector<std::pair<Color, std::string>> m_published;
+
+        // buffers messages until newline is reached
+        // guarded by m_print_directly_lock
+        std::vector<std::pair<Color, std::string>> m_unpublished;
```
autoantwort, line 102: @BillyONeal Now that you have implemented the "error document" type stuff, I am right that this should be a `DiagnosticLine` etc. instead.

BillyONeal: I'm not sure yet, I ran into a problem where I'm not sure exactly how "status" type information should be conveyed through this infrastructure; I'm working on it over here: https://github.com/microsoft/vcpkg-tool/compare/main...BillyONeal:vcpkg-tool:message-sink-line

As part of that I realized that https://github.com/microsoft/vcpkg-tool/pull/1137 touches the same area and would be easier to merge first.

autoantwort: Why does a "status" need to be conveyed here at all? This are only informational messages. 
Or what is the problem you are trying to solve here?

BillyONeal: Some messages, like "I am about to touch the network now", are time sensitive and under normal conditions must be printed. However, errors/warnings like "the download failed" might need to be suppressed, if a subsequent retry / alternate makes the overall process succeed.

For example, when downloading a file, in "time order" let's say this happens:

1. We try to download the file from an asset cache. This should get a timely message because it touches the network.
2. Attempting to contact the asset cache fails. This normally emits an error.
3. We try to download the file from upstream. This needs a timely message because it touches the network.
4. Download from the real upstream succeeds. We need to 'go back in time' and not emit the error from #2.
5. We try to submit the freshly downloaded file back to the asset cache. This needs a timely message because it touches the network.
6. Submitting back to the asset cache fails for some reason. That normally emits an error, but in this context needs to be reduced to a warning because we can continue without it.

If any of this is happening on 'background' threads, even the 'timely' messages need to be held until the next synchronization point with the thread that owns the console. This is what 'statusln' is for in my WIP. They need to share one channel rather than just passing both MessageSink and DiagnosticContext to handle this background case where a caller wants to keep the original-time-order of diagnostics and status.

autoantwort: > If any of this is happening on 'background' threads, even the 'timely' messages need to be held until the next synchronization point with the thread that owns the console. 

This PR already does this (for the scope of the PR)

> background case where a caller wants to keep the original-time-order of diagnostics and status

That is already the case in the PR. But yeah this PR don't know when a message ends (if it spans multiple lines), the whole reason for the error document type stuff to be created. But could this not simply be solved by buffering `DiagnosticLine`s instead and the use of DiagnostigContext instead of MessageSink in this PR, or what breaks then?
Currently: every message (regardless of the type) -> MessageSinkBuffer -> MessageSink 
Future: every message (regardless of the type) -> DiagnosticLineBuffer -> MessageSink

BillyONeal: > But could this not simply be solved by buffering DiagnosticLines instead and the use of DiagnostigContext instead of MessageSink in this PR, or what breaks then?

Then 'inner' parts have no way to emit the 'intended to be timely' messages. What I'm doing is:

1. Add one function, `statusln` to `DiagnosticContext` where these timely messages go. Normal buffering of errors won't buffer these, but background thread stuff will. Notably, there is no non-`ln` version because it needs to be a reasonable buffering point.
2. Teach `downloads.cpp` et al. to use `DiagnosticContext`, which implied teaching `system.process.cpp` how to use `DiagnosticContext`.

autoantwort: > Then 'inner' parts have no way to emit the 'intended to be timely' messages. 

Iiuc 'intended to be timely' = must be printed immediately: The inner parts that don't run in the background can print there stuff immediately via the MessageSink/DiagnosticContext and the stuff in the background thread is never allowed to print stuff immediately otherwise you get interleaved messages with the build output. 
So I don't understand why we need this extra "print messages immediately" channel when the only possible states are "print everything immediately" or "print nothing immediately". 
Maybe I should just wait and see what your resulting code looks like 😅

BillyONeal: >So I don't understand why we need this extra "print messages immediately" channel when the only possible states are "print everything immediately" or "print nothing immediately".

No, there's a third condition, which is step 4 in my example above. We need to buffer errors and warnings from the inner operation, because we may want to swallow and not emit them if a subsequent attempt succeeds, but we must not buffer any of the timely status messages.

Example: https://github.com/BillyONeal/vcpkg-tool/blob/9aa671863a68ef90d0c355e4594bd9925d9df083/src/vcpkg/base/downloads.cpp#L926-L966

autoantwort: Aah thanks for the example! :) 
But iiuc this "problem" is not caused by this PR and already existed beforehand?

BillyONeal: Yes, it has nothing to do with your PR. It's a problem I ran into trying to `DiagnosticContext`-ize the downloads stuff, which I want to do in order to have confidence that your PR is correct.

src/vcpkg/binarycaching.cpp
```diff
@@ -2417,92 +2420,165 @@ namespace vcpkg
...
 
                 size_t num_destinations = 0;
-                for (auto&& provider : m_config.write)
+                for (auto&& provider : ro_data->m_config.write)
                 {
-                    if (!provider->needs_zip_file() || request.zip_path.has_value())
+                    if (!provider->needs_zip_file() || action_to_push.request.zip_path.has_value())
                     {
-                        num_destinations += provider->push_success(request, out_sink);
+                        num_destinations += provider->push_success(action_to_push.request, stdout_sink);
```
BillyONeal, line 2557:
```suggestion
                        num_destinations += provider->push_success(action_to_push.request, m_bg_msg_sink);
```

Just noting that I noticed this, I have already fixed it.

BillyONeal (approved):

BillyONeal:
@autoantwort I pushed some changes here, can you let me know if you are happy with them? Thanks!

src/vcpkg/binarycaching.cpp
```diff
@@ -2106,9 +2106,155 @@ namespace vcpkg
                 get_environment_variable(EnvironmentVariableGitHubSha).value_or("")};
     }
 
-    static ExpectedL<BinaryProviders> make_binary_providers(const VcpkgCmdArguments& args, const VcpkgPaths& paths)
+    void ReadOnlyBinaryCache::fetch(View<InstallPlanAction> actions)
+    {
+        std::vector<const InstallPlanAction*> action_ptrs;
```
BillyONeal, line 2111: This block is just moved up from 2325

src/vcpkg/binarycaching.cpp
```diff
@@ -2106,9 +2106,155 @@ namespace vcpkg
...
+                }
+            }
+        }
+
+        return Util::fmap(statuses, [](CacheStatus* s) {
+            return s->get_available_provider() ? CacheAvailability::available : CacheAvailability::unavailable;
+        });
+    }
+
+    void BinaryCacheSynchronizer::add_submitted() noexcept
```
BillyONeal, line 2216: This starts meaningfully new code.

include/vcpkg/binarycaching.h
```diff
@@ -211,6 +211,41 @@ namespace vcpkg
         std::unordered_map<std::string, CacheStatus> m_status;
     };
 
+    struct BinaryCacheSyncState;
+    struct BinaryCacheSynchronizer
+    {
+        using backing_uint_t = std::conditional_t<sizeof(size_t) == 4, uint32_t, uint64_t>;
+        using counter_uint_t = std::conditional_t<sizeof(size_t) == 4, uint16_t, uint32_t>;
```
autoantwort, line 218: Why does this depend on size_t?

autoantwort (commented):
LGTM

include/vcpkg/binarycaching.h
BillyONeal: There are a lot of 64 bit machines without 32 bit atomics, and a lot of 32 bit machines without 64 bit atomics, and I wanted to choose something least likely to put us into lockful atomics world.

BillyONeal:
Thanks for the contribution!

BillyONeal:
time results

Neumann-A:
Why is this so inconsistent? I would have expected less variance in the result. Especially for stuff taking 1d and longer.

dg0yt:
Does the artifact size with static linkage explain most of the inconsistency? In particular when ports install executables.

Neumann-A:
> Does the artifact size with static linkage explain most of the inconsistency? In particular when ports install executables.

Hmm maybe. The android triplets are mor ore less consistent and the -static and -static-md are also more or less consistent. @BillyONeal do you have storage data for the different triplets?

BillyONeal:
The difference being mostly a function of how big the binary cache size is is my supposition as well. I don't have those stats though. For instance, the triplets with an LLVM have more improvement. The improvement for macOS seems bigger, which might be explained by not being in the same data center as the caches.

PR #934, opened 2023-02-27
PostBuildChecks: Append output to stdout-<triplet>.log

autoantwort:
Closes https://github.com/microsoft/vcpkg/issues/11219


include/vcpkg/base/files.h
```diff
@@ -142,9 +142,14 @@ namespace vcpkg
 
     struct WriteFilePointer : FilePointer
     {
+        enum class Append
+        {
+            YES,
+            NO,
+        };
```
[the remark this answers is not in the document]
autoantwort: `open_for_write(file_path, WriteFilePointer::Append::YES, ec);` is more expressive than `open_for_write(file_path, true, ec);`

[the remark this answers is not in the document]
Neumann-A: My editor(VS Code) will already show it as:
`open_for_write(file_path: file_path, append: true, ec: ec);`

What about ?
```
struct append {
    bool b;
    append(bool a) : b(a) {};
};
```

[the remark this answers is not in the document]
Neumann-A: Implicitly constructable from bool
Implicitly convertible to bool (if wanted) e.g. (`::fopen(file_path.c_str(), append == Append::YES ? "ab" : "wb");`)

also casting the current enum order to bool has a probably unintended result ;)

Neumann-A, line 149:
```suggestion
        enum class Append
        {
            NO = 0,
            YES,
        };
```
still needs this to be consistent.

include/vcpkg/base/files.h
```diff
@@ -320,8 +325,13 @@ namespace vcpkg
         ReadFilePointer open_for_read(const Path& file_path, LineInfo li) const;
         ExpectedL<ReadFilePointer> try_open_for_read(const Path& file_path) const;
 
-        virtual WriteFilePointer open_for_write(const Path& file_path, std::error_code& ec) = 0;
+        virtual WriteFilePointer open_for_write(const Path& file_path,
+                                                WriteFilePointer::Append append,
+                                                std::error_code& ec) = 0;
+        virtual WriteFilePointer open_for_write(const Path& file_path, std::error_code& ec);
```
BillyONeal, line 331:
```suggestion
        WriteFilePointer open_for_write(const Path& file_path, std::error_code& ec);
```

include/vcpkg/base/files.h
```diff
@@ -320,8 +325,13 @@ namespace vcpkg
...
         WriteFilePointer open_for_write(const Path& file_path, LineInfo li);
+        virtual WriteFilePointer open_for_appending(const Path& file_path, std::error_code& ec);
+        WriteFilePointer open_for_appending(const Path& file_path, LineInfo li);
```
BillyONeal, line 334:
```suggestion
        virtual WriteFilePointer open_for_append(const Path& file_path, std::error_code& ec);
        WriteFilePointer open_for_append(const Path& file_path, LineInfo li);
```

include/vcpkg/base/messages.h
```diff
@@ -412,6 +412,32 @@ namespace vcpkg
             this->print(c, msg::format(m, args...).append_raw('\n'));
         }
 
+        void println_warning(const LocalizedString& s);
```
BillyONeal, line 415: I'm weakly against this: output redirection and formatting the error prefix on are separate concerns and I don't think `sink.println(msg::format_error(msgFoo))` is that onerous. But I am on record as not being a fan of any "format wrappers" so we should get other maintainers' thoughts.

src/vcpkg/base/messages.cpp
```diff
@@ -490,6 +490,20 @@ namespace
     StdErrMessageSink stderr_sink_instance;
 }
 
+namespace vcpkg
+{
+
```
BillyONeal, line 495:
```suggestion
```

Also it seems like these should be moved into the order that they are next to the other MessageSink members.

src/vcpkg/base/messages.cpp
```diff
@@ -490,6 +490,20 @@ namespace
...
+    void MessageSink::println_warning(const LocalizedString& s)
+    {
+        print(Color::warning, msg::format(msg::msgWarningMessage).append(s).append_raw('\n'));
```
BillyONeal, line 498: If this member survives.

```suggestion
        println(Color::warning, msg::format_warning(s));
```

src/vcpkg/base/messages.cpp
```diff
@@ -490,6 +490,20 @@ namespace
...
+    }
+
+    void MessageSink::println_error(const LocalizedString& s)
+    {
+        print(Color::error, msg::format(msg::msgErrorMessage).append(s).append_raw('\n'));
```
BillyONeal, line 503: If this member survives.

```suggestion
        println(Color::error, msg::format_error(s));
```

src/vcpkg/build.cpp
```diff
@@ -973,8 +973,28 @@ namespace vcpkg
...
+                    Checks::msg_check_exit(VCPKG_LINE_INFO,
+                                           out_file.write(sv.data(), 1, sv.size()) == sv.size(),
+                                           msgErrorWhileWriting,
+                                           msg::path = stdoutlog);
+                }
+            };
+            FileSink file_sink{fs, stdoutlog};
+            return perform_post_build_lint_checks(
+                action.spec, paths, pre_build_info, build_info, scfl.source_location, file_sink);
+        }();
```
BillyONeal, line 997: I really don't think making this an immediately invoked lambda does anything.

src/vcpkg/build.cpp
```diff
@@ -973,8 +973,28 @@ namespace vcpkg
         }
 
         const BuildInfo build_info = read_build_info(fs, paths.build_info_file_path(action.spec));
-        const size_t error_count =
-            perform_post_build_lint_checks(action.spec, paths, pre_build_info, build_info, scfl.source_location);
+        const size_t error_count = [&]() {
+            struct FileSink : MessageSink
```
BillyONeal, line 977: This seems a reasonable thing on its own, no reason to make it local to this function.

src/vcpkg/build.cpp
```diff
@@ -973,8 +973,28 @@ namespace vcpkg
...
+            {
+                WriteFilePointer out_file;
+                StringView stdoutlog;
```
BillyONeal, line 980: Member `StringView` is a bit scary because it makes the lifetime here depend on outliving the input. And worse, that's only "discoverable" when an error happens.

```suggestion
                Path stdoutlog;
```

src/vcpkg/build.cpp
```diff
@@ -973,8 +973,28 @@ namespace vcpkg
...
+                FileSink(Filesystem& fs, StringView stdoutlog)
+                    : out_file(fs.open_for_appending(stdoutlog, VCPKG_LINE_INFO)), stdoutlog(stdoutlog)
```
BillyONeal, line 982:
```suggestion
                    : m_stdoutlog(stdoutlog), m_out_file(fs.open_for_appending(m_stdoutlog, VCPKG_LINE_INFO))
```

We had to construct a Path anyway here. (This does require flipping the order in which the members are declared)

BillyONeal (changes requested):

src/vcpkg/build.cpp
autoantwort: It allows 
`const size_t error_count = [&]() { ...`, but I can also use 
```c++
size_t error_count = 0;
{
  ...
  error_count = ...;
}
```

autoantwort:
I have now moved the `FileSink` into the new `message_sinks.h` header. I have created a new header because `messages.h` should not include `files.h` because `messages.h` is included everywhere but `FileSink` is nearly needed nowhere. Now it is only necessary to include `base/fwd/message_sinks.h` to get `stdout_sink` etc.

include/vcpkg/base/files.h
```diff
@@ -140,13 +140,14 @@ namespace vcpkg
         ExpectedL<Unit> try_read_all_from(long long offset, void* buffer, std::uint32_t size);
     };
 
+    enum class Append
```
BillyONeal, line 143: This should be in the fwd

BillyONeal (approved):
Thanks!

include/vcpkg/base/files.h
autoantwort: Done. I also removed some unnecessary included (the fwd header was only included transitively by other not needed includes)

BillyONeal (approved):

BillyONeal:
Thanks!

PR #996, opened 2023-03-31
Extract Binary Cache refactoring changes from microsoft/vcpkg-tool#908

ras0219-msft:


ras0219-msft:
Merging without additional review because this is a subset of #908.

PR #1001, opened 2023-04-02
Remove gcov due to flakiness

ras0219-msft:
GCOV is causing a lot of builds to fail randomly and succeed on retry. This PR removes it.

e.g. https://dev.azure.com/vcpkg/public/_build/results?buildId=87614&view=logs&j=d6955a24-b4a1-58cd-bf05-f0addec06f6d&t=234a88e1-e3e9-55cf-a2d8-fa41708c6fae

### DIFF
--- a/azure-pipelines/pipelines.yml
+++ b/azure-pipelines/pipelines.yml
@@ -43,7 +43,7 @@ jobs:
         fi
       displayName: 'Rush install, build and test vcpkg-artifacts'
     - bash: |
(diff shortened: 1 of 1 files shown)


dg0yt:
I already proposed to turn these steps' errors into non-fatal.

ras0219-msft:
I'm not sure why we should keep the steps around if they:

- Add time to the build
- Often fail
- Aren't required to merge

If we want to perform one-off investigations into code coverage, that doesn't need to run as part of our pipeline (running locally is vastly more important in that case so one can try out changes).

On the other hand, if we wanted to enforce monotonic coverage requirements, the tool must be of sufficiently high quality to not fail spuriously.

BillyONeal (approved):

PR #1016, opened 2023-04-11
Add new ways to report build failures and giving hint messages from portfile.cmake

autoantwort:
Motivation 1: In cases like https://github.com/microsoft/vcpkg/issues/30744 users should not create issues (problem was insufficient disk space). In `vcpkg_execute_build_process` we could add a call to `vcpkg_fail_with_user_interaction_required` with a proper error message saying the disk is full.

Motivation 2: increase visibility or warnings since users report issues like this: https://github.com/microsoft/vcpkg/issues/30592#issuecomment-1493079488 (`Could not find make.  Please install it through your package manager.`) => `vcpkg_fail_with_user_interaction_required`. Another example would be https://github.com/microsoft/vcpkg/issues/31276 or https://github.com/microsoft/vcpkg/issues/31312 or https://github.com/microsoft/vcpkg/issues/35151 or https://github.com/microsoft/vcpkg/issues/36541 or https://github.com/microsoft/vcpkg/issues/35991 or https://github.com/microsoft/vcpkg/issues/36011

Motivation 3: Many ports emits warnings like the following 
```
alsa currently requires the following libraries from the system package manager:
    autoconf libtool
These can be installed on Ubuntu systems via sudo apt install autoconf libtool
```
This is confusing if the libs are already installed and the build is successful. These messages should only be printed when the build fails and they should be printed after the "how to report an issue" block to be more visible. =>  `vcpkg_user_hint_on_failure`

Also see docs PR https://github.com/microsoft/vcpkg-docs/pull/60


src/vcpkg/build.cpp
```diff
@@ -979,7 +979,19 @@ namespace vcpkg
...
-            return ExtendedBuildResult{BuildResult::BUILD_FAILED, stdoutlog, std::move(error_logs)};
+            ExtendedBuildResult result{BuildResult::BUILD_FAILED, stdoutlog, std::move(error_logs)};
+            const auto user_required_path =
+                buildpath / Strings::concat("required-user-interaction-", action.spec.triplet(), ".txt");
+            if (fs.exists(user_required_path, VCPKG_LINE_INFO))
+            {
+                result.user_required_interaction = fs.read_contents(user_required_path, VCPKG_LINE_INFO);
+            }
+            const auto user_hints_path = buildpath / Strings::concat("user-hints-", action.spec.triplet(), ".txt");
+            if (fs.exists(user_hints_path, VCPKG_LINE_INFO))
```
Thomas1664, line 990: Better make sure that it's a regular file

autoantwort: Since it is only an internal interface I think it is ok here. Crashing in case of an folder is a good thing. And I don't care if someone wants to use symbolic links in the future.

ras0219-msft:
First, I strongly agree we need to improve these cases. Better build error messages for users has tons of potential. I have a few different areas I wanted to comment on, so apologies in advance for the length.

### Restating Cases
Listing out some more error cases that the user can fix, along with the diagnostic difficulty:
- Missing system libraries (hard)
- Permissions (hard -- looks the same as other issues?)
- MAX_PATH (medium)
- Out of Disk (easier)
- Out of Memory (easier)
- Missing Windows SDK (easier)
- Compiler issues (easier?)

Under this lens, I think your proposed mechanisms are intended to handle:
- Difficult to diagnose / specific-to-port (fail-hints)
- Easy to diagnose and generic (fail-require-user-interaction)

Instead of separating these two cases entirely, what about only having hints and, if hints are provided, reduce the prominence of the issue message?

### Implementation: Port vs Tool
I see a design decision between putting the port in charge versus creating heuristics in the vcpkg tool. Port-based approaches are more easily iterated on, but are also very limiting. It is quite painful to try to regex through build logs from an "error handler" inside CMake (as we've done for some things). On the other hand, in the tool itself progress is slower and it often wouldn't be "worth it" to handle highly exceptional cases.

A bonus of implementing things in the tool is that they get localized :).

### Port-approach
I'm not convinced that raw text known-paths in `buildtrees/` are the right extension interface. Some alternatives to consider in no particular order:
- Named Pipe / Unix domain socket back to the vcpkg parent process (complex to implement, but opens up interesting doors)
- In-Band sequences to standard out (a la CI systems):
    `message("::hintonfail::You may need to install xyz from your system package manager")`
    (https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions)
- Instead of having the file path baked into the scripts, we could use an environment variable:
    `file(APPEND "$ENV{HINT_FILE}" "Hint Message\n")`
    (https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files)

Fortunately all these should be retrofit-able underneath a CMake function call interface (as you've implemented).

autoantwort:
> Under this lens, I think your proposed mechanisms are intended to handle:
>
 >  - Difficult to diagnose / specific-to-port (fail-hints)
  > -  Easy to diagnose and generic (fail-require-user-interaction)

Probably I should not hide the description in the docs PR: https://github.com/microsoft/vcpkg-docs/pull/60

PR #1025, opened 2023-04-14
Expose vcpkglib's Threads dependency

SchaichAlonso:
The vcpkglib target generates a precompiled header, which later during the build process is consumed by the vcpkg-target target.

However, vcpkglib is compiled with threading support, while vcpkg-test is not, causing the PCH being rejected.

Enforce multithreadened target mode to all vcpkglib consumers by marking its' Threads::Threads dependency public

Fixes: microsoft/vcpkg#30863


ras0219-msft:
LGTM, thanks for the quick fix!

PR #1029, opened 2023-04-18
x-update-baseline should not make overlay-triplet paths absolute

JavierMatosD:
This PR fixes a bug where the command `x-update-baseline` makes overlay-triplet paths absolute in vcpkg-configuration.json.


ras0219-msft (approved):
Can you add either unit tests or an end-to-end test for this?

ras0219-msft (approved):

PR #1036, opened 2023-04-23
Provide better diagnostics to non-admin users of 'integrate install'.

dubejf:
Building up on #1034, this PR treats failures as warnings in `integrate_install_msbuild14()`.

### 1) Happy path

```
.\vcpkg.exe integrate install
[green] Applied user-wide integration for this vcpkg root.
CMake projects should use: "-DCMAKE_TOOLCHAIN_FILE=..../vcpkg/scripts/buildsystems/vcpkg.cmake"

All MSBuild C++ projects can now #include any installed libraries. Linking will be handled automatically. Installing new libraries will make them instantly available.
```

No change. This is fine.

### 2) With %LOCALAPPDATA%\vcpkg set as read-only

```
.\vcpkg.exe integrate install
[red] write_contents("C:\Users\...\AppData\Local\vcpkg\vcpkg.user.props"): permission denied
```

No change. This failure mode seems unlikely, and the simplicity of a hard error is reasonable to me.

### 3) Declining UAC elevation to modify MSBuild14 files under to C:\Program Files (x86)\

```
.\vcpkg.exe integrate install
[yellow] warning: failed to install system targets file to C:\Program Files (x86)\MSBuild/Microsoft.Cpp/v4.0/V140/ImportBefore/Default/vcpkg.system.props
[yellow] warning: Integration was not applied.
CMake projects should use: "-DCMAKE_TOOLCHAIN_FILE=...../vcpkg/scripts/buildsystems/vcpkg.cmake"

All MSBuild C++ projects can now #include any installed libraries. Linking will be handled automatically. Installing new libraries will make them instantly available.
```

For users only interested in VS2017 and later toolchains, this is more informative than the current version, which fails with either

```
[current version]
.\vcpkg.exe integrate install
[yellow] warning: Integration was not applied.
```

or 

```
[current version]
.\vcpkg.exe integrate install
[red] failed to install system targets file to C:\Program Files (x86)\MSBuild/Microsoft.Cpp/v4.0/V140/ImportBefore/Default/vcpkg.system.props
```

IMO the text in the new version should be improved ("Integration was not applied" should talk about partial success, "All MSBuild C++ projects..." is not strictly true here), but I didn't want to introduce new strings.

What do you think?


dubejf:

src/vcpkg/commands.integrate.cpp
```diff
@@ -272,51 +272,46 @@ namespace vcpkg::Commands::Integrate
...
+        Checks::msg_check_exit(VCPKG_LINE_INFO,
+                               !fs.exists(SYSTEM_WIDE_TARGETS_FILE, IgnoreErrors{}),
+                               msgSystemTargetsInstallFailed,
+                               msg::path = SYSTEM_WIDE_TARGETS_FILE);
 
-            Checks::msg_check_exit(VCPKG_LINE_INFO,
-                                   fs.exists(SYSTEM_WIDE_TARGETS_FILE, IgnoreErrors{}),
-                                   msgSystemTargetsInstallFailed,
-                                   msg::path = SYSTEM_WIDE_TARGETS_FILE);
+        if (!fs.exists(SYSTEM_WIDE_TARGETS_FILE, IgnoreErrors{}))
```
ras0219-msft, line 308: This `if()` will never fail -- because `msg_check_exit()` immediately before will terminate the program.

You probably wanted to remove the `msg_check_exit()` call.

ras0219-msft (changes requested):

src/vcpkg/commands.integrate.cpp
dubejf: You're right, that shouldn't be there. Sorry about that.

BillyONeal (approved):

src/vcpkg/commands.integrate.cpp
```diff
@@ -336,9 +328,16 @@ namespace vcpkg::Commands::Integrate
...
+        install_complete = integrate_install_msbuild14(fs);
 #endif
-        msg::println(Color::success, msgAppliedUserIntegration);
+        if (install_complete)
+        {
+            msg::println(Color::success, msgAppliedUserIntegration);
+        }
+        else
+        {
+            msg::println_warning(msgIntegrationFailed);
```
ras0219-msft, line 339: This change also causes `integrate install` to exit with a 0 exit code (SUCCESS) when the msbuild integration fails.

I think you will need to check `install_complete` below and conditionalize the `exit_success`.

ras0219-msft (changes requested):
Sorry for the piecemeal review -- I didn't review closely enough the first time.

In your proposal, it seems like your intent is to keep all the "external" behavior the same (failures are still failures, successes are successes), but just change the messages displayed to the user to be more helpful. In that case, we need the exit codes to remain the same so that scripts/CI systems will react in the same way.

src/vcpkg/commands.integrate.cpp
```diff
@@ -272,51 +272,41 @@ namespace vcpkg::Commands::Integrate
                 switch (user_choice)
                 {
                     case ElevationPromptChoice::YES: break;
-                    case ElevationPromptChoice::NO:
-                        msg::println_warning(msgPreviousIntegrationFileRemains);
-                        Checks::exit_fail(VCPKG_LINE_INFO);
+                    case ElevationPromptChoice::NO: msg::println_warning(msgPreviousIntegrationFileRemains); break;
```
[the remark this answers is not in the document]
dubejf: Sorry, I didn't explain myself correctly. I did want to propose a change in behavior.

`integrate install` _used to_ perform these steps, in this order:

1- Write a props file to `%PROGRAMFILES(x86)%`.
2- Write props/targets files to `%LOCALAPPDATA%`.
3- Display success and usage messages.

PR #1034 reversed to order of the first two operations. The new order is:

1- Write props/targets files to `%LOCALAPPDATA%`.
2- Write a props file to` %PROGRAMFILES(x86)%`.
3- Display success and usage messages.

This means that a user with no write access to `%PROGRAMFILES(x86)%` now gets a mostly working integration (except for VS2015/MSBuild14).

However, for this mostly happy unprivileged user, the command terminates with an error and doesn't display any usage message or indication of (partial) success.

Perhaps it would be better not to treat a failure to write to `%PROGRAMFILES(x86)%` as a hard error, but instead to report it as a warning, and to print the usage message at the end? This is what the PR is proposing.

There is no observable change if the MSBuild14 directory is writable, or if its content is up to date. Other errors, such as failure to write `%LOCALAPPDATA%`, still result in termination. This is just about being tolerant to FS errors when writing to `%PROGRAMFILES(x86)%`.

Anyway, that was the initial aim. What do you think of this approach?

Personally, after thinking about it for a few days, I don't think that this PR can be applied in its actual state. In fact, I am considering closing it. For instance, saying `All MSBuild C++ projects can now #include any installed libraries...` in the case of partially applied integration is misleading. Unless the messages are reworded, I think that the status quo is better.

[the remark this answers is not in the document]
ras0219-msft: I think it's a good change to "do what we can", but I just think we should still report a nonzero exit code if there was partial failure.

Agreed that rewording the `All MSBuild C++ projects...` message would be very good. In the case of partial VS2015 failure, perhaps something like `Visual Studio 2017 and later can now #include any installed libraries`.

src/vcpkg-test/dependencies.cpp
```diff
@@ -8,6 +8,7 @@
 #include <vcpkg/triplet.h>
 
 #include <memory>
+#include <numeric>
```
dubejf, line 11: Required for std::iota() on VS 17.5.4

src/vcpkg/commands.integrate.cpp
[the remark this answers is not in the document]
dubejf: Done! Thanks for helping wording the messages.

include/vcpkg/base/message-data.inc.h
```diff
@@ -198,6 +198,11 @@ DECLARE_MESSAGE(AuthenticationMayRequireManualAction,
                 "",
                 "One or more {vendor} credential providers requested manual action. Add the binary source "
                 "'interactive' to allow interactivity.")
+DECLARE_MESSAGE(AutomaticLinkingForMSBuild15AndLater,
```
ras0219-msft, line 201:
```suggestion
DECLARE_MESSAGE(AutomaticLinkingForVS2017AndLater,
```

include/vcpkg/base/message-data.inc.h
```diff
@@ -1478,7 +1483,7 @@ DECLARE_MESSAGE(IntegrateZshHelp,
                 (),
                 "'zsh' is a terminal program which should be unlocalized.",
                 "Enable zsh tab-completion.")
-DECLARE_MESSAGE(IntegrationFailed, (), "", "Integration was not applied.")
+DECLARE_MESSAGE(IntegrationFailedMSBuild14, (), "", "Integration was not applied for Visual Studio 2015.")
```
ras0219-msft, line 1486:
```suggestion
DECLARE_MESSAGE(IntegrationFailedVS2015, (), "", "Integration was not applied for Visual Studio 2015.")
```

ras0219-msft (approved):

BillyONeal (approved):

PR #1039, opened 2023-04-25
Add z-extract command / artifacts uses z-extract 

JavierMatosD:
* Exposes the extract_archive function via z-extract command
* Removes archivers from vcpkg-artifacts and uses z-extract instead
* modifies unit tests

Note:
* removes progress bar for extraction in artifacts world


include/vcpkg/base/message-data.inc.h
```diff
@@ -887,6 +887,7 @@ DECLARE_MESSAGE(ExportUnsupportedInManifest,
                 "vcpkg export does not support manifest mode, in order to allow for future design considerations. "
                 "You may use export in classic mode by running vcpkg outside of a manifest-based project.")
 DECLARE_MESSAGE(ExtendedDocumentationAtUrl, (msg::url), "", "Extended documentation available at '{url}'.")
+DECLARE_MESSAGE(ExtractHelp, (), "", "Extracts archive to a given location.")
```
ras0219-msft, line 890:
```suggestion
DECLARE_MESSAGE(ExtractHelp, (), "", "Extracts an archive to a given location.")
```

src/vcpkg/commands.cpp
```diff
@@ -84,6 +85,7 @@ namespace vcpkg::Commands
         {"create", Create::perform_and_exit},
         {"deactivate", command_deactivate_and_exit},
         {"edit", Edit::perform_and_exit},
+        {"x-extract", extract_command_and_exit},
```
ras0219-msft, line 88:
```suggestion
        {"z-extract", z_extract_command_and_exit},
```

src/vcpkg/commands.extract.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include <vcpkg/vcpkgcmdarguments.h>
+#include <vcpkg/vcpkgpaths.h>
+
+namespace vcpkg::Commands
+{
+    const CommandStructure ExtractCommandStructure = {
+        [] { return msg::format(msgExtractHelp); },
+        2,
+        2,
+        {{}, {}, {}},
```
ras0219-msft, line 15: As discussed offline, this probably needs some argument to "override" the format detection to support unzipping formats like `vsix`, `nupkg`, etc.

include/vcpkg/base/message-data.inc.h
BillyONeal: Proposal:

```
DECLARE_MESSAGE(ExtractHelp, (), "", "Extracts an archive.")
```

dg0yt:
Is there a test that this correctly restores the executable bit of those platforms which support it?

src/vcpkg/commands.cpp
```diff
@@ -85,6 +86,7 @@ namespace vcpkg::Commands
         {"create", Create::perform_and_exit},
         {"deactivate", command_deactivate_and_exit},
         {"edit", Edit::perform_and_exit},
+        {"z-extract", extract_command_and_exit},
```
dan-shaw, line 89: I think this should be `command_extract_and_exit` or in a namespace `Extract::` to keep the same style as the other functions.

src/vcpkg/commands.extract.cpp
```diff
@@ -0,0 +1,191 @@
...
+            std::string value = iter->second;
+            try
+            {
+                return std::stoi(value);
+            }
+            catch (std::exception&)
+            {
+                msg::println_error(msgErrorInvalidStripOption, msg::option = OPTION_STRIP, msg::value = value);
+                Checks::exit_fail(VCPKG_LINE_INFO);
+                return 0;
```
dan-shaw, line 73: I think this return is unreachable.

include/vcpkg/archives.h
```diff
@@ -13,16 +13,39 @@
 
 namespace vcpkg
 {
+    enum ExtractionType
```
dan-shaw, line 16:
```suggestion
    enum class ExtractionType
```

vcpkg-artifacts/registries/RemoteRegistry.ts
```diff
@@ -90,7 +90,8 @@ export class RemoteRegistry extends ArtifactRegistry {
 
     const file = await acquireArtifactFile(this.session, locations, `${this.safeName}-registry.zip`, {}, {force: true});
     if (await file.exists()) {
-      await unpackZip(this.session, file, this.cacheFolder, {}, { strip: -1 });
+      const targetLocation = this.cacheFolder.fsPath;
+      await vcpkgExtract(this.session, file.fsPath, targetLocation, "--strip=1");
```
dan-shaw, line 94: Maybe we should pass in a number instead of the full `--strip=` argument

vcpkg-artifacts/vcpkg.ts
```diff
@@ -44,6 +44,16 @@ export function vcpkgFetch(session: Session, fetchKey: string): Promise<string>
   });
 }
 
+export function vcpkgExtract(session: Session, archive: string, target:string, strip:string): Promise<string> {
+  return runVcpkg(session.vcpkgCommand, ['z-extract', archive, target, strip]).then((output) => {
+    return output;
+  }, (error) => 
+ {
+    session.channels.warning('Failed to extract: ' + archive + '\n');
```
BillyONeal, line 52: This needs localization. But I think just emitting what vcpkg.exe said is fine. Note that I couldn't do that for the download one because `--machine-readable-progress`.

vcpkg-artifacts/session.ts
```diff
@@ -100,9 +97,6 @@ export class Session {
 
   /** register installer functions here */
   private installers = new Map<string, InstallerTool>([
-    ['nuget', installNuGet],
-    ['unzip', installUnZip],
-    ['untar', installUnTar],
```
BillyONeal, line 105: I'm pretty sure you still need these, they just become:

```suggestion
    ['nuget', installVcpkgExtract],
    ['unzip', installVcpkgExtract],
    ['untar', installVcpkgExtract],
```

BillyONeal (changes requested):

include/vcpkg/base/message-data.inc.h
```diff
@@ -850,6 +856,10 @@ DECLARE_MESSAGE(ErrorRequirePackagesList,
                 (),
                 "",
                 "`vcpkg install` requires a list of packages to install in classic mode.")
+DECLARE_MESSAGE(ErrorInvalidStripOption,
```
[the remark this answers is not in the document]
JavierMatosD: If 0 is given then, it just disregards the strip option.

src/vcpkg/commands.extract.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+        ExtractedArchive archive = {temp_dir,
+                                    fs.get_regular_files_recursive_lexically_proximate(temp_dir, VCPKG_LINE_INFO)};
+
+        auto mapping = strip_map(archive, strip_count);
+
+        for (const auto& file : mapping)
+        {
+            const auto& source = file.first;
+            const auto& destination = file.second;
```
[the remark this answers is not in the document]
JavierMatosD: I believe that `strip` should only remove leading directories (not files).

src/vcpkg/commands.extract.cpp
```diff
@@ -0,0 +1,191 @@
...
+    };
+
+    static ExtractionType get_extraction_type(const ParsedArguments& options)
+    {
+        static const std::unordered_map<std::string, ExtractionType> extraction_type_map = {
+            {"tar", ExtractionType::TAR},
+            {"zip", ExtractionType::ZIP},
+            {"nupkg", ExtractionType::NUPKG},
+            {"msi", ExtractionType::MSI},
+            {"exe", ExtractionType::EXE},
```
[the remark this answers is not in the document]
JavierMatosD: The purpose of those settings it to "override" detection that's based on the file extension. I.e., if you wanted to unpack a vsix file you could just say `--extract-type=zip` and it will just try to unzip it regardless of the extension.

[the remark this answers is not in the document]
JavierMatosD: I removed the `--extract-type` option. It now just falls back to cmake if it encounters a unique extension.

vicroms (commented):
Partial review

include/vcpkg/archives.h
```diff
@@ -13,16 +13,36 @@
...
     // Extract `archive` to `to_path` using `cmake_tool`. (CMake's built in tar)
     void extract_tar_cmake(const Path& cmake_tool, const Path& archive, const Path& to_path);
-    // Extract `archive` to `to_path`, deleting `to_path` first.
     void extract_archive(const Filesystem& fs,
                          const ToolCache& tools,
                          MessageSink& status_sink,
                          const Path& archive,
                          const Path& to_path);
+    // set `to_path` to `archive` contents.
+    void set_directory_to_archive_contents(const Filesystem& fs,
```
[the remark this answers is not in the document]
JavierMatosD: I renamed it to "set_directory..." because the function deletes whatever is in the target directory before extracting, essentially setting the directory to the contents of the archive.

src/vcpkg/archives.cpp
```diff
@@ -127,38 +127,77 @@ namespace
...
-                                  const ToolCache& tools,
-                                  MessageSink& status_sink,
-                                  const Path& archive,
-                                  const Path& to_path)
+}
+
+namespace vcpkg
+{
+    static ExtractionType guess_extraction_type(const Path& archive)
+
```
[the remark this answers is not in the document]
JavierMatosD: I don't know if it would be meaningful to add test cases for this function since it essentially just calls `Path::extension()`. Additionally, this isn't new. I just sucked out the logic that was previously in `extract_archive` into its own function. The logic is exactly the same. It just returns an enum. I can add the test cases if you still think it's necessary. :)

include/vcpkg/archives.h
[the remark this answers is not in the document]
vicroms: What about `set_directory_contents_to_archive_contents`?

vicroms (changes requested):
Needs a documentation PR

include/vcpkg/archives.h
```diff
@@ -13,16 +13,36 @@
 
 namespace vcpkg
 {
+    enum class ExtractionType
+    {
+        UNKNOWN,
```
BillyONeal, line 18: Normally enum class members are CamelCase. Though I'm not sure this enum should exist...

BillyONeal (changes requested):
In addition to nitpicks noted here, console progress for artifacts was damaged; you may wish to just pull from https://github.com/BillyONeal/vcpkg-tool/tree/extract_command

(I'm only Request Changes due to the console progress damage)

JavierMatosD:
> Is there a test that this correctly restores the executable bit of those platforms which support it?

I added a test to ensure the executable bits are preserved after extracting.

src/vcpkg/commands.extract.cpp
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    std::vector<std::pair<Path, Path>> get_archive_deploy_operations(const ExtractedArchive& archive,
+                                                                     StripSetting strip_setting)
+    {
+        std::vector<std::pair<Path, Path>> result;
+
+        const auto temp_dir = archive.temp_path;
+        const auto base_path = archive.base_path;
+        const auto proximate = archive.proximate_to_temp;
```
Thomas1664, line 85:
```suggestion
        const auto& temp_dir = archive.temp_path;
        const auto& base_path = archive.base_path;
        const auto& proximate = archive.proximate_to_temp;
```

Are those copies necessary?

src/vcpkg/commands.extract.cpp
```diff
@@ -0,0 +1,166 @@
...
+                first = std::find_if_not(first, last, is_slash);
+            }
+
+            if (first == last)
+            {
+                continue;
+            }
+
+            result.emplace_back(temp_dir / Path{prox_path},
+                                Path{base_path} / std::string(first, static_cast<size_t>(last - first)));
```
Thomas1664, line 109:
```suggestion
            result.emplace_back(temp_dir / prox_path,
                                base_path / std::string(first, static_cast<size_t>(last - first)));
```

`prox_path` and `base_path` are already `Path`. Maybe also consider replacing `auto` with the actual type to avoid confusion here.

src/vcpkg/commands.extract.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+        fs.remove_all(temp_dir, VCPKG_LINE_INFO);
+    }
+
+    void command_extract_and_exit(const VcpkgCmdArguments& args, const VcpkgPaths& paths)
+    {
+        auto& fs = paths.get_filesystem();
+        auto parse_args = args.parse_arguments(ExtractCommandStructure);
+        auto archive_path = Path{parse_args.command_arguments[0]};
+        auto destination_path = Path{parse_args.command_arguments[1]};
```
Thomas1664, line 147: Maybe you should also check that both paths aren't the same (and maybe also check that the archive isn't in a sub-path of the path that should be archived)?

src/vcpkg/commands.extract.cpp
```diff
@@ -0,0 +1,166 @@
...
+        2,
+        3,
+        {{}, {EXTRACT_SETTINGS}, {}},
+        nullptr,
+    };
+
+    static StripSetting get_strip_setting(const ParsedArguments& options)
+    {
+        auto iter = options.settings.find(OPTION_STRIP);
+        if (iter != options.settings.end())
```
Thomas1664, line 31:
```suggestion
        if (iter == options.settings.end())
```

Maybe consider inverting the condition to avoid this long branch?

BillyONeal (approved):

PR #1055, opened 2023-05-04
telemetry: track actions environmental variable

dan-shaw:
This will help us differentiate between different Action pipeline runs

### DIFF
--- a/src/vcpkg/vcpkgcmdarguments.cpp
+++ b/src/vcpkg/vcpkgcmdarguments.cpp
@@ -1,3 +1,4 @@
+#include <vcpkg/base/hash.h>
 #include <vcpkg/base/json.h>
 #include <vcpkg/base/system.debug.h>
 #include <vcpkg/base/system.h>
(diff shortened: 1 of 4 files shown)


src/vcpkg/vcpkgcmdarguments.cpp
```diff
@@ -677,6 +679,13 @@ namespace vcpkg
             Debug::println("Detected CI environment: ", *ci_env);
             get_global_metrics_collector().track_string(StringMetric::DetectedCiEnvironment, *ci_env);
         }
+
+        if (auto gh_repo = gh_repo_env.get())
+        {
+            auto gh_repo_hash = Hash::get_string_hash(*gh_repo, Hash::Algorithm::Sha256);
+            Debug::println("Github repo: ", gh_repo_hash);
+            get_global_metrics_collector().track_string(StringMetric::GithubRepo, gh_repo_hash);
```
vicroms, line 687: Probably better to create a `MetricsSubmission` object and call `get_global_metrics_collector()` just once at the end, like it is done in `VcpkgCmdArguments::track_feature_flag_metrics()`.

PR #1082, opened 2023-05-27
Remove `tls12-download`.

teo-tsirpanis:
Part of microsoft/vcpkg#30753.


dan-shaw:
From vcpkg team: we don't want to remove this yet, see above comment. Closing this PR for now

PR #1162, opened 2023-08-22
replaces ${VERSION} variable with correct version in generated spdx files

JulZimmermann:
Fix for https://github.com/microsoft/vcpkg/issues/28235


src/vcpkg/spdx.cpp
```diff
@@ -8,6 +8,19 @@
...
+    static StringView versionReplaceString = "${VERSION}";
+
+    auto fixedRef = ref.to_string();
+    const auto pos = fixedRef.find(versionReplaceString.data());
+
+    if(pos != std::string::npos) {
+        fixedRef.replace(pos, versionReplaceString.size(), version.data());
+    }
+
+    return fixedRef;
```
autoantwort, line 21:
```suggestion
    return Strings::replace_all(ref, "${VERSION}", version);
```

ras0219-msft (approved):
This LGTM as a patch to the existing "hack" -- the long term fix is to have the helpers properly communicate their resources back to the tool.

@JulZimmermann We'd also really appreciate it if you'd be willing to share your use cases for the SBOM/SPDX files, either here on GitHub or via email at vcpkg@microsoft.com.

Thanks!

PR #1175, opened 2023-08-31
Separate user and machine output 

autoantwort:
Fixes https://github.com/microsoft/vcpkg/issues/33444, https://github.com/microsoft/vcpkg/issues/30180 and https://github.com/microsoft/vcpkg/issues/34805

I also noticed that when using depend-info and some warnings are destroying the output


BillyONeal:
The current "everything goes to stdout" behavior is intended, not a bug. /cc @ras0219-msft

autoantwort:
At least for some commands where is destroys the format of the output it should go to stderr :)

ras0219-msft:
First, I want to agree with the problem statement that when we are issuing structured machine output, we should separate all "user messages" into stderr. Note though, this is _not_ just when we use `Color::warning` or `Color::error`!

For example, take `--x-json`. The _only_ `print` that we should send to stdout is the single, final write of the serialized json object (ignoring the case of streaming). _Every_ other print statement in the entire program is a user message.

Indeed, the vcpkg codebase emits user-facing messages 99% of the time and machine structured output only in very specific circumstances. Off the top of my head:
- `--x-json`
- `depend-info` (inc. `--dgml`)
- `fetch`
- `x-download` (Maybe? I think the only structured output of x-download is the error code)

For the vast majority of commands, all output is user-facing output and none is considered "machine readable".

As an aside, we must note the downside of splitting output streams: you lose synchronization. For example, if we emit `I'm going to load file: abc` to stdout (because it's not an "error" message) and then `error: file not found` to stderr (because it's an "error"), you would lose synchronization between the two. This implies that all non-machine output should go to the same channel.

One way to do this would be to simply make `stderr` the default print channel for the entire tool (since 99% of our print calls are non-machine-readable). Then, the handful of commands above would explicitly print to `stdout` with their machine-readable output.

However, this makes interacting with most of the vcpkg tool more painful than it needs to be; in most cases, we intentionally would print _nothing_ to stdout. I'm not sure what others experiences are, but I'm regularly annoyed by typing `curl | less` or `curl | tee foo.log` and then remembering that I have to `curl 2>&1 | less` after the pipeline fails to do what I want.

**Therefore, I think the most reasonable path forward for flags such as `--x-json` et al is to switch the "default" pipe from stdout to stderr when those flags are passed. Then, those commands can explicitly issue their machine-readable calls to stdout.**

I could see some future argument that says the default pipe should always be stderr, but I'd need to discuss that more.

autoantwort:
> Therefore, I think the most reasonable path forward for flags such as --x-json is to switch the "default" pipe from stdout to stderr when those flags are passed. Then, those commands can explicitly issue their machine-readable calls to stdout.

👍 The only downside is that this is challenging to implement the nice way, at least I have not found one. The "problem" is that there are a lot of prints before the "command selection" is done. So we have to decide which "default" pipe should be used before the normal command selection is done. I will push a solution but maybe someone has a better idea 😅 

An simpler to implement solution would be to output the machine readable output to stderr, but that can also be non intuitive, but a message that says you have to direct stderr could help.

src/vcpkg/vcpkgcmdarguments.cpp
```diff
@@ -388,16 +388,44 @@ namespace vcpkg
...
+                    }
+                }
+            }
         }
 
-        args.forwardable_arguments = args.parser.get_remaining_args();
+        if (use_stderr)
+        {
+            msg::set_default_output_stream(OutputStream::StdErr);
+        }
```
autoantwort, line 424: The ugly hack is here. But this now also works with `vcpkg depend-info 'pcl[core,visualization]' --forMat=dOt --debug > test.dot`

src/vcpkg/vcpkgcmdarguments.cpp
```diff
@@ -388,16 +388,44 @@ namespace vcpkg
         if (const auto p = args.debug_env.get()) args.debug = *p;
 
         auto maybe_command = args.parser.extract_first_command_like_arg_lowercase();
+        args.forwardable_arguments = args.parser.get_remaining_args();
+        bool use_stderr = Util::contains(args.forwardable_arguments, "--x-json");
```
BillyONeal, line 392: Instead of hard coding stuff like this, can we add a field to CommandMetadata that controls the setting?

Alternately, can we just do it at the top of the command function in question instead?

BillyONeal:
> I could see some future argument that says the default pipe should always be stderr, but I'd need to discuss that more.

I think it would be good if console parse errors go directly to stderr too. In that case we generally don't expect the program to continue so synchronization issues and concerns about accidentally causing pipelines to fail should not be concerns.

src/vcpkg/vcpkgcmdarguments.cpp
autoantwort: The problem then it is too late. There are Debug print statements before a command is selected.

autoantwort: For example here:
https://github.com/microsoft/vcpkg-tool/blob/e658895c8b03f64f4176ced5d16063a451c028ce/src/vcpkg.cpp#L279-L292
After line 283 we already have to know it output should go to stderr or stdout

autoantwort: But one could argue that we don't care if debug output breaks things

BillyONeal: I would like to resolve those as 'debug output always goes to stderr' and 'console parse failures always go to stderr' on the grounds that neither of those types of output should apply to the edge cases @ras0219-msft enumerated above; I'm asking around and will keep you posted.

BillyONeal:
@AugP @vicroms @JavierMatosD @data-queue @ras0219-msft and I discussed this today and agreed with the following outcomes:

1. We disagree with @BillyONeal 's proposal to send all debug output to stderr, but we agree that we don't mind if it is written inconsistently.
2. We agree with @BillyONeal 's proposal that all console parse failure messages should go to stderr in order to avoid the 'parse time machine' problem.

Do you mind if I morally take over this PR and make those changes?

autoantwort:
> Do you mind if I morally take over this PR and make those changes?

No you can do that :)

BillyONeal (approved):

BillyONeal:
@autoantwort  Can you confirm you are happy with the changes I just pushed?

autoantwort:
Seems that something is not working:
grafik

Maybe we also need e2e tests

autoantwort:
@BillyONeal Maybe this warning also must go to stderr by default

autoantwort:
Also 
grafik

BillyONeal:
> @BillyONeal Maybe this warning also must go to stderr by default

I agree.

> Also grafik

Hmmm I thought I got that one? Let me check again...

BillyONeal:
Ah the sendmetrics one is in vcpkg.cpp rather than vcpkgcmdarguments.cpp

BillyONeal:
OK please check again

autoantwort:
I don't know if we want to support that, but 
grafik

src/vcpkg/commands.z-preregister-telemetry.cpp
```diff
@@ -62,7 +62,9 @@ namespace vcpkg
         }
         else
         {
-            msg::println_warning(msgVcpkgSendMetricsButDisabled);
+            msg::write_unlocalized_text_to_stderr(
+                Color::warning,
+                LocalizedString::from_raw(WarningPrefix).append(msgVcpkgSendMetricsButDisabled).append_raw('\n'));
```
autoantwort, line 67: Maybe 
stderr_sink.println_warnin
```suggestion
            stderr_sink.println_warning(msgVcpkgSendMetricsButDisabled);
```

BillyONeal: I don't want to introduce use of the sinks where the polymorphism is not needed right now but I could see an argument in the future that all console output should be polymorphic like this. (But that seems way outside the scope of this PR)

PR #1211, opened 2023-09-25
Fix crash while parsing malformed manifest files.

autoantwort:
Fixes https://github.com/microsoft/vcpkg/issues/33973

### DIFF
--- a/src/vcpkg/commands.install.cpp
+++ b/src/vcpkg/commands.install.cpp
@@ -1105,6 +1105,7 @@ namespace vcpkg
             if (!maybe_manifest_scf)
             {
                 print_error_message(maybe_manifest_scf.error());
+                msg::println();
                 msg::println(msgExtendedDocumentationAtUrl, msg::url = docs::manifests_url);
                 Checks::exit_fail(VCPKG_LINE_INFO);
             }
--- a/src/vcpkg/sourceparagraph.cpp
+++ b/src/vcpkg/sourceparagraph.cpp
@@ -510,10 +510,15 @@ namespace vcpkg
             return t;
         }
 
+        static std::string map_illegal_names(std::string&& name)
+        {
+            return Json::IdentifierDeserializer::is_ident(name) ? std::move(name) : "<error>";
+        }
+
         Optional<DependencyRequestedFeature> visit_string(Json::Reader& r, StringView sv) const override
         {
             return Json::IdentifierDeserializer::instance.visit_string(r, sv).map(
-                [](std::string&& name) { return std::move(name); });
+                [](std::string&& name) { return map_illegal_names(std::move(name)); });
         }
 
         Optional<DependencyRequestedFeature> visit_object(Json::Reader& r, const Json::Object& obj) const override
@@ -523,7 +528,7 @@ namespace vcpkg
             r.required_object_field(type_name(), obj, NAME, name, Json::IdentifierDeserializer::instance);
             r.optional_object_field(obj, PLATFORM, platform, PlatformExprDeserializer::instance);
             if (name.empty()) return nullopt;
-            return DependencyRequestedFeature{std::move(name), std::move(platform)};
+            return DependencyRequestedFeature{map_illegal_names(std::move(name)), std::move(platform)};
         }
 
         const static DependencyFeatureDeserializer instance;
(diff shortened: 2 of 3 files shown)


src/vcpkg/commands.install.cpp
```diff
@@ -1105,6 +1105,7 @@ namespace vcpkg
             if (!maybe_manifest_scf)
             {
                 print_error_message(maybe_manifest_scf.error());
+                msg::println();
```
autoantwort, line 1108: Fixes
image

BillyONeal:
Seeing as `IdentifierDeserializer` doesn't return nullopt when the input isn't a valid identifier I suspect the right fix is just deleting the `check_exit`...

autoantwort:
> I suspect the right fix is just deleting the check_exit...

It was there to prevent using "core" and "default" as feature names because depending of the internal representation that are valid feature names. So this check_exit gives you fail fast errors and enforces invariant so that you don't get non obvious errors.  

> Seeing as IdentifierDeserializer doesn't return nullopt

I changed that to return an nullopt, but then the error message is:
```
error: while loading /Users/leanderSchulten/git_projekte/vcpkg/vcpkg.json:
$.dependencies[0].features[0].name (an identifier): "core" is not a valid identifier. Identifiers must be lowercase alphanumeric+hypens and not reserved (see https://learn.microsoft.com/vcpkg/users/manifests for more information)
$.dependencies[0].features[0]: mismatched type: expected a feature in a dependency
```

BillyONeal:
> It was there to prevent using "core" and "default" as feature names because depending of the internal representation that are valid feature names. So this check_exit gives you fail fast errors and enforces invariant so that you don't get non obvious errors.

Yeah I understand what the intent was, I'm just saying that we have competing intents here, and the rest of the JSON deserializer infrastructure appears to prohibit the deserialized types having semantic constraints like this.

src/vcpkg/sourceparagraph.cpp
```diff
@@ -510,10 +510,15 @@ namespace vcpkg
...
+        {
+            return Json::IdentifierDeserializer::is_ident(name) ? std::move(name) : "<error>";
+        }
+
         Optional<DependencyRequestedFeature> visit_string(Json::Reader& r, StringView sv) const override
         {
             return Json::IdentifierDeserializer::instance.visit_string(r, sv).map(
-                [](std::string&& name) { return std::move(name); });
+                [](std::string&& name) { return map_illegal_names(std::move(name)); });
         }
```
ras0219-msft, line 522: This should work exactly like `Dependency` below:

```suggestion
        Optional<DependencyRequestedFeature> visit_string(Json::Reader& r, StringView sv) const override
        {
            return Json::IdentifierDeserializer::instance.visit_string(r, sv).map(
                [&r](std::string&& name) {
                  if (!Json::IdentifierDeserializer::is_ident(name))
                    r.add_generic_error(type_name(), msg::format(msgInvalidDependency));
                  return std::move(name);
                });
        }
```

autoantwort: This code would still result in a crash since the constructor of `DependencyRequestedFeature` `check_exit` that the feature name is not "core" or "default" because there is a member `default-features` that must be used.

src/vcpkg/commands.install.cpp
BillyONeal: FYI I repeated this in https://github.com/microsoft/vcpkg-tool/pull/1219 but https://github.com/microsoft/vcpkg-tool/pull/1203 resolves it by making `print_error_message` itself add the extra newline since I found other occurrences.

src/vcpkg/sourceparagraph.cpp
BillyONeal: > This code would still result in a crash since the constructor of `DependencyRequestedFeature` `check_exit` that the feature name is not "core" or "default" because there is a member `default-features` that must be used.

The point is that things that are deserialized can't have check_exits like that.

Notably, the assert doesn't even succeed in maintaining the invariant that it won't be empty, since:

```
DependencyRequestedFeature f{"nonempty"};
auto moved = std::move(f);
assert(f.name.empty()); // passes on most standard library implementations
```

BillyONeal:
Please let me know what you think of https://github.com/microsoft/vcpkg-tool/pull/1219

BillyONeal:
Closed in favor of https://github.com/microsoft/vcpkg-tool/pull/1219 on which I marked you as a co-author for discovering the 'bones' of the fix. Thanks again!

PR #1213, opened 2023-09-26
Search for absolute paths in parallel

Thomas1664:
In #694 I proposed the use of parallel for_each for some use cases. The problem was that it was possible that the program is exited during a loop iteration.
Here I checked that functions inside the loop don't exit.

I observed that checking for absolute paths is the post build check that needs by far the longest time. This check opens almoast every non-binary file that a port installed and searches for absolute paths. This results in linear complexity. If a file isn't too big, parsing doesn't take that long but it adds up if there are multiple header files. Using parallel for_each as proposed here can therefore lead to significant speedup. The speedup increases with number and size of installed header files. I observed ~3x speedup for nlohmann-json:x64-windows (release build).

Note that `file_contains_absolute_paths` does nothing if the file is a binary. However, I considered the number of binaries low in comparison to the number of header files so I think we don't waste that much if we had a few no-op iterations.

### DIFF
--- a/include/vcpkg/base/parallel-algorithms.h
+++ b/include/vcpkg/base/parallel-algorithms.h
@@ -8,6 +8,27 @@
 
 namespace vcpkg
 {
+    template<class F>
+    inline void execute_in_parallel(size_t work_count, F&& work)
+    {
+        const size_t thread_count = static_cast<size_t>(get_concurrency());
+        const size_t num_threads = std::max(static_cast<size_t>(1), std::min(thread_count, work_count));
+
+        std::vector<std::future<void>> workers;
+        workers.reserve(num_threads - 1);
+
+        for (size_t i = 0; i < num_threads - 1; ++i)
+        {
+            workers.emplace_back(std::async(std::launch::async, [&work]() { work(); }));
+        }
+        work();
+
+        for (auto&& w : workers)
+        {
+            w.get();
+        }
+    }
+
     template<class It, class F>
     void parallel_for_each(It begin, size_t work_count, F&& cb)
     {
@@ -21,30 +42,38 @@ namespace vcpkg
             return;
         }
 
-        const size_t thread_count = static_cast<size_t>(get_concurrency());
-        const size_t num_threads = std::max(static_cast<size_t>(1), std::min(thread_count, work_count));
-
-        std::vector<std::future<void>> workers;
-        workers.reserve(num_threads - 1);
-
         std::atomic_size_t next{0};
-        auto work = [&]() {
+
+        execute_in_parallel(work_count, [&]() {
             size_t i;
             while (i = next.fetch_add(1, std::memory_order_relaxed), i < work_count)
             {
                 cb(*(begin + i));
             }
-        };
+        });
+    }
 
-        for (size_t i = 0; i < num_threads - 1; ++i)
+    template<class InIt, class OutIt, class F>
+    void parallel_transform(InIt begin, size_t work_count, OutIt out_begin, F&& cb)
+    {
+        if (work_count == 0)
         {
-            workers.emplace_back(std::async(std::launch::async, [&work]() { work(); }));
+            return;
         }
-        work();
+        if (work_count == 1)
+        {
+            *out_begin = cb(*begin);
+            return;
+        }
 
-        for (auto&& w : workers)
-        {
-            w.get();
-        }
+        std::atomic_size_t next{0};
+
+        execute_in_parallel(work_count, [&]() {
+            size_t i;
+            while (i = next.fetch_add(1, std::memory_order_relaxed), i < work_count)
+            {
+                *(out_begin + i) = cb(*(begin + i));
+            }
+        });
     }
 }
--- a/include/vcpkg/base/parallel-algorithms.h
+++ b/include/vcpkg/base/parallel-algorithms.h
@@ -9,7 +9,7 @@
 namespace vcpkg
 {
     template<class F>
-    inline void execute_in_parallel(size_t work_count, F&& work)
+    inline void execute_in_parallel(size_t work_count, F&& work) noexcept
     {
         const size_t thread_count = static_cast<size_t>(get_concurrency());
         const size_t num_threads = std::max(static_cast<size_t>(1), std::min(thread_count, work_count));
@@ -19,7 +19,7 @@ namespace vcpkg
 
         for (size_t i = 0; i < num_threads - 1; ++i)
         {
-            workers.emplace_back(std::async(std::launch::async, [&work]() { work(); }));
+            workers.emplace_back(std::async(std::launch::async | std::launch::deferred, [&work]() { work(); }));
         }
         work();
 
@@ -29,8 +29,8 @@ namespace vcpkg
         }
     }
 
-    template<class It, class F>
-    void parallel_for_each(It begin, size_t work_count, F&& cb)
+    template<class RanIt, class F>
+    void parallel_for_each_n(RanIt begin, size_t work_count, F cb) noexcept
     {
         if (work_count == 0)
         {
@@ -45,16 +45,19 @@ namespace vcpkg
         std::atomic_size_t next{0};
 
         execute_in_parallel(work_count, [&]() {
-            size_t i;
-            while (i = next.fetch_add(1, std::memory_order_relaxed), i < work_count)
+            size_t i = 0;
+            while (i < work_count)
             {
-                cb(*(begin + i));
+                if (next.compare_exchange_weak(i, i + 1, std::memory_order_relaxed))
+                {
+                    cb(*(begin + i));
+                }
             }
         });
     }
 
-    template<class InIt, class OutIt, class F>
-    void parallel_transform(InIt begin, size_t work_count, OutIt out_begin, F&& cb)
+    template<class RanItSource, class RanItTarget, class F>
+    void parallel_transform(RanItSource begin, size_t work_count, RanItTarget out_begin, F&& cb) noexcept
     {
         if (work_count == 0)
         {
@@ -69,10 +72,13 @@ namespace vcpkg
         std::atomic_size_t next{0};
 
         execute_in_parallel(work_count, [&]() {
-            size_t i;
-            while (i = next.fetch_add(1, std::memory_order_relaxed), i < work_count)
+            size_t i = 0;
+            while (i < work_count)
             {
-                *(out_begin + i) = cb(*(begin + i));
+                if (next.compare_exchange_weak(i, i + 1, std::memory_order_relaxed))
+                {
+                    *(out_begin + i) = cb(*(begin + i));
+                }
             }
         });
     }
--- a/include/vcpkg/base/parallel-algorithms.h
+++ b/include/vcpkg/base/parallel-algorithms.h
@@ -1,0 +1,66 @@
+#pragma once
+
+// At this time only MSVC supports parallel algorithms by default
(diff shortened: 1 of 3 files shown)


src/vcpkg/postbuildlint.cpp
```diff
@@ -1281,13 +1282,17 @@ namespace vcpkg
         const auto stringview_paths = Util::fmap(string_paths, [](std::string& s) { return StringView(s); });
 
         std::vector<Path> failing_files;
-        for (auto&& file : fs.get_regular_files_recursive(dir, IgnoreErrors{}))
-        {
+        std::mutex mtx;
+        auto files = fs.get_regular_files_recursive(dir, IgnoreErrors{});
+        auto work = [&](auto&& file) {
```
Thomas1664, line 1287: I get a compiler error (VS 22 17.7.4) when I replace `auto&&` with `path&&`: `cannot convert argument 1 from Path to Path&&`
 It works on GCC. Is this a STL bug?

src/vcpkg/postbuildlint.cpp
```diff
@@ -1281,13 +1282,17 @@ namespace vcpkg
...
             if (file_contains_absolute_paths(fs, file, stringview_paths))
             {
-                failing_files.push_back(file);
+                std::lock_guard lock{mtx};
+                failing_files.push_back(std::move(file));
             }
-        }
+        };
+
+        vcpkg_par_unseq_for_each(files.begin(), files.end(), work);
```
BillyONeal, line 1295: This takes a lock which is very much not `_unseq`.

include/vcpkg/base/parallel-algorithms.h
```diff
@@ -0,0 +1,66 @@
...
+            size_t i;
+            while (i = next.fetch_add(1, std::memory_order_relaxed), i < static_cast<size_t>(work_count))
+            {
+                cb(std::move(*(begin + i)));
+            }
+        };
+
+        for (size_t i = 0; i < num_threads - 1; ++i)
+        {
+            workers.emplace_back(std::async(std::launch::async, work));
```
BillyONeal, line 53: `launch::async` with `future<void>` is basically the same as using `std::thread`, I think we should do that?

autoantwort: I think the benefit of `std::async` is that it maybe uses a thread pool

include/vcpkg/base/parallel-algorithms.h
```diff
@@ -0,0 +1,66 @@
...
+        if (begin == end)
+        {
+            return;
+        }
+        if (begin + 1 == end)
+        {
+            return cb(std::move(*begin));
+        }
+
+        const auto thread_count = std::thread::hardware_concurrency() * 2;
```
autoantwort, line 34:
```suggestion
        const auto thread_count = get_concurrency();
```
You then have to include `#include <vcpkg/base/system.h>`

autoantwort:
You should probably change the implementation of `cmd_execute_and_capture_output_parallel` to reduce code duplication.

Thomas1664:
> You should probably change the implementation of `cmd_execute_and_capture_output_parallel` to reduce code duplication.

I really should. This is where I originally stole the code.

include/vcpkg/base/parallel-algorithms.h
Thomas1664: I think this doesn't make much sense because this is ignored by `std::for_each` on Windows anyway.

src/vcpkg/postbuildlint.cpp
Neumann-A: have you tried `const Path&&` ?

include/vcpkg/base/parallel-algorithms.h
Neumann-A: Hmm still needs to adjust to `VCPKG_MAX_CONCURRENCY/VCPKG_CONCURRENCY`

autoantwort: > Hmm still needs to adjust to `VCPKG_MAX_CONCURRENCY/VCPKG_CONCURRENCY`

`get_concurrency()` does that

src/vcpkg/postbuildlint.cpp
Thomas1664: Doesn't work either. It works with `const Path&` but that's not what I want. The lambda compiles to `vcpkg::check_no_absolute_paths_in'::2'::<lambda_2>::operator()<vcpkg::Path &>`

autoantwort: > Unlike the rest of the parallel algorithms, for_each is not allowed to make copies of the elements in the sequence even if they are [TriviallyCopyable](https://en.cppreference.com/w/cpp/named_req/TriviallyCopyable).

From https://en.cppreference.com/w/cpp/algorithm/for_each

include/vcpkg/base/parallel-algorithms.h
Thomas1664: If we had access to the underlying thread pool we could use [SetThreadpoolThreadMaximum](https://learn.microsoft.com/en-us/windows/win32/api/threadpoolapiset/nf-threadpoolapiset-setthreadpoolthreadmaximum).

autoantwort: Or we use the custom implementation also on windows

Neumann-A: or just use `hwloc` ;) A wait vcpkg is a package manager and cannot use other stuff and has to come up with its own solution...

include/vcpkg/base/parallel-algorithms.h
```diff
@@ -0,0 +1,50 @@
...
+#include <vcpkg/base/system.h>
+
+#include <atomic>
+#include <future>
+#include <vector>
+
+namespace vcpkg
+{
+    template<class It, class F>
+    void parallel_for_each(It begin, size_t work_count, F&& cb)
```
autoantwort, line 12: We could also add an overload for containers like in the Util namespace

Thomas1664:
> You should probably change the implementation of `cmd_execute_and_capture_output_parallel` to reduce code duplication.

`cmd_execute_and_capture_output_parallel` uses something equivalent to `std::transform`

Thomas1664:
Does `fs.read_contents(file, IgnoreErrors{})` exit in the error case? If so, can we just skip this file?

Thomas1664:
> The function name suggest no. But even if, why is this a blocker for using parallel loops?

I don't know, but Robert didn't like it [here](https://github.com/microsoft/vcpkg-tool/pull/694#discussion_r982716332):

autoantwort:
In my opinion it depends on the likelihood of the error. If the error is expected to happen (for example a malformed manifest file), the function should use `ExpectedL` to prevent interleaved console output, but when an error is unexpected (`fs.read_contents()` for a file that exists), calling Checks::check_exit inside a parallel loop is ok (also the loop does not print anything)

include/vcpkg/base/parallel-algorithms.h
BillyONeal: > I think the benefit of `std::async` is that it maybe uses a thread pool

It can't use a thread pool with `launch::async`.

autoantwort: https://github.com/microsoft/STL/blob/6735beb0c2260e325c3a4c971ec5b75427e9305f/stl/inc/future#L662
and 
> Use tasks (the [concurrency::task](https://learn.microsoft.com/en-us/cpp/parallel/concrt/reference/task-class?view=msvc-170) class) when you write asynchronous code. The task class uses the Windows ThreadPool as its scheduler, not the Concurrency Runtime.

From https://learn.microsoft.com/en-us/cpp/parallel/concrt/task-parallelism-concurrency-runtime?view=msvc-170

Thomas1664:
> Does `fs.read_contents(file, IgnoreErrors{})` exit in the error case? If so, can we just skip this file?

I confirmed that this doesn't exit by deleting all files of the port before the loop starts. We just skip those files in this case.

include/vcpkg/base/parallel-algorithms.h
BillyONeal: > https://github.com/microsoft/STL/blob/6735beb0c2260e325c3a4c971ec5b75427e9305f/stl/inc/future#L662

This is nonconforming: http://eel.is/c++draft/futures.async#3.1

and if I ever get to actually ship the rewritten `<future>` I did in 2016 this nonconformance got fixed.

include/vcpkg/base/parallel-algorithms.h
```diff
@@ -0,0 +1,79 @@
...
+        {
+            workers.emplace_back(std::async(std::launch::async, [&work]() { work(); }));
+        }
+        work();
+
+        for (auto&& w : workers)
+        {
+            w.get();
+        }
+    }
```
BillyONeal, line 30:
```suggestion
    template<class F>
    inline void execute_in_parallel(size_t work_count, F work)
    {
        if (work_count == 0) { return; }
        const size_t thread_count = static_cast<size_t>(std::thread::hardware_concurrency());
        const size_t num_threads = std::min(thread_count, work_count) - 1;

        std::vector<std::thread> workers;
        workers.reserve(num_threads);
        for (size_t i = 0; i < num_threads; ++i)
        {
            workers.emplace_back(work);
        }

        work();

        for (auto&& w : workers)
        {
            w.join();
        }
    }
```

BillyONeal:
> > The function name suggest no. But even if, why is this a blocker for using parallel loops?
> 
> I don't know, but Robert didn't like it [here](https://github.com/microsoft/vcpkg-tool/pull/694#discussion_r982716332):

`try_load_port` engages a *HUGE* amount of code that is not prepared to be called from multiple threads, expects to write directly to the console, etc. That needs careful design and synchronization and auditing of everything that might be transitively called, which was not done there.

This by contrast is much more limited.

BillyONeal (commented):
This looks mostly good to me, only nitpicks

Thomas1664:
> `try_load_port` engages a _HUGE_ amount of code that is not prepared to be called from multiple threads, expects to write directly to the console, etc. That needs careful design and synchronization and auditing of everything that might be transitively called, which was not done there.

It would already be an improvement if this function is called only once per port: https://github.com/microsoft/vcpkg/issues/34095

include/vcpkg/base/parallel-algorithms.h
```diff
@@ -0,0 +1,85 @@
...
+
+    template<class RanIt, class F>
+    void parallel_for_each_n(RanIt begin, size_t work_count, F cb) noexcept
```
autoantwort, line 33:
```suggestion
    template<class T>
    void parallel_for_each_n(View<T> view, F cb) noexcept
```

BillyONeal:
Thanks for the improvement!

BillyONeal:
This change and/or https://github.com/microsoft/vcpkg-tool/pull/1255 are almost certainly why we are seeing macos hang while building Python2:

vcpkg is hung on 2 threads

thread 1 is waiting for a future

thread 2 is in boyer moore horspool

BillyONeal:
Maybe it is https://github.com/llvm/llvm-project/issues/63822 ?

BillyONeal:
I'm giving https://github.com/microsoft/vcpkg-tool/compare/main...try-to-fix-macos-hang a shot but given that it seems to be Python2 every time I'm more inclined to blame Boyer Moore :/

PR #1216, opened 2023-09-27
vcpkg new / validate name

JavierMatosD:
vcpkg new should validate name as an identifier. 

Before:

```
vcpkg new --name HelloWorld --version 1.0
```

After:

```
vcpkg new --name HelloWorld --version 1.0

error: "HelloWorld" is not a valid identifier. Identifiers must be lowercase alphanumeric+hypens and not reserved (see https://learn.microsoft.com/vcpkg/commands/new for more information)
```

### DIFF
--- a/src/vcpkg/commands.new.cpp
+++ b/src/vcpkg/commands.new.cpp
@@ -52,7 +52,7 @@ namespace vcpkg
         nullptr,
     };
 
-    bool isValidName(const std::string& name)
+    static bool isValidName(const std::string& name)
     {
         for (char ch : name)
         {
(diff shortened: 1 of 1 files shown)


src/vcpkg/commands.new.cpp
```diff
@@ -77,6 +89,13 @@ namespace vcpkg
                 return msg::format_error(msgNewNameCannotBeEmpty);
             }
 
+            if (!isValidName(*name))
```
autoantwort, line 92: Maybe 
```suggestion
            if (!Json::IdentifierDeserializer::is_ident(*name))
```

BillyONeal (changes requested):
I agree with @autoantwort

PR #1218, opened 2023-09-28
vcpkg new / validate name

JavierMatosD:
vcpkg new should validate name as an identifier. 

Before:

```
vcpkg new --name HelloWorld --version 1.0
```

After:

```
vcpkg new --name HelloWorld --version 1.0

error: "HelloWorld" is not a valid identifier. Identifiers must be lowercase alphanumeric+hypens and not reserved (see https://learn.microsoft.com/vcpkg/commands/new for more information)
```

### DIFF
--- a/src/vcpkg/commands.new.cpp
+++ b/src/vcpkg/commands.new.cpp
@@ -1,5 +1,6 @@
 #include <vcpkg/base/checks.h>
 #include <vcpkg/base/files.h>
+#include <vcpkg/base/jsonreader.h>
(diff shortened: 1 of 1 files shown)


autoantwort:
Same as https://github.com/microsoft/vcpkg-tool/pull/1216?

JavierMatosD:
> Same as #1216?

Yeah, I used the wrong remote branch for the PR :)

BillyONeal (approved):

PR #1290, opened 2023-11-28
Make printing thread safe

Thomas1664:
This allows printing from multiple threads.

### DIFF
--- a/src/vcpkg/base/messages.cpp
+++ b/src/vcpkg/base/messages.cpp
@@ -307,9 +307,12 @@ namespace vcpkg::msg
     {
         if (sv.empty()) return;
 
+        static std::mutex mtx;
+
         if (is_console)
         {
             WORD original_color = 0;
+            std::lock_guard lck(mtx);
             if (c != Color::none)
             {
                 CONSOLE_SCREEN_BUFFER_INFO console_screen_buffer_info{};
@@ -340,6 +343,7 @@ namespace vcpkg::msg
         {
             const char* pointer = sv.data();
             ::size_t size = sv.size();
+            std::lock_guard lck(mtx);
 
             while (size != 0)
             {
@@ -365,7 +369,7 @@ namespace vcpkg::msg
         return write_unlocalized_text_impl(c, sv, stderr_handle, stderr_is_console);
     }
 #else
-    static void write_all(const char* ptr, size_t to_write, int fd)
+    static void write_all(const char* ptr, size_t to_write, int fd, std::unique_lock<std::mutex>& lck)
     {
         while (to_write != 0)
         {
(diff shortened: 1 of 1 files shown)


Thomas1664:
> That said, I'm not sure this is the correct approach. It doesn't matter how thread safe `print` is if callers are doing
> 
> ```
> print("the answer is ");
> ```

I think this is the correct approach because everything that exits (which is almost everything) prints something before the program is terminated and it is nearly impossible to wrap all instances with mutexes. If println wants to do 2 prints that should be synchronized, it needs its own mutex. I would argue that your other example simply uses print wrong because it should use format instead of printing twice.

Thomas1664:
> That said, I'm not sure this is the correct approach. It doesn't matter how thread safe `print` is if callers are doing
> 
> ```
> print("the answer is ");
> print(std::to_string(42));
> ```

I know but this isn't the point of this PR. Here, the goal really is to make single prints thread safe. However, the `Console` also allows printing multiple "buffered" strings with different colours at once. Additional PRs are required to ensure output of multiple things isn't interleaved.

autoantwort:
> I know but this isn't the point of this PR. Here, the goal really is to make single prints thread safe.

Motivation, benefits?

Thomas1664:
> > I know but this isn't the point of this PR. Here, the goal really is to make single prints thread safe.
> 
> Motivation, benefits?

Anything that prints something can't be parallelized right now. Protection with mutexes on the calling site doesn't work because prints may hide behind function calls.
This PR would allow to parallelize tasks that potentially print to console PRs like #908 and #1256 to move forward.

autoantwort:
> This PR would allow to parallelize tasks that potentially print to console, i.e. PRs like https://github.com/microsoft/vcpkg-tool/pull/908

At least not https://github.com/microsoft/vcpkg-tool/pull/908, because the printing would still be interleaved with this PR.

autoantwort:
In https://github.com/microsoft/vcpkg-tool/pull/1256 you could also use a MessageSink that internally locks before printing. But then you must still be sure that every message is printed via one print call, which you have to check manually which would be solved by https://github.com/microsoft/vcpkg-tool/pull/1323 on a type system level.

Thomas1664:
> In #1256 you could also use a MessageSink that internally locks before printing. But then you must still be sure that every message is printed via one print call, which you have to check manually which would be solved by #1323 on a type system level.

I don't like these approaches because you would have to pass this down the call stack and potentially down to things like `value_or_exit`. IMO it's easier to not worry about this at all.

Thomas1664:
> > This PR would allow to parallelize tasks that potentially print to console, i.e. PRs like #908
> 
> At least not #908, because the printing would still be interleaved with this PR.

Yeah, but again the point here is about _within_ a single print call.

PR #1348, opened 2024-02-17
Move json data

Thomas1664:


BillyONeal (commented):
I find this change very scary as we've had horrible bugs trying to move things around in the JSON DOM before. Do you have a benchmark showing this as meaningful?

Thomas1664:
> I find this change very scary as we've had horrible bugs trying to move things around in the JSON DOM before.

I don't think moving things around in the JSON DOM is scary as most of the moves are at the end of a scope so there's nothing special about them. However, I do see the argument that re-using a moved-from variable is a possible source of bugs.

> Do you have a benchmark showing this as meaningful?

I didn't benchmark this (but will try).

BillyONeal:
>I don't think moving things around in the JSON DOM is scary as most of the moves are at the end of a scope so there's nothing special about them.

The thing with DOMs is that their contents are shared around by different components. Moving something that was actually a component of an overall DOM has been a bug we have written a lot of.

>This PR improves the performance of create_dependency_graph_snapshot by 25% (105 us vs. 79 us).

What is that in the context of an overall command using this?

autoantwort:
> What is that in the context of an overall command using this?

This is only called after you build a package for seconds, minutes or hours.

talregev:
Why we are using json? Can we switch to yaml format?

Thomas1664:
> Why we are using json? Can we switch to yaml format?

I'm almost certain we can't. Due to the versioning feature, json needs to be maintained forever anyways. Changing other json files to yml would also break all users.

On the other hand, I can see an argument for using yml because it would make it a lot easier to implement [Robert's suggestion](https://github.com/microsoft/vcpkg-tool/pull/1377#discussion_r1566512751). With json, we can't just append new objects to a file, we basically need to rewrite it. That being said, I think we could in theory use yml at least for new stuff.

dg0yt:
YAML has its own set of problems. This is a what-about discussion and doesn't belong here.

src/vcpkg/commands.set-installed.cpp
```diff
@@ -55,74 +55,84 @@ namespace vcpkg
...
 
-            Json::Object snapshot;
-            snapshot.insert(JsonIdJob, job);
-            snapshot.insert(JsonIdVersion, Json::Value::integer(0));
-            snapshot.insert(JsonIdSha, Json::Value::string(*github_sha));
-            snapshot.insert(JsonIdRef, Json::Value::string(*github_ref));
-            snapshot.insert(JsonIdScanned, Json::Value::string(CTime::now_string()));
-            snapshot.insert(JsonIdDetector, detector);
+        snapshot.insert(JsonIdVersion, Json::Value::integer(0));
+        snapshot.insert(JsonIdSha, Json::Value::string(*args.github_sha.get()));
```
Thomas1664, line 72:
```suggestion
        snapshot.insert(JsonIdSha, Json::Value::string(*github_sha));
```

src/vcpkg/commands.set-installed.cpp
```diff
@@ -55,74 +55,84 @@ namespace vcpkg
...
-                {
-                    return nullopt;
-                }
-                auto spec = action.spec.to_string();
-                map.insert(
-                    {spec, fmt::format("pkg:github/vcpkg/{}@{}", spec, scfl->source_control_file->to_version())});
+                return nullopt;
             }
+            auto spec = action.spec.to_string();
+            map.emplace(std::move(spec),
```
Thomas1664, line 93:
```suggestion
            map.emplace(spec,
```

PR #1349, opened 2024-02-17
Use emplace instead of insert

Thomas1664:


BillyONeal (approved):
Most of these are node-based containers where emplace can be a pessimization, as emplace often forces allocating a node even if no new node is inserted.

However I don't think performance of any of this matters. Given that apparently these things are no longer aggregates at least this avoids extra {}s.

Thanks :)

Thomas1664:
> Most of these are node-based containers where emplace can be a pessimization, as emplace often forces allocating a node even if no new node is inserted.
> 
> However I don't think performance of any of this matters. Given that apparently these things are no longer aggregates at least this avoids extra {}s.
> 
> Thanks :)

@BillyONeal how is this a pessimization? Previously, elements were also constructed regardless of a new node was inserted.

PR #1354, opened 2024-02-26
Add quotes around the VCPKG_ROOT and Z_POWERSHELL_EXE variable path assignments.

jcbrill:
Add quotes around the VCPKG_ROOT and Z_POWERSHELL_EXE variable path assignments.

Without adding quotes around the set variable path assignments, paths that contain a close parenthesis character `)` would cause the command/batch file to exit prematurely and vs batch call chains (e.g., vcvarsall.bat) to fail.

### DIFF
--- a/vcpkg-init/vcpkg-init.ps1
+++ b/vcpkg-init/vcpkg-init.ps1
@@ -109,22 +109,22 @@ IF EXIST $null DEL $null
 
 :: Figure out where VCPKG_ROOT is
 IF EXIST "%~dp0.vcpkg-root" (
-  SET VCPKG_ROOT=%~dp0
+  SET "VCPKG_ROOT=%~dp0"
 )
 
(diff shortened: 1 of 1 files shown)


jcbrill:
For testing with VS2022:
* `vcpkg-init.ps1` was renamed to `vcpkg-init.cmd`.
* The internal paths were updated to those in the original VS2022 file.
* The signature block from the original VS022 file was added.
* The modified `vcpkg-init.cmd` file was copied into `%VSINSTALLDIR%\vc\vcpkg`.

All four assignments were tested locally with:
* VS2022 installed into `C:\Program Files (x86)`
* Powershell path of `C:\Program Files (x86)\WindowsPowerShell\v1.0`
* USERPROFILE path of `C:\Users (x86)\test`
* Running VS2022 vcvarsall.bat works as expected.

PR #1357, opened 2024-02-28
Fixed crash on non-existent port with version override

dforsten:
Fixes this issue reported in the vcpkg project:
https://github.com/microsoft/vcpkg/issues/36994

The "entry" function does not return an error but an empty RegistryEntry in some situations, needs to be checked for.

An alternative solution would be to refactor all registry implementations to never return an empty RegistryEntry, but an error instead. 
That should be done in a separate issue after this fix is merged.

### DIFF
--- a/azure-pipelines/end-to-end-tests-dir/registries.ps1
+++ b/azure-pipelines/end-to-end-tests-dir/registries.ps1
@@ -383,10 +383,9 @@ finally
     Pop-Location
 }
 
-
 # test builtin registry
 Write-Trace "test builtin registry with baseline"
-$manifestDir = "$TestingRoot/manifest"
+$manifestDir = "$TestingRoot/manifest-builtin"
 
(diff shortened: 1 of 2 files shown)


ras0219-msft:
Thank you for the fix, this looks great!

Could you also add an end-to-end test that verifies this behavior and prevents future regressions? It should hopefully be pretty straightforward to extend https://github.com/microsoft/vcpkg-tool/blob/main/azure-pipelines/end-to-end-tests-dir/registries.ps1 to cover this case.

BillyONeal:
I added such a test. I verified that the un-fixed product failed:

test failure evidence

BillyONeal (approved):

dforsten:

BillyONeal:
Thanks again for your help!

dforsten:
Glad to be of help,
thanks for creating a great tool for the C++ community!

PR #1371, opened 2024-03-24
Colorize multi-triplet dot graphs

dg0yt:
Just https://github.com/microsoft/vcpkg-tool/pull/1371/commits/8361878a4bc4dab9c5dcae6344974e12d76fd160
Depends on (for now, includes) #1369 and #1370.

Example:
~~~
vcpkg depend-info proj:arm64-linux proj tiff[core] --triplet x64-linux-dynamic --dot | dot -Tpng -oimg.png && xdg-open img.png
~~~
Output:
img

### DIFF
--- a/src/vcpkg/commands.depend-info.cpp
+++ b/src/vcpkg/commands.depend-info.cpp
@@ -173,7 +173,17 @@ namespace vcpkg
 
         for (const auto& package : depend_info)
         {
-            fmt::format_to(std::back_inserter(s), "\"{}\";\n", package.package);
+            const char* node_style = "";
+            if (Strings::ends_with(package.package, ":host"))
+            {
+                node_style = " [color=gray fontcolor=gray]";
+            }
+            else if (Strings::contains(package.package, ':'))
+            {
+                node_style = " [color=blue fontcolor=blue]";
+            }
+
+            fmt::format_to(std::back_inserter(s), "\"{}\"{};\n", package.package, node_style);
             if (package.dependencies.empty())
             {
                 empty_node_count++;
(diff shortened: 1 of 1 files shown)


autoantwort:
So non default triplet -> red; default triplet -> black; host triplet -> gray? 
Maybe use blue instead of red, red looks like an error 😅

dg0yt:
> Maybe use blue instead of red, red looks like an error 😅

Done. 
(And there is room for "error", such as "unsupported".)

BillyONeal:
I'm a bit concerned that grey is a bit too close to the background color; I'm worried about accessibility requirements, that sort of thing.

How about DimGray, HTML color \#696969 ?
or perhaps DarkSlateGray, HTML color \#2F4F4F ?

BillyONeal (approved):
I am happy with the code here pending quibbles over the specific colors and an Ok from our PM team /cc @AugP

BillyONeal:
> I'm a bit concerned that grey is a bit too close to the background color; I'm worried about accessibility requirements, that sort of thing.
> 
> How about DimGray, HTML color #696969 ? or perhaps DarkSlateGray, HTML color #2F4F4F ?

I note that on https://graphviz.org/doc/info/colors.html they use black text on `gray` and white text on `dimgray` and `darkslategray` suggesting that the latter two are more appropriate for contrast against a white background

AugP:
I have no objections to this, assuming it passes our accessibility compliance requirements. @BillyONeal can you confirm if this conforms to our accessibility requirements?

dg0yt:
After some playing with different candidates, I switched to `gray51`. This is the lightest (=most distant from black) gray which is still label with white text (=distant enough from white) on https://graphviz.org/doc/info/colors.html.

img

dg0yt:
FTR all default native requests remain B/W.

BillyONeal:
Thanks :)

PR #1378, opened 2024-04-02
Workaround for statically linking with GCC 9 and libpthread

TrueBrain:
See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=52590 for context, and https://github.com/microsoft/vcpkg/issues/37842 for more details on my analysis towards this problem.

Was first introduced with 2023-03-29 release, as compiler changed from RHEL 7.5 (GCC 4.something) to Ubuntu 16.04 (GCC 9).

Possibly there are nicer ways to fix this, but this felt appropriate. Tested it on my machines, works as expected.

### DIFF
--- a/azure-pipelines/vcpkg-linux/toolchain.cmake
+++ b/azure-pipelines/vcpkg-linux/toolchain.cmake
@@ -16,6 +16,8 @@ endfunction()
 
 add_toolchain_linker_flag("-Wl,--rpath-link=/crossrootfs/x64/lib/x86_64-linux-gnu")
 add_toolchain_linker_flag("-Wl,--rpath-link=/crossrootfs/x64/usr/lib/x86_64-linux-gnu")
+# Workaround for https://gcc.gnu.org/bugzilla/show_bug.cgi?id=52590
+set(CMAKE_EXE_LINKER_FLAGS "-Wl,--whole-archive -lpthread -Wl,--no-whole-archive")
 
 set(CMAKE_C_COMPILER /usr/bin/gcc)
 set(CMAKE_CXX_COMPILER /usr/bin/g++)


BillyONeal:
Wow, I don't think I would have found this in a million years! I'll give it a shot soon...

BillyONeal:
> Was first introduced with 2023-03-29 release, as compiler changed from RHEL 7.5 (GCC 4.something) to Ubuntu 16.04 (GCC 9).

Technically, it was GCC9 on RHEL and is GCC11 (from Mariner) now. But yeah

TrueBrain:
> Technically, it was GCC9 on RHEL and is GCC11 (from Mariner) now. But yeah

You sure about Mariner?

https://github.com/microsoft/vcpkg-tool/blob/main/azure-pipelines/vcpkg-linux/Dockerfile#L50

Says GCC 9. I ask, as I tested my fix with GCC 9, not with GCC 11 :)

BillyONeal:
> > Technically, it was GCC9 on RHEL and is GCC11 (from Mariner) now. But yeah
> 
> You sure about Mariner?
> 
> https://github.com/microsoft/vcpkg-tool/blob/main/azure-pipelines/vcpkg-linux/Dockerfile#L50 https://github.com/microsoft/vcpkg-tool/blob/main/azure-pipelines/vcpkg-linux/toolchain.cmake#L6
> 
> Both say GCC 9. I ask, as I tested my fix with GCC 9, not with GCC 11 :) Otherwise I need to retest with GCC 11 :)

We aren't using that compiler, we are using the libstdc++ it goes with though.

TrueBrain:
Ah, okay, I see what you mean. Yeah, okay, so I did build the container (via the `vcpkg-linux` `Dockerfile`) correctly :D So indeed, I tested this combination. Good!

BillyONeal:
Repeating an observation by @Osyotr here:

image

BillyONeal:
To which I replied:

I see, I think I got wires crossed between how our musl builds work and how our glibc builds worked. In the old world, we said -static-libgcc -static-libstdc++ but not -static: https://github.com/microsoft/vcpkg-tool/blob/13fd7ab03f3c3e9482a7b615ad9879df76bb1a65/azure-pipelines/signing.yml#L227

TrueBrain:
Ooh, nice catch!

Removing the `-static` part also fixes the issue, and seems to be the better solution here :) The RedHat version than has the same runtime dependencies as the Ubuntu version, and no longer crashes on thread joins on the machines I tested it on.

Osyotr:
Note: statically linked `vcpkg-muslc` executable works on glibc-based systems (tested on RHEL7).

BillyONeal:
Thanks again for finding the problem; I credited you and @Osyotr in the alternative fix and will make sure you're noted in the release notes. Thanks!

PR #1392, opened 2024-04-25
parallel operations for some binary caching providers

### LINKED ISSUE #38404: [vcpkg-tool] Parallelise binary caching operations
With a remote binary cache, the current implementations can take a notable period of time for each individual package operation. As an example, in my experience, the AWS S3 backend typically requires approx. 2-5 seconds per package. A solid chunk of this time is networking latency and other delays that can be effectively reduced by performing multiple operations in parallel (tasks/threads).

This would be a fairly simple transformation (or at least as simple as anything involving threading gets) to using `parallel_for_each` or `parallel_transform` in place of the current for loops for the caching tools/operations which support parallelisation

Crzyrndm:
https://github.com/microsoft/vcpkg/issues/38404

This adds parallel operations using `parallel_for_each` /  `parallel_transform` to `ObjectStorageProvider`. There is currently no limitation placed on the parallelisation past that imposed by the threadpool.

This impacts the following caching providers
* `GcsStorageTool`
* `AwsStorageTool`
* `CosStorageTool`

Questions
- [ ] This adds parallelisation at the provider level. Is this the right place to add it?
- [ ] Testing?
  * I couldn't find any tests for the affected caching providers and this isn't something I would normally test for anyway (implementation detail).

### DIFF
--- a/src/vcpkg/binarycaching.cpp
+++ b/src/vcpkg/binarycaching.cpp
@@ -1049,8 +1049,9 @@ namespace
         {
             if (!request.zip_path) return 0;
             const auto& zip_path = *request.zip_path.get();
-            std::atomic_size_t upload_count = 0;
-            parallel_for_each(m_prefixes, [&](std::string const& prefix) {
+            size_t upload_count = 0;
+            for (const auto& prefix : m_prefixes)
+            {
                 auto res = m_tool->upload_file(make_object_path(prefix, request.package_abi), zip_path);
                 if (res)
                 {
@@ -1060,7 +1061,7 @@ namespace
                 {
                     msg_sink.println_warning(res.error());
                 }
-            });
+            }
             return upload_count;
         }
 
--- a/src/vcpkg/binarycaching.cpp
+++ b/src/vcpkg/binarycaching.cpp
@@ -7,6 +7,7 @@
 #include <vcpkg/base/json.h>
 #include <vcpkg/base/message_sinks.h>
 #include <vcpkg/base/messages.h>
+#include <vcpkg/base/parallel-algorithms.h>
 #include <vcpkg/base/parse.h>
 #include <vcpkg/base/strings.h>
 #include <vcpkg/base/system.debug.h>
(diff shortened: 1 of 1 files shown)


Crzyrndm:

BillyONeal:
In general, you're parallelizing the download here, but we might expect most of the time this to not be CPU limited and thus not improve as a result of parallelization. Do you have a real world use case where this did end up being meaningfully faster?

Crzyrndm:
For clarity - my experience here is with an AWS S3 cache specifically
I fully agree that this is an IO limited task and therefore threading isn't exactly the best model. This was more a quick demo to show there were significant potential improvements

Taking a project I have with 10 small dependencies (download size is minimal)
`> $Env:VCPKG_BINARY_SOURCES="clear;x-aws,s3://<test bucket name>/,read"`

## AWS S3 cache miss

Using the default vcpkg release binary - querying for presence takes 9.5s
image
Using the custom vcpkg binary from this branch - querying for presence takes 1.7s
image

Note that typically I only use the cache for CI (GHA) only and the time per package can be 1-3 seconds. The bucket region may play a significant part in that but I havn't got around to testing yet

## AWS S3 cache hit

Using the default release binary - query and download takes 23s
image
Using the custom vcpkg binary - query and download takes 4.3s
image

So approx  a 5x improvement for both query and download and that should be fairly proportional to the total number of dependencies (if this was done correctly). Writing back doesn't scale the same way because it's done after each package is built so we just ignore that for the moment

BillyONeal:
I agree a real world 5x improvement is worth opening a can of worms for; OK

Crzyrndm:
Thoughts on where to go with this?
If you've got an example of parallelisation of shell commands that would be more appropriate than the quick and dirty `parallel_*` I used here I could take a whack at that. The data flow does lend itself to parallelisation well and the CLI tools for AWS/GCP are setup for multi-process operations to my knowledge so the correctness _probably_ isn't as much of an issue as it could be (famous last words...)

Crzyrndm:
`cmd_execute_and_capture_output_parallel` is doing pretty much the same thing I'm doing here (would have to significantly reshuffle the AWS/GCP/etc. API to use that directly. The minimal form is to use the parallel functions to replace the loop as has been done.

* Added a comment to the `stat` and `download_file` declarations noting the need for the impl to be thread safe
* As far as I can tell, all operations within the parallel block either do not modify state or operate on data that is intended for the specific package

src/vcpkg/binarycaching.cpp
```diff
@@ -988,38 +993,36 @@ namespace
...
+                    const auto& abi = plan->package_abi().value_or_exit(VCPKG_LINE_INFO);
+                    auto tmp = make_temp_archive_path(m_buildtrees, plan->spec);
+                    auto res = m_tool->download_file(make_object_path(m_prefix, abi), tmp);
+                    if (res)
+                    {
+                        return ZipResource{ZipResource(std::move(tmp), RemoveWhen::always)};
+                    }
+                    else
+                    {
+                        out_sink.println_warning(res.error());
```
Crzyrndm, line 1007: Missed prints not being synchronised currently in previous assessment.
* https://github.com/microsoft/vcpkg-tool/pull/1290
* https://github.com/microsoft/vcpkg-tool/pull/1323

 This particular print could be extracted relatively easily and is the only print inside the parallel execution.

BillyONeal:
Thanks for your submission, I'm sorry at the time we didn't have the infrastructure to make this easy to do. We now have that so I've effectively implemented this for you in https://github.com/microsoft/vcpkg-tool/pull/2088 ; would you be willing to review over there? Thanks!

PR #1414, opened 2024-05-28
Fix x-update-baseline to use reference field

data-queue:
Invoking `x-update-baseline` incorrectly updates the baseline field when the registry is configured with a `reference`. The correct behavior is to use the latest commit of the reference branch when set.

See microsoft/vcpkg#38701


BillyONeal (commented):
Can you add a test?

BillyONeal (approved):

PR #1436, opened 2024-06-19
end-to-end-tests: Backup/restore GITHUB_ACTIONS

dg0yt:
Fixes grouping in GHA output.


dg0yt:
Example (before):
Bildschirmfoto vom 2024-06-19 09-28-38
Should continue with `[11/46]` etc.

BillyONeal:
Awesome, thank you!

PR #1457, opened 2024-07-16
Update fmt to v11.0.2

tchaikov:
this changeset fixes the build with fmt 11:

- prevent fmt from formatting StringView as a sequence
- include fmt/ranges.h for using fmt::join()


tchaikov:

tchaikov:
closing. it breaks the build with fmt 10.

tchaikov:
@WangWeiLin-MV in case you are interested.

BillyONeal (approved):
Thanks!

PR #1473, opened 2024-08-08
[upgrade] Don't crash on non-installed ports

Thomas1664:
Fixes https://github.com/microsoft/vcpkg/issues/21117

output before:

```console
error: The following packages are not installed:
  azure-storage-cpp:x64-windows
internal error: D:\a\_work\1\s\src\vcpkg\commands.upgrade.cpp(183): vcpkg has crashed; no additional details are available.
Please open an issue at https://github.com/microsoft/vcpkg/issues/new?template=other-type-of-bug-report.md&labels=category:vcpkg-bug with detailed steps to reproduce the problem.
```

output after:

```console
error: The following packages are not installed:
  azure-storage-cpp:x64-windows
```


GitMensch:
Not sure, just wondering: Would it be useful to issue a warning, then install?

(the goal would be a command that ensures you have the most current version, no matter if an old is installed or not)

BillyONeal:
@GitMensch I believe `install --recurse` has the behavior you want?

BillyONeal:
Ah, nope :/

BillyONeal (commented):
I'm not sure I like this resolution because it's undoing what appears to be a very explicit attempt to emit all possible errors rather than stopping at the first error. Making a competing solution shortly...

BillyONeal:
image

Ouch it's even more broken

Thomas1664:
> Ouch it's even more broken

I think that's a different issue.

> I'm not sure I like this resolution because it's undoing what appears to be a very explicit attempt to emit all possible errors rather than stopping at the first error.

I restored emitting all errors.

BillyONeal:
Thanks!

PR #1476, opened 2024-08-12
get_modified_clean_environment: Make it possible to keep vars by prefix

autoantwort:


cenit:
do you confirm that the new function makes visible all the env variables matching with CUDA_PATH?
i mean, if on the system we have defined CUDA_PATH, CUDA_PATH_V11_2 and CUDA_PATH_V10_3 all of them are visible as before right?

autoantwort:
Should be. This probably needs end to end tests

BillyONeal (approved):

Neumann-A:
This broke my builds for some reason.

PR #1479, opened 2024-08-19
Suggest running `git pull` when a version is not found in the database.

teo-tsirpanis:
Might help diagnosing failures caused by an updated `builtin-baseline` and an outdated Git registry.

### DIFF
--- a/include/vcpkg/base/message-data.inc.h
+++ b/include/vcpkg/base/message-data.inc.h
@@ -413,7 +413,10 @@ DECLARE_MESSAGE(
     "build system. To suppress this message, add set(VCPKG_POLICY_SKIP_ARCHITECTURE_CHECK enabled)")
 DECLARE_MESSAGE(ChecksFailedCheck, (), "", "vcpkg has crashed; no additional details are available.")
 DECLARE_MESSAGE(ChecksUnreachableCode, (), "", "unreachable code was reached")
-DECLARE_MESSAGE(ChecksUpdateVcpkg, (), "", "updating vcpkg by rerunning bootstrap-vcpkg or running `git pull` may resolve this failure.")
+DECLARE_MESSAGE(ChecksUpdateVcpkg,
+                (),
+                "",
+                "updating vcpkg by rerunning `git pull` and bootstrap-vcpkg may resolve this failure.")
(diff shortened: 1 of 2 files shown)


include/vcpkg/base/message-data.inc.h
```diff
@@ -413,7 +413,7 @@ DECLARE_MESSAGE(
     "build system. To suppress this message, add set(VCPKG_POLICY_SKIP_ARCHITECTURE_CHECK enabled)")
 DECLARE_MESSAGE(ChecksFailedCheck, (), "", "vcpkg has crashed; no additional details are available.")
 DECLARE_MESSAGE(ChecksUnreachableCode, (), "", "unreachable code was reached")
-DECLARE_MESSAGE(ChecksUpdateVcpkg, (), "", "updating vcpkg by rerunning bootstrap-vcpkg may resolve this failure.")
+DECLARE_MESSAGE(ChecksUpdateVcpkg, (), "", "updating vcpkg by rerunning bootstrap-vcpkg or running `git pull` may resolve this failure.")
```
autoantwort, line 416:
```suggestion
DECLARE_MESSAGE(ChecksUpdateVcpkg, (), "", "updating vcpkg by rerunning `git pull` and bootstrap-vcpkg may resolve this failure.")
```

teo-tsirpanis: Done.

BillyONeal (changes requested):
Thanks for your contribution to vcpkg! Unfortunately I think there are some serious problems with this change:

1. The `ChecksUpdateVcpkg` message is used for several cases that are not 'baseline sha nonexistent'. In fact, the majority of cases where this is used won't be helped by a `git pull`. (For what it's worth, I strongly dislike the current message and have been campaigning to remove it entirely, but at least bootstrap-vcpkg won't accidentally change unrelated git repos and stuff)
2. The advice won't work if the current working directory isn't VCPKG_ROOT, or if the vcpkg copy is the one-liner install, or the one shipping with Visual Studio.
3. This won't do anything if the user isn't using the builtin registry.

I think this message needs to only be emitted when it would do something and needs tests.

teo-tsirpanis:
Thanks for the feedback. Let's keep the message as-is.

PR #1485, opened 2024-09-02
Parallelize ObjectStorageProvider downloads

petamas:
After starting to use Boost from vcpkg in our project, the number of used ports jumped from 20 to ~100. Similarly, restoring all packages from an AWS S3 backend jumped from 2 minutes to almos 10 minutes, which is way too long for us.

Since AWS has a rate limit on individual file downloads, but there's no limit on downloading files in parallel, I've decided to parallelize the download process on the `ObjectStorageProvider` level.

I hope this doesn't break Google Cloud Storage and Tencent Cloud Object Storage backends, but I have no way of testing them. If needed, I can add a `supports_parallel_downloads()` method to `IObjectStorageTool`, and constrain the parallelization to AWS.

### DIFF
--- a/src/vcpkg/binarycaching.cpp
+++ b/src/vcpkg/binarycaching.cpp
@@ -7,6 +7,7 @@
 #include <vcpkg/base/json.h>
 #include <vcpkg/base/message_sinks.h>
 #include <vcpkg/base/messages.h>
+#include <vcpkg/base/parallel-algorithms.h>
 #include <vcpkg/base/parse.h>
 #include <vcpkg/base/strings.h>
 #include <vcpkg/base/system.debug.h>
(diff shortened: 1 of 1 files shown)


petamas:
You're right, it's basically the same. I've also experienced about 5x speedup using this parallelization, similar to @Crzyrndm.

However, if [this comment](https://github.com/microsoft/vcpkg-tool/pull/1392#discussion_r1630364582) is right, i.e. printing is not thread-safe, then my solution has the same issue. Unfortunately, the fix is not trivial, because extracting `out_sink.println_warning()` is not enough: the underlying `cmd_execute_and_capture_output` may also print in case `--debug` is passed. :-/

BillyONeal:
Thanks for your submission, I'm sorry at the time we didn't have the infrastructure to make this easy to do. We now have that so I've effectively implemented this for you in https://github.com/microsoft/vcpkg-tool/pull/2088 ; would you be willing to review over there? Thanks!

PR #1532, opened 2024-10-25
fix: repest to VCPKG_CMAKE_SYSTEM_VERSION.

thegoodtgg:
According to [VCPKG_CMAKE_SYSTEM_VERSION not work. #41723](https://github.com/microsoft/vcpkg/issues/41723), pass value of VCPKG_CMAKE_SYSTEM_VERSION as target to "vcvarsall.bat"

### DIFF
--- a/src/vcpkg/commands.build.cpp
+++ b/src/vcpkg/commands.build.cpp
@@ -311,10 +311,10 @@ namespace vcpkg
     }
 
 #if defined(_WIN32)
-    static ZStringView to_vcvarsall_target(StringView cmake_system_name)
+    static ZStringView to_vcvarsall_target(StringView cmake_system_name, StringView cmake_system_version)
     {
-        if (cmake_system_name.empty()) return "";
-        if (cmake_system_name == "Windows") return "";
+        if (cmake_system_name.empty()) return { cmake_system_version.data(), cmake_system_version.size() };
+        if (cmake_system_name == "Windows") return { cmake_system_version.data(), cmake_system_version.size()};
         if (cmake_system_name == "WindowsStore") return "store";
 
         Checks::msg_exit_with_error(VCPKG_LINE_INFO, msgUnsupportedSystemName, msg::system_name = cmake_system_name);
@@ -597,7 +597,7 @@ namespace vcpkg
         }
 
         const auto arch = to_vcvarsall_toolchain(pre_build_info.target_architecture, toolset, pre_build_info.triplet);
-        const auto target = to_vcvarsall_target(pre_build_info.cmake_system_name);
+        const auto target = to_vcvarsall_target(pre_build_info.cmake_system_name, pre_build_info.cmake_system_version);
 
         return vcpkg::Command{"cmd"}.string_arg("/d").string_arg("/c").raw_arg(
             fmt::format(R"("{}" {} {} {} {} 2>&1 <NUL)",


BillyONeal:
The documentation for `VCPKG_CMAKE_SYSTEM_VERSION` is that we pass it to CMake as `CMAKE_SYSTEM_VERSION`. Can you explain the relationship between `CMAKE_SYSTEM_VERSION` and `vcvarsall.bat` that suggests this is the correct outcome?

Thanks for your contribution!

thegoodtgg:
Yes. I see the desc about `VCPKG_CMAKE_SYSTEM_VERSION` from documentation.
When using with "ninja" generator, AFAIK，`CMAKE_SYSTEM_VERSION` make no sense. People who want using a different "Windows SDK Version", have to specify a target explicitly to `vcvarsall.bat`.
And when using CMake with "Visual Studio" generator, specify `CMAKE_SYSTEM_VERSION`  would definitely change the "Windows SDK Version" the proj using.
To make vcpkg build packages using a different version of "Windows SDK", `VCPKG_CMAKE_SYSTEM_VERSION` maybe a choice.

src/vcpkg/commands.build.cpp
```diff
@@ -311,10 +311,10 @@ namespace vcpkg
...
 
 #if defined(_WIN32)
-    static ZStringView to_vcvarsall_target(StringView cmake_system_name)
+    static ZStringView to_vcvarsall_target(StringView cmake_system_name, StringView cmake_system_version)
     {
-        if (cmake_system_name.empty()) return "";
-        if (cmake_system_name == "Windows") return "";
+        if (cmake_system_name.empty()) return {cmake_system_version.data(), cmake_system_version.size()};
+        if (cmake_system_name == "Windows") return {cmake_system_version.data(), cmake_system_version.size()};
         if (cmake_system_name == "WindowsStore") return "store";
```
BillyONeal, line 318: I have not analyzed everything this code is doing, but returning "store" in one path and returning a version number in another path is almost certainly wrong.

Moreover, this does not meet the ZStringView requirements because it's returning a pointer that it claims is null terminated with inputs that are not null terminated.

src/vcpkg/commands.build.cpp
```diff
@@ -597,7 +597,7 @@ namespace vcpkg
         }
 
         const auto arch = to_vcvarsall_toolchain(pre_build_info.target_architecture, toolset, pre_build_info.triplet);
-        const auto target = to_vcvarsall_target(pre_build_info.cmake_system_name);
+        const auto target = to_vcvarsall_target(pre_build_info.cmake_system_name, pre_build_info.cmake_system_version);
```
BillyONeal, line 600: If you want this to select the Windows SDK, I believe you need something like this instead?

```c++
auto vcvarsall_options = toolset.vcvarsall_options;
if (!pre_build_info.cmake_system_version.empty()) {
    vcvarsall_options.push_back(fmt::format("-winsdk={}", pre_build_info.cmake_system_version));
}
```

based on

vcvarsall help text

BillyONeal (changes requested):
This 'request changes' is for the ZStringView out of contract use. I'm going to confirm with other maintainers if the overall change seems correct but it seems reasonable to me given the documentation page you mentioned.

src/vcpkg/commands.build.cpp
thegoodtgg: I think "vcvarsall.bat" is enough. And I have test it locally.
Using Visual Studio 2022:
image
Using Visual Studio 2019:
image
I believe on the old or new version of visual studio, "vcvarsall.bat" can achieve the same function. 
Your suggestion about using "Toolset::vcvarsall_options" is more reasonable.

thegoodtgg:
Windows SDK Version doesn't belong to msvc toolset. And as positional arguments, "winsdk_version" should be placed after "platform_type". So using ”vcvarsall_options“ maybe not be a good choice. 
```C++
        const auto arch = to_vcvarsall_toolchain(pre_build_info.target_architecture, toolset, pre_build_info.triplet);
        const auto target = to_vcvarsall_target(pre_build_info.cmake_system_name);
        // New function
        const auto winsdk = to_vcvarsall_winsdk(pre_build_info.cmake_system_version);

        return vcpkg::Command{"cmd"}.string_arg("/d").string_arg("/c").raw_arg(
            fmt::format(R"("{}" {} {} {} {} {} 2>&1 <NUL)",
                        toolset.vcvarsall,
                        Strings::join(" ", toolset.vcvarsall_options),
                        arch,
                        target,
                        winsdk,
                        tonull));
```

BillyONeal:
>Windows SDK Version doesn't belong to msvc toolset.

Getting it into the environment does belong to `vcvarsall.bat` though.

>And as positional arguments, "winsdk_version" should be placed after "platform_type".

Sure, I apologize for being unclear. I was posting that as a 'for example', the main point being that this needs to go into a `-winsdk={}` parameter instead of listed as a plain version number on `vcvarsall.bat`'s command line.

> ```c++
> // New function
> const auto winsdk = to_vcvarsall_winsdk(pre_build_info.cmake_system_version);
> ```

Sure, separating into a separate function could be fine. I don't think it's required though.

thegoodtgg:
If we don't check the specified version of windows sdk  installed, we could use
```
const auto winsdk = pre_build_info.cmake_system_version;
```

lygstate:
It's a good things, can this be merged?

BillyONeal:
> It's a good things, can this be merged?

It needs to be fixed to actually work before it can actually merge.

Closing due to no resolution of code review comments in > 6 months.

We would be happy to accept a different PR that implements the *effect* of this change provided:

1. Documentation links, or a demonstration, that upstream CMake uses `CMAKE_SYSTEM_VERSION` to set the Windows SDK version.
2. Selecting the Windows SDK version using `-winsdk` passed to vcvars et al.

PR #1553, opened 2024-12-13
Change tool metadata file format to JSON

vicroms:
Continuation of #1490
Related: #1277
 
This PR contains several changes:

* Moves `vcpkgTools.xml` data into a JSON file and removes the XML parsing code.
* Adds the architecture field to the tool metadata.
* Removes the requirement of force system binaries on arm64 Linux platforms.


src/vcpkg/tools.cpp
```diff
@@ -62,82 +64,115 @@ namespace vcpkg
...
-        return parse_tool_data_from_xml(XML, XML_PATH, tool, "openbsd");
+        auto data = get_raw_tool_data(tool_data_table, tool, hp, "openbsd");
 #else
         return nullopt;
 #endif
-    }
-
-    Optional<ToolData> parse_tool_data_from_xml(StringView XML, StringView XML_PATH, StringView tool, StringView os)
-    {
-        static const char* XML_VERSION = "2";
```
vicroms, line 84: New file format is missing this document version field. Is this a feature we want to preserve?

src/vcpkg/tools.cpp
```diff
@@ -62,82 +63,114 @@ namespace vcpkg
...
+        {
+            return r.join();
+        }
+
+        if (auto tool_data = maybe_tool_data.get())
+        {
+            return *tool_data;
+        }
+
+        return msg::format(msgErrorWhileParsingToolData, msg::path = origin);
```
BillyONeal, line 87:
```suggestion
        return maybe_tool_data.value_or_exit(VCPKG_LINE_INFO);
```

Any errors in parsing should have been reported by `Json::Reader`, so if we get here vcpkg has a bug and we shouldn't have separate localized messages for bugs.

src/vcpkg/tools.cpp
```diff
@@ -62,82 +63,114 @@ namespace vcpkg
...
 
-    static Optional<ToolData> parse_tool_data_from_xml(StringView XML, StringView XML_PATH, StringView tool)
+    ExpectedL<std::vector<ToolDataEntry>> parse_tool_data(StringView contents, StringView origin)
     {
+        auto as_json = Json::parse(contents, origin);
+        if (!as_json)
+        {
+            return as_json.error();
+        }
+        auto as_value = std::move(as_json).value(VCPKG_LINE_INFO).value;
```
BillyONeal, line 73: This should have error handling like that in `Json::parse_object` rather than crashing if it can't be interpreted as a value here.

src/vcpkg/tools.cpp
```diff
@@ -62,82 +63,114 @@ namespace vcpkg
...
+    static ExpectedL<std::vector<ToolDataEntry>> parse_tool_data_file(const Filesystem& fs, Path path)
+    {
+        std::error_code ec;
+        auto contents = fs.read_contents(path, ec);
+        if (ec)
+        {
+            return format_filesystem_call_error(ec, __func__, {path});
+        }
+
+        return parse_tool_data(contents, path);
```
BillyONeal, line 99:
```suggestion
        return fs.try_read_contents(path).then([](FileContents&& fc) -> ExpectedL<std::vector<ToolDataEntry>> {
            return parse_tool_data(fc.content, Path{fc.origin});
        });
```

src/vcpkg/tools.cpp
```diff
@@ -1006,11 +1043,112 @@ namespace vcpkg
...
+
+            r.required_object_field(type_name(), obj, "name", value.tool, Json::UntypedStringDeserializer::instance);
+            r.required_object_field(type_name(), obj, "os", value.os, Json::UntypedStringDeserializer::instance);
+            r.required_object_field(
+                type_name(), obj, "version", value.version, Json::UntypedStringDeserializer::instance);
+
+            std::string arch_str;
+            if (r.optional_object_field(obj, "arch", arch_str, Json::ArchitectureDeserializer::instance))
+            {
+                auto it = arch_map.find(arch_str);
```
BillyONeal, line 1100: Can you use `to_cpu_architecture` instead of a second duplicate table?

BillyONeal (approved):
Thanks for the feature!

azure-pipelines/end-to-end-tests-dir/fetch.ps1
```diff
@@ -6,90 +6,107 @@ if (-not $IsMacOS -and -not $IsLinux) {
...
+        "archive": "cmake-3.22.2-macos-universal.tar.gz"
+    },
+    {
+        "name": "cmake",
+        "os": "linux",
+        "version": "3.22.2",
+        "executable": "cmake-3.22.2-linux-x86_64/bin/cmake",
+        "url": "https://github.com/Kitware/CMake/releases/download/v3.22.2/cmake-3.22.2-linux-x86_64.tar.gz",
+        "sha512": "579e08b086f6903ef063697fca1dc2692f68a7341dd35998990b772b4221cdb5b1deecfa73bad9d46817ef09e58882b2adff9d64f959c01002c11448a878746b",
+        "archive": "cmake-3.22.2linux-x86_64.tar.gz"
```
autoantwort, line 67: Since this entry has no architecture entry this version would be used on arm64 devices. Maybe the arch field should be mandatory.

autoantwort (commented):
Can you also add a json schema for this file?

src/vcpkg-test/tools.cpp
```diff
@@ -140,3 +67,204 @@ TEST_CASE ("extract_prefixed_nonquote", "[tools]")
...
+        "os": "linux",
+        "version": "2.7.4",
+        "arch": "notanarchitecture"
+    }]
+})",
+                                        "invalid_arch.json");
+    REQUIRE(!invalid_arch.has_value());
+    CHECK("invalid_arch.json: error: $.tools[0].arch (a CPU architecture): Invalid architecture: notanarchitecture. "
+          "Expected "
+          "one of: x86,x64,arm,arm64,arm64ec,s390x,ppc64le,riscv32,riscv64,loongarch32,loongarch64,mips64\n"
```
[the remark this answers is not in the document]
dg0yt:
```suggestion
          "one of: x86, x64, arm, arm64, arm64ec, s390x, ppc64le, riscv32, riscv64, loongarch32, loongarch64, mips64\n"
```

azure-pipelines/end-to-end-tests-dir/fetch.ps1
BillyONeal: I think it's reasonable for there to be arch-less entries like scripts blobs or similar. A big reason to do this work *is* to enable arm64, but truly doing well there is going to imply minting arm64 binaries and similar; fully getting that on sounds like a separate feature though.

autoantwort: Ok i hadn't thought about scripts, so one must simply be careful when adding entries

Neumann-A: (or user "arch" : "neutral" as an indicator for stuff which does not dependent on arch similar to how the vs channel manifest works)

src/vcpkg/vcpkgcmdarguments.cpp
```diff
@@ -358,6 +358,7 @@ namespace vcpkg
         args.parser.parse_option(
             SwitchBuiltinRegistryVersionsDir, StabilityTag::Experimental, args.builtin_registry_versions_dir);
         args.parser.parse_option(SwitchRegistriesCache, StabilityTag::Experimental, args.registries_cache_dir);
+        args.parser.parse_option(SwitchToolDataFile, StabilityTag::ImplementationDetail, args.tools_data_file);
```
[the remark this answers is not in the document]
BillyONeal: Actually it's z-, no docs needed.

BillyONeal:
>To do:
>Update Ninja version to one with ARM64 support.

I don't believe this is needed to land this work.

>Test with full registry build.

I don't believe this is needed because all this stuff is already tested by the fetch tests.

azure-pipelines/end-to-end-tests-dir/fetch.ps1
BillyONeal: I don't think an explicit 'neutral' value is substantially clearer than just not having the field, and would be 'annoying' as we couldn't reuse the existing `CPUArchitecture` enum and friends.

BillyONeal:
Interestingly, finally breaking the `<regex>` dependency reduced size slightly, even with all this extra JSON parsing code:

image

PR #1555, opened 2024-12-19
Change subprocesses to run at normal priority (#42715)

Binary-Song:
Fixes https://github.com/microsoft/vcpkg/issues/42715

The priority was changed to NORMAL_PRIORITY_CLASS so vcpkg can perform its tasks (building, zipping, etc) much faster.

### DIFF
--- a/include/vcpkg/base/contractual-constants.h
+++ b/include/vcpkg/base/contractual-constants.h
@@ -546,6 +546,7 @@ namespace vcpkg
     inline constexpr StringLiteral EnvironmentVariableVcpkgForceSystemBinaries = "VCPKG_FORCE_SYSTEM_BINARIES";
     inline constexpr StringLiteral EnvironmentVariableVcpkgKeepEnvVars = "VCPKG_KEEP_ENV_VARS";
     inline constexpr StringLiteral EnvironmentVariableVcpkgMaxConcurrency = "VCPKG_MAX_CONCURRENCY";
+    inline constexpr StringLiteral EnvironmentVariableVcpkgSubprocessPriority = "VCPKG_SUBPROCESS_PRIORITY";
     inline constexpr StringLiteral EnvironmentVariableVcpkgNoCi = "VCPKG_NO_CI";
     inline constexpr StringLiteral EnvironmentVariableVcpkgNuGetRepository = "VCPKG_NUGET_REPOSITORY";
     inline constexpr StringLiteral EnvironmentVariableVcpkgOverlayPorts = "VCPKG_OVERLAY_PORTS";
--- a/include/vcpkg/base/message-data.inc.h
+++ b/include/vcpkg/base/message-data.inc.h
@@ -1147,6 +1147,11 @@ DECLARE_MESSAGE(EnvInvalidMaxConcurrency,
                 (msg::env_var, msg::value),
                 "{value} is the invalid value of an environment variable",
                 "{env_var} is {value}, must be > 0")
+DECLARE_MESSAGE(EnvInvalidPriority,
+                (msg::env_var, msg::value),
+                "'{value} is a user-supplied value for {env_var} environment variable.'",
+                "invalid value \"{value}\" for {env_var}."
+                "Valid values are '', 'IDLE', 'BELOW_NORMAL', 'NORMAL', 'ABOVE_NORMAL', 'HIGH' and 'REALTIME'")
 DECLARE_MESSAGE(EnvStrFailedToExtract, (), "", "could not expand the environment string:")
 DECLARE_MESSAGE(EnvPlatformNotSupported, (), "", "Build environment commands are not supported on this platform")
 DECLARE_MESSAGE(EnvVarMustBeAbsolutePath, (msg::path, msg::env_var), "", "{env_var} ({path}) was not an absolute path")
--- a/include/vcpkg/base/system.h
+++ b/include/vcpkg/base/system.h
@@ -62,6 +62,10 @@ namespace vcpkg
     const Optional<Path>& get_program_files_platform_bitness();
 
     unsigned int get_concurrency();
+    
+#if defined(_WIN32)
+    DWORD get_subprocess_priority();
+#endif
 
     Optional<CPUArchitecture> guess_visual_studio_prompt_target_architecture();
 }
--- a/include/vcpkg/base/system.h
+++ b/include/vcpkg/base/system.h
@@ -62,7 +62,7 @@ namespace vcpkg
     const Optional<Path>& get_program_files_platform_bitness();
 
     unsigned int get_concurrency();
-    
+
 #if defined(_WIN32)
     DWORD get_subprocess_priority();
 #endif
--- a/locales/messages.json
+++ b/locales/messages.json
@@ -676,6 +676,8 @@
   "EndOfStringInCodeUnit": "found end of string in middle of code point",
   "EnvInvalidMaxConcurrency": "{env_var} is {value}, must be > 0",
   "_EnvInvalidMaxConcurrency.comment": "{value} is the invalid value of an environment variable An example of {env_var} is VCPKG_DEFAULT_TRIPLET.",
+  "EnvInvalidPriority": "invalid value \"{value}\" for {env_var}.Valid values are '', 'IDLE', 'BELOW_NORMAL', 'NORMAL', 'ABOVE_NORMAL', 'HIGH' and 'REALTIME'",
+  "_EnvInvalidPriority.comment": "'{value} is a user-supplied value for {env_var} environment variable.' An example of {env_var} is VCPKG_DEFAULT_TRIPLET.",
   "EnvPlatformNotSupported": "Build environment commands are not supported on this platform",
   "EnvStrFailedToExtract": "could not expand the environment string:",
   "EnvVarMustBeAbsolutePath": "{env_var} ({path}) was not an absolute path",
--- a/src/vcpkg/base/system.cpp
+++ b/src/vcpkg/base/system.cpp
@@ -733,6 +733,53 @@ namespace vcpkg
         return concurrency;
     }
 
+#if defined(_WIN32)
+    DWORD get_subprocess_priority()
+    {
+        static const DWORD priority = []() {
+            const auto opt_user_defined_priority = get_environment_variable(EnvironmentVariableVcpkgSubprocessPriority);
+            if (!opt_user_defined_priority.has_value())
+            {
+                return IDLE_PRIORITY_CLASS;
+            }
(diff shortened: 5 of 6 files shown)


dg0yt:
I expected only a small effect in vcpkg-tool CI, but ... I see that Windows end-to-end test for this PR took only 15 min, vs. 25 min in previous runs.

FTR this might increase memory pressure, with much more active cores as described in https://github.com/microsoft/vcpkg/issues/42715.

autoantwort:
My experience is that ci times are not a reliable measurement because you never know under which conditions it was run. So local benchmarks would be more reliable.

dg0yt:
True, but these test aren't massive workloads anyways.

Binary-Song:
> FTR this might increase memory pressure, with much more active cores as described in [microsoft/vcpkg#42715](https://github.com/microsoft/vcpkg/issues/42715).

That is true, but this can be solved by adjusting VCPKG_MAX_CONCURRENCY. On the other hand, there is currently no way of telling vcpkg to run build tools at normal priority, which is a huge waste on CI resources. In comparison, tools like cmake/ninja always build at normal priority, letting the user figure out how many cores to use without running out of mem, so I think it makes sense for vcpkg to do the same. 

On a side note, zipping (the post build packaging phase) was ridiculously slow for llvm when run in low priority, that may be a 30 min vs 5 min difference (I will probably submit a benchmark when I have time).

BillyONeal:
This was done intentionally in https://github.com/ras0219-msft/vcpkg/commit/a24ccdfc01aba83b833af528b8f1a6ec6db020a4 

@ras0219-msft Do you remember why this was set to IDLE_PRIORITY_CLASS? That seems wrong. I could imagine below normal but not idle. This might explain why we have seen such variable perf in CI.

BillyONeal:
Is the problem fixed if you use BELOW_NORMAL rather than NORMAL? I can see an argument of not wanting to make interactive stuff on the system stuttery during a build being a reason to choose lower than normal, but idle may be triggering other forms of throttling like you describe in your issue filing.

Binary-Song:
Unfortunately no. With VCPKG_MAX_CONCURRENCY set to 24, BELOW_NORMAL makes the compilers occupy around 10 cores, while NORMAL makes them occupy all 24 cores. I just changed the priority of Ninja.exe while it is running, and the effect was instant.

BELOW_NORMAL:
image
image
NORMAL:
image
image

Neumann-A:
I think this is kind of intended. Otherwise machines become completely unresponsive. 
Is this Win10?

I don't see the same problem with 24 AMD cores and Win11 building LLVM

dg0yt:
How does Windows (10 vs. 11) handle power vs. efficiency cores for these scheduling levels?

Binary-Song:
I have a win11 machine with 32 AMD cores in my workplace and it also has this issue.

Binary-Song:
> I think this is kind of intended. Otherwise machines become completely unresponsive. 

I think that is a trade-off that should be left for the users to decide. Currently we already have VCPKG_MAX_CONCURRENCY for that.

Thomas1664:
> I think this is kind of intended. Otherwise machines become completely unresponsive.

I think setting process priority to idle was just a workaround for poor memory management by Windows. Also, this was changed to idle back in 2017 so I would expect this to be fixed in Windows by now. High memory usage is just a problem if you run out of swap. My MacBook sits constantly at 90% memory usage and still runs flawlessly. 

Even if idle priority would still be necessary, I don't think locking users into slow builds is a good idea. At the bare minimum there should be an escape hatch.

Binary-Song:
@BillyONeal What do you think? If you merge this, you are the hero that saves users millions of dollars 😉!

Thomas1664:
> @BillyONeal What do you think? If you merge this, you are the hero that saves users millions of dollars 😉!

Now they'll never merge this cause it means that Microsoft would loose millions of dollars

Binary-Song:
@BillyONeal

JavierMatosD:
This looks like the right change to me, but will bring up to the team.

BillyONeal:
> I think this is kind of intended. Otherwise machines become completely unresponsive.

That is the case. We want builds to yield to normal other work to keep the system responsive, though of course we don't want to leave perfectly good cores idle.

Is the behavior you observe affected by power management settings or something similar? It's hard to evaluate the risk vs. benefit of this change without a better understanding of why it's happening.

BillyONeal:
Couple more questions:

* The CPU listed is a desktop part; is there somehow a battery connected to this system Windows is managing?
* Do you have power management settings turned on?

Could you try adding a call to [`SetProcessInformation`](https://learn.microsoft.com/windows/win32/api/processthreadsapi/nf-processthreadsapi-setprocessinformation#remarks) which explicitly marks vcpkg's process as not Eco QoS?

```
//
// HighQoS
// Turn EXECUTION_SPEED throttling off. 
// ControlMask selects the mechanism and StateMask is set to zero as mechanisms should be turned off.
//

PowerThrottling.ControlMask = PROCESS_POWER_THROTTLING_EXECUTION_SPEED;
PowerThrottling.StateMask = 0;

SetProcessInformation(GetCurrentProcess(), 
                      ProcessPowerThrottling, 
                      &PowerThrottling,
                      sizeof(PowerThrottling));
```

cenit:
https://github.com/microsoft/vcpkg/issues/39935

this was also my observation, fully related to the same problem

BillyONeal:
@cenit Would you be willing to try the `SetProcessInformation` workaround? I would try this but the problem doesn't repro for me on the 14900HX.

cenit:
> @cenit Would you be willing to try the `SetProcessInformation` workaround? I would try this but the problem doesn't repro for me on the 14900HX.

yes tomorrow I can test. Can you format your code as a patch file so that I can apply it without any possible mistake?

BillyONeal:
> > @cenit Would you be willing to try the `SetProcessInformation` workaround? I would try this but the problem doesn't repro for me on the 14900HX.
> 
> yes tomorrow I can test. Can you format your code as a patch file so that I can apply it without any possible mistake?

#1570 ?

cenit:
ok perfect, i'll let you know

cenit:
OS Build: 19045.5247
CPU: i7-13850HX power plugged
Power mode: "Better performances"

Building ffmpeg with currently released vcpkg:
building_ffmpeg_current_vcpkg

Building ffmpeg with noQoS vcpkg (#1570):
building_ffmpeg_noQoS_vcpkg

Building ffmpeg with normal-priority vcpkg (#1555)
building_ffmpeg_normalpriority_vcpkg

Building ffmpeg with currently released vcpkg under wsl2:
building_ffmpeg_wsl2_vcpkg

The two PR are improving the situation only a little, maybe nothing if we consider errors in this very non-scientific measurement.
On the other hand, WSL2 on the same laptop clearly shows much better performances in building packages and fully usage of the CPU, with the currently released executable

cenit:
in any case something changed in the windows kernel heuristic of win10 (please note all my data is related to win10, we still didn't move to 11), because i perfectly remember that the usual picture in the task manager was with only the efficiency cores taken by the compiler...

Binary-Song:
@BillyONeal  I tried. Didn't work either. Would you mind trying a demo that I wrote to see if the problem can be reproduced on your side? https://github.com/Binary-Song/vcpkg_test_project

BillyONeal:
RE: https://github.com/microsoft/vcpkg-tool/pull/1555#issuecomment-2592407251

@cenit it looks like neither change fixes it in your case, but you also aren't seeing the 'things are only running on E cores' reported by @Binary-Song .

@Binary-Song No repro for me, sorry:

all the cores being used once build starts

BillyONeal:
@cenit Stupid question: Is the vcpkg directory where builds are happening excluded from Defender in your tests?

>in any case, vcpkg on windows is clearly the slowest of the platforms

I agree that sucks but unfortunately given that this PR doesn't fix it that argues against taking it.

@Binary-Song I think adding a switch to do this might be acceptable but fixing one system and breaking responsiveness for everyone else doesn't seem like the correct tradeoff to me.

Binary-Song:
@BillyONeal Yeah you are right. I tried this on the win 11 machine today and did not repro either. This is likely to be an OS bug that got fixed later. I will do some VM experiments about this. Adding a switch might be a reasonable solution if this is indeed an OS bug.

cenit:
> @cenit Stupid question: Is the vcpkg directory where builds are happening excluded from Defender in your tests?
> 
> > in any case, vcpkg on windows is clearly the slowest of the platforms
> 
> I agree that sucks but unfortunately given that this PR doesn't fix it that argues against taking it.
> 
> @Binary-Song I think adding a switch to do this might be acceptable but fixing one system and breaking responsiveness for everyone else doesn't seem like the correct tradeoff to me.

on the laptop considered, there is CS and no way to change any settings.

Anyway, here there are the results when building llvm, and there is a difference wrt ffmpeg

Building llvm with currently released vcpkg:
building_llvm_official_vcpkg

Building llvm with noQoS vcpkg (https://github.com/microsoft/vcpkg-tool/pull/1570):
building_llvm_noQoS_vcpkg

Building ffmpeg with normal-priority vcpkg (https://github.com/microsoft/vcpkg-tool/pull/1555):
building_llvm_normalpriority_vcpkg

As you can see, only the normal-priority version really made a difference, and only in this case (and this is expected, since EcoQoS is not in the win10 api). 

Anyway, can you explain why the llvm toolchain "see" the different behaviour while the ffmpeg one no? Do we call external tools in vcpkg in different parts and so all of them should be upgraded to see the scenario upgraded also for ffmpeg?
In any case, is it a kernel bug for win10, since you all say that win11 does not reproduce these behaviours. What is the approach in this case?

BillyONeal:
Given that it's clear we are unlikely to merge this in this form, I've set the PR to 'draft' status. We would be happy to accept adding an opt-in switch to turn on such behavior, but as proposed we are worried about breaking the responsiveness goals for which we intentionally run at lower priorities given that current versions of Windows do not seem to experience an issue.

Binary-Song:
@BillyONeal How about adding an env var `VCPKG_SUBPROCESS_PRIORITY`? Allowed values: `IDLE`, `BELOW_NORMAL`, `NORMAL`, `ABOVE_NORMAL`, `HIGH`, `REALTIME`

Binary-Song:
@BillyONeal  Also, do I need to consider non-Windows?

Binary-Song:

Binary-Song:
@BillyONeal

Binary-Song:
@BillyONeal Review please

Binary-Song:
@BillyONeal May I get a reponse before I get GTA 6?

jimwang118:
@JavierMatosD Please check whether this PR can be merged.

PR #1556, opened 2024-12-28
Output log files as collapsed sections in ci

autoantwort:
Results in 
image

### DIFF
--- a/include/vcpkg/commands.build.h
+++ b/include/vcpkg/commands.build.h
@@ -95,12 +95,14 @@ namespace vcpkg
     LocalizedString to_string(const BuildResult build_result);
     LocalizedString create_user_troubleshooting_message(const InstallPlanAction& action,
                                                         const VcpkgPaths& paths,
+                                                        const std::vector<std::string>& error_logs,
                                                         const Optional<Path>& issue_body);
     inline void print_user_troubleshooting_message(const InstallPlanAction& action,
                                                    const VcpkgPaths& paths,
+                                                   const std::vector<std::string>& error_logs,
                                                    Optional<Path>&& issue_body)
     {
-        msg::println(Color::error, create_user_troubleshooting_message(action, paths, issue_body));
+        msg::println(Color::error, create_user_troubleshooting_message(action, paths, error_logs, issue_body));
     }
 
     /// <summary>
--- a/src/vcpkg/commands.build.cpp
+++ b/src/vcpkg/commands.build.cpp
@@ -207,7 +207,7 @@ namespace vcpkg
                     msg::print(Color::warning, warnings);
                 }
                 msg::println_error(create_error_message(result, spec));
-                msg::print(create_user_troubleshooting_message(*action, paths, nullopt));
+                msg::print(create_user_troubleshooting_message(*action, paths, {}, nullopt));
                 return 1;
             }
             case BuildResult::Excluded:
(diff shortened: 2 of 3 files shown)


JavierMatosD (approved):

include/vcpkg/commands.build.h
```diff
@@ -95,12 +95,14 @@ namespace vcpkg
     LocalizedString to_string(const BuildResult build_result);
     LocalizedString create_user_troubleshooting_message(const InstallPlanAction& action,
                                                         const VcpkgPaths& paths,
+                                                        const std::vector<std::string>& error_logs,
```
BillyONeal, line 98: Why string and not path?

src/vcpkg/commands.build.cpp
```diff
@@ -1690,30 +1690,110 @@ namespace vcpkg
...
                                spec_name,
                                "]+Build+error+on+",
                                triplet,
-                               "&body=Copy+issue+body+from+",
-                               Strings::percent_encode(path));
+                               "&body=",
+                               Strings::percent_encode(body));
+    }
+
+    enum class CIType
```
BillyONeal, line 1703: I think we already have something like this floating around for telemetry purposes and command line settings to control whether it happens; I believe we should reuse that if possible.

BillyONeal:
I think this implements https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2152662 "Include logs in the console output for bad CI systems"

src/vcpkg/commands.build.cpp
BillyONeal: See https://github.com/microsoft/vcpkg-tool/blob/b733d994a01e4e0392dde2d6f1adc75bb9561809/src/vcpkg/vcpkgcmdarguments.cpp#L18-L67

src/vcpkg/commands.build.cpp
```diff
@@ -1690,30 +1690,110 @@ namespace vcpkg
...
+        {
+            // https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#grouping-log-lines
+            output.append_raw("::group::").append_raw(title).append_raw('\n');
+        }
+        else if (type == CIType::GitLabCi)
+        {
+            // https://docs.gitlab.com/ee/ci/jobs/job_logs.html#custom-collapsible-sections
+            using namespace std::chrono;
+            const auto timestamp = duration_cast<seconds>(system_clock::now().time_since_epoch()).count();
+            output.append_raw(fmt::format("\\e[0Ksection_start:{}:SECTION_NAME[collapsed=true]\r\\e[0K", timestamp))
```
BillyONeal, line 1744: `SECTION_NAME` is almost certainly incorrect; the example docs in the URI replaced that with a real name. Do you have an example of a test instance that has used this?

src/vcpkg/commands.build.cpp
```diff
@@ -1690,30 +1690,110 @@ namespace vcpkg
...
+                .append_raw(title)
+                .append_raw('\n');
+        }
+        else if (type == CIType::Azure)
+        {
+            output.append_raw("##vso[task.uploadfile]").append_raw(fs.absolute(file, VCPKG_LINE_INFO)).append_raw('\n');
```
BillyONeal, line 1750: It seems unlikely to me that `absolute` is correct here; note that when we make the `VcpkgPaths` we `chdir`. Given that this is almost certainly intended to be relative to the buildtree rather than 'overall absolute' it seems like the path already needs to be absolute before we get here?

src/vcpkg/commands.build.cpp
```diff
@@ -1690,30 +1690,110 @@ namespace vcpkg
...
         {
+            auto ci = detect_ci_type();
             const auto path = issue_body.get()->generic_u8string();
-            result.append_indent().append_raw(make_gh_issue_open_url(spec_name, triplet_name, path)).append_raw('\n');
-            if (!paths.get_filesystem().find_from_PATH("gh").empty())
+            const auto body =
+                ci.map([&](auto) {
+                      return fmt::format("Copy issue body from collapsed section \"{}\" in the ci log output",
+                                         issue_body.get()->filename());
+                  }).value_or(Strings::concat("Copy issue body from ", path));
```
BillyONeal, line 1794: This text needs to be localized.

BillyONeal (changes requested):
Only true blocker is the localized text. I'm weakly against not reusing the CI detection in `VcpkgCmdArguments`. Everything else is nitpicks.

src/vcpkg/commands.build.cpp
BillyONeal: image

They look already absolute so I'll just remove this.

BillyONeal (approved):
Please confirm you are happy with the fixes I pushed in https://github.com/microsoft/vcpkg-tool/pull/1556/commits/6e7a1aa21c6e496c0d85e752ff065e7b2f2a6387 and un-mark 'draft' if you are. Thanks!

src/vcpkg/commands.build.cpp
```diff
@@ -1702,115 +1702,113 @@ namespace vcpkg
...
+            auto& fs = paths.get_filesystem();
+            const auto path = issue_body->generic_u8string();
+            const bool collapsible = is_collapsible_ci_kind(detected_ci);
             const auto body =
-                ci.map([&](auto) {
-                      return fmt::format("Copy issue body from collapsed section \"{}\" in the ci log output",
-                                         issue_body.get()->filename());
-                  }).value_or(Strings::concat("Copy issue body from ", path));
+                collapsible ? msg::format(msgCopyIssueBodyFromCollapsibleSection, msg::path = issue_body->filename())
+                            : msg::format(msgCopyIssueBodyFromFile, msg::path = path);
```
autoantwort, line 1803: I did not localize these messages since it breaks the automatic handling of issues in the `vcpkg` repo: 
https://github.com/microsoft/vcpkg/blob/96dbba1e03d1ecc7c7096e97d5a09ff3c58e21df/.github/workflows/check_issues.yml#L49
https://github.com/microsoft/vcpkg/blob/96dbba1e03d1ecc7c7096e97d5a09ff3c58e21df/.github/workflows/check_issues.yml#L80

But localizing the text maybe improves the rate of correctly submitted issues because the users understand the text and the check could probably be adapted to the new messages.

autoantwort:
Thank you for implementing the suggestions. I forgot about this PR 😅

src/vcpkg/commands.build.cpp
BillyONeal: As far as I understand it this is explaining to the user how to make such an issue, not the content of the issue itself?

autoantwort: It is the content of the issue itself

BillyONeal: Hmmmm... OK, commented and delocalized that part.

PR #1597, opened 2025-02-12
[Rework] Binary Cache: Add write-back support

fran6co:
Trying to implement https://github.com/microsoft/vcpkg-tool/pull/1406 with the new changes in the code

### DIFF
--- a/include/vcpkg/binarycaching.h
+++ b/include/vcpkg/binarycaching.h
@@ -29,6 +29,13 @@
 
 namespace vcpkg
 {
+    struct BinaryProviderContainer
+    {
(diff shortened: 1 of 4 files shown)


fran6co:
I'm not sure I like the `writeback` option and I think it makes more sense like the original PR to write-back on every write provider given that we expect the package to be there if we have to build the package, so a write source not having a package it's more about being cleared at some point and the expectation is to contain the packages

fran6co:
In any case, I think push_success should check if the package is there already to avoid uploading when it doesn't need to. I could add the check just on writeback, but is there a reason why it's set to overwrite on push_success even if the package is already there?

fran6co:
it looks like it will never be implemented at this rate

PR #1639, opened 2025-04-05
windows: Fix crash when curl is used to download assets

autoantwort:
Fixes https://github.com/microsoft/vcpkg/issues/44832


BillyONeal (approved):
It's probably worth changing create_directory and/or create_directories to just make that a no-op but this fixes the immediate customer problem. Thanks for running it down!

PR #1643, opened 2025-04-08
[WIP] Upload binaries by chunks to azure as blocklist

talregev:
Fix https://github.com/microsoft/vcpkg/issues/44060
This code is written by  @quyykk
It taken from #1043

Upload binaries by chunks.

I started an initial code, to see if I am in the correct direction.
I want to do CI driven development, and I want to upload is working on the CI. 

I will be appreciated the help to know how to do it on this CI, or vcpkg CI.

Thank you ahead for the help.

### DIFF
--- a/include/vcpkg/base/downloads.h
+++ b/include/vcpkg/base/downloads.h
@@ -104,6 +104,14 @@ namespace vcpkg
                                     StringView display_path,
                                     const Optional<std::string>& maybe_sha512);
 
+    bool put_blocklist(DiagnosticContext& context,
+                       StringView url,
+                       const SanitizedUrl& sanitized_url,
+                       View<std::string> headers,
+                       const Path& file,
+                       std::size_t file_size,
+                       std::size_t chunk_size);
+
     bool store_to_asset_cache(DiagnosticContext& context,
                               StringView raw_url,
                               const SanitizedUrl& sanitized_url,
--- a/src/vcpkg/base/downloads.cpp
+++ b/src/vcpkg/base/downloads.cpp
@@ -17,7 +17,10 @@
 
 #include <vcpkg/commands.version.h>
 
+#include <iomanip>
 #include <set>
+#include <sstream>
+#include <iostream>
 
 using namespace vcpkg;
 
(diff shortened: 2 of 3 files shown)


talregev:
https://dev.azure.com/vcpkg/public/_build/results?buildId=114459&view=logs&j=f79cfdd7-47a8-597f-8f57-dc3e21a8f2ad

talregev:
@BillyONeal I take the code from #1043 and change it that it can now able to upload very large size binary cache to vcpkg ci, by chunks. Currently I set the chunk to 500MiB, but it can very until 4000Mib.
### Reference
* https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob?tabs=microsoft-entra-id#remarks

| Service version | Maximum block size (via Put Block) | Maximum blob size (via Put Block List) | Maximum blob size via single write operation (via Put Blob) |
| --- | --- |--- | --- |
| Version 2019-12-12 and later | 4000 MiB | Approximately 190.7 TiB (4000 MiB X 50,000 blocks) | 5000 MiB |

The code is taking the file binary from the standard input as it not ideal as it point out in the original PR.
I want to add this as a experimental feature, meaning to apply it only with x flag. I didn't have a name yet, also possible to take the chunk size from the user by this flag.

I tested it on vcpkg ci, and it working:
https://dev.azure.com/vcpkg/public/_build/results?buildId=114461&view=logs&j=f79cfdd7-47a8-597f-8f57-dc3e21a8f2ad

I also write you in discord. 

Let me know what you think.

dg0yt:
I still think this needs to consider testability first. IMO for the feature to merged there must be an end-to-end test in this repo which completes in seconds and which can be run locally. And with this testability in place from the start, the development of that feature could be accelerated.

This is how it could be done:
- Allow for configuration of the chunk size. Tests could use small chunks. No need rebuild qtdeclarative.
- Allow testing with a mock upload command. No need to test with GHA or AZP.
- Collect the pieces recorded by the mock upload command and check that the combined result is the expected result. Needed, because there is no other validation of binary artifacts fetched from cache AFAICT.

talregev:
This is wip. I already testing in vcpkg ci. I found an issue on windows. not sure why, but on llvm it not uploading. I switch to different way.
I think I found a way to upload files without stdin, and without switch to libcurl, but it need to upgrade curl.
I didn't understand your way to test it. I am not sure what you mean by mock upload command, and collect the pieces recorded.
Can you elaborate on that?

dg0yt:
> I already testing in vcpkg ci.

I see that, and this was in response to these attempts. I think it will save you a lot of time by not having to test with big packages in CI.

> I am not sure what you mean by mock upload command, and collect the pieces recorded.

The mock upload command could be a simple (shell, powershell etc.) script wich is used instead of the regular curl executable. 
Maybe the mock command part of the suggestion is overkill when you establish a quick test in vcpkg-tool CI by setting a small chunk size for a small test port. 
But verifying the download is part of end-to-end testing, not overkill.

talregev:
> I see that, and this was in response to these attempts. I think it will save you a lot of time by not having to test with big packages in CI.

On small packages it already working, including windows. It upload gz-gui and gz-sim. Only on very big package as llm it fail.

talregev:
> The mock upload command could be a simple (shell, powershell etc.) script wich is used instead of the regular curl executable.
Maybe the mock command part of the suggestion is overkill when you establish a quick test in vcpkg-tool CI by setting a small chunk size for a small test port.
But verifying the download is part of end-to-end testing, not overkill.

I still don't understand how to do it.

dg0yt:
> On small packages it already working, including windows. It upload gz-gui and gz-sim. Only on very big package as llvm it fail.

Small: two chunks? Very big: more than two chunks?

talregev:
@dg0yt There is a way to send the binary data to azure as base64 string?

dg0yt:
> @dg0yt There is a way to send the binary data to azure as base64 string?

I don't think so. Nobody really wants to transfer 4 GB instead of 3 GB.

PR #1656, opened 2025-04-19
Force LC_ALL=C for test execution

dg0yt:
Fixes running `make test` in a linux system with curl localizations.


BillyONeal (approved):
Thanks!

PR #1659, opened 2025-04-20
Optimize rename_or_delete to stop retrying after EXDEV

Thomas1664:
Fixes https://github.com/microsoft/vcpkg/issues/45130

Please note that there is already code in place to handle `std::errc::cross_device_link` for binary caching:

https://github.com/microsoft/vcpkg-tool/blob/d092b51ab6eb219e9af393d29214eeb4ad9f4750/src/vcpkg/binarycaching.cpp#L232-L249

However, copy-and-delete was only tried after calling `fs.rename_or_delete()` which tries to move the file after increasing timeouts. I decided to not move the copy-and-delete behavior to `fs.rename_or_delete()` because the file is first copied to a temproary location in `binarycaching.cpp`.

Files on different filesystems can't be renamed:

Linux: https://man7.org/linux/man-pages/man2/rename.2.html

Windows seems to support moving files but not folders across devices: https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-movefile

`std::errc::cross_device_link`: https://en.cppreference.com/w/cpp/error/errno_macros

Stack Overflow: https://stackoverflow.com/questions/43206198/what-does-the-exdev-cross-device-link-not-permitted-error-mean

### DIFF
--- a/src/vcpkg/base/files.cpp
+++ b/src/vcpkg/base/files.cpp
@@ -2045,6 +2045,11 @@ namespace vcpkg
                 this->remove_all(old_path, ec);
                 return false;
             }
+            else if (ec == std::make_error_condition(std::errc::cross_device_link))
+            {
+                // If old_path and new_path are on different file systems, trying again will never help.
+                return false;
+            }
 
             std::this_thread::sleep_for(delay);
             this->rename(old_path, new_path, ec);
--- a/src/vcpkg/binarycaching.cpp
+++ b/src/vcpkg/binarycaching.cpp
@@ -241,8 +241,8 @@ namespace
                     // either we need to make a copy or the rename failed because buildtrees and the binary
                     // cache write target are on different filesystems, copy to a sibling in that directory and rename
                     // into place
-                    // First copy to temporary location to avoid race between different instances trying to upload the
-                    // same archive, e.g. if 2 machines try to upload to a shared binary cache.
+                    // First copy to temporary location to avoid race between different vcpkg instances trying to upload
+                    // the same archive, e.g. if 2 machines try to upload to a shared binary cache.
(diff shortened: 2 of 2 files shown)


Thomas1664:
Hmm...

https://github.com/microsoft/vcpkg-tool/blob/d092b51ab6eb219e9af393d29214eeb4ad9f4750/src/vcpkg/binarycaching.cpp#L224-L236

The comment suggests `can_attempt_rename` == true if the binary cache is on the same file system as buildtrees but it is never checked for this...

Also, if the file is first copied to a temporary location, the original file in buildtrees (`zip_path`) will never be deleted:

https://github.com/microsoft/vcpkg-tool/blob/d092b51ab6eb219e9af393d29214eeb4ad9f4750/src/vcpkg/binarycaching.cpp#L238-L248

src/vcpkg/binarycaching.cpp
```diff
@@ -231,15 +233,16 @@ namespace
                 std::error_code ec;
                 if (can_attempt_rename)
                 {
-                    // if we are already on the same filesystem we can just rename into place
                     m_fs.rename_or_delete(zip_path, archive_path, ec);
```
Thomas1664, line 236: Before #802, files were copied to the binary cache, meaning the original file in zip_path did still exist after this function ended. However, this file is gone with `rename_or_delete()`

src/vcpkg/base/files.cpp
```diff
@@ -2045,6 +2045,11 @@ namespace vcpkg
                 this->remove_all(old_path, ec);
                 return false;
             }
+            else if (ec == std::make_error_condition(std::errc::cross_device_link))
```
dg0yt, line 2048: Only that `ec` was already lost at this point, due to L2041.

#1669

PR #1662, opened 2025-04-25
Remove `x-gha` binary cache provider

vicroms:
Recent [changes in GitHub Actions Cache's API](https://github.com/actions/cache/discussions/1510) have rendered the `x-gha` binary caching provider useless.
 
In a chat with GitHub Actions engineers, the following points were made:
 
- Their internal APIs were not intended to be consumed the way our program does.
- Consuming their internal APIs directly is not a paved path they endorse.
- Possible breaking changes coming in the future.
- Only intended method to consume the service is through [actions/cache](https://github.com/actions/cache).
 
Given these reasons, we have decided to remove support for GitHub Actions Cache as a binary caching provider. This may change if a public API ever becomes available.

Some alternatives to migrate from `x-gha`:

1) Use [GitHub Packages with NuGet](https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-github-packages?pivots=linux-runner), this method uses vcpkg's caching mechanism and preserves caching granularity per port. This is the method recommended by the vcpkg team.
2) Use [`actions/cache`](https://github.com/actions/cache) in your workflow to cache the `installed` directory, this is not integrated with vcpkg but is the method endorsed by the GitHub Actions team. The downside of this method is that the whole installed tree gets cached as a single artifact, so single port updates invalidate all packages.


dg0yt:
> * Their internal APIs were not intended to be consumed the way our program does.
> 
> * Consuming their internal APIs directly is not a paved path they endorse.

IOW they changed their mind since https://github.com/microsoft/vcpkg-tool/pull/836#issuecomment-1426319120.

BillyONeal:
> IOW they changed their mind since [#836 (comment)](https://github.com/microsoft/vcpkg-tool/pull/836#issuecomment-1368043769) / [#836 (comment)](https://github.com/microsoft/vcpkg-tool/pull/836#issuecomment-1426319120).

It's less 'we changed our mind' and more 'the service we were told was OK got totally sunset'

src/vcpkg/binarycaching.cpp
```diff
@@ -1843,17 +1663,7 @@ namespace
...
-                {
-                    return add_error(
-                        msg::format(msgInvalidArgumentRequiresZeroOrOneArgument, msg::binary_source = "gha"),
-                        segments[2].first);
-                }
-
-                handle_readwrite(state->gha_read, state->gha_write, segments, 1);
-
-                state->binary_cache_providers.insert("gha");
+                return add_error(msg::format(msgUnknownBinaryProviderType), segments[0].first);
```
vicroms, line 1666: Add an error message that summarizes reasons for removal and alternatives.

BillyONeal:
> Is there a reason that this won't work? (See [microsoft/vcpkg#45116 (comment)](https://github.com/microsoft/vcpkg/issues/45116#issuecomment-2829014919))

The owners of that API have indicated that there is no stable API we could use for this use case and that the only "paved path" they won't regularly break is `actions/cache@v3`+

(The new API *also* would require adding a huge number of dependencies to vcpkg e.g. protocol buffers even if we wanted to attempt that)

LegalizeAdulthood:
OK, thanks for the additional info.  I hadn't realized protobuf would be required `:)`.

The nuget tutorial for package caching is lacking some information, I filed a doc bug on improving that.  Basically, the recipe given was one that would only work for linux.

BillyONeal:
> The nuget tutorial for package caching is lacking some information, I filed a doc bug on improving that. Basically, the recipe given was one that would only work for linux.

I don't see anything restricting to Linux in https://learn.microsoft.com/vcpkg/consume/binary-caching-nuget ; can you clarify?

LegalizeAdulthood:
Look here instead: https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-github-packages

I wrote up a gist that explains how I got everything working:
[Using NuGet Package Caching with VcPkg and GitHub Actions](https://gist.github.com/LegalizeAdulthood/4e839bf183164ed56ea02aa4834d77e1)

BillyONeal:
> Look here instead: https://learn.microsoft.com/vcpkg/consume/binary-caching-github-packages

@LegalizeAdulthood Did you try the Windows/Linux toggle at the top?

The Windows/Linux toggle

BillyONeal (approved):
I'm sad that this is broken with no practical replacement but it is what we've got :(

LegalizeAdulthood:
> @LegalizeAdulthood Did you try the Windows/Linux toggle at the top?

Oh, you mean the button that's way far away from the actual code such that it is off screen while I'm reading the code? `:)`

No, never even noticed it.

JavierMatosD (approved):

DockedFerret800:
> I don't know who can do this but the documentation page https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-github-actions-cache should urgently be updated to say that the method described no longer works.

https://github.com/microsoft/vcpkg-docs/pull/478

johnwason:
I maintain https://github.com/johnwason/vcpkg-action which has an option to use this feature and found the tests were failing. My old strategy runs the vcpkg cli twice. The first run to capture a list of the packages to build and uses this as a cache key. This cache key stores the local binary cache files from vcpkg. The second run can then use these files, or update a new cache if the cache key doesn't exist. It isn't a perfect solution but it does appear to still work.

LegalizeAdulthood:
> I maintain https://github.com/johnwason/vcpkg-action which has an option to use this feature and found the tests were failing. My old strategy runs the vcpkg cli twice. The first run to capture a list of the packages to build and uses this as a cache key. This cache key stores the local binary cache files from vcpkg. The second run can then use these files, or update a new cache if the cache key doesn't exist. It isn't a perfect solution but it does appear to still work.

It sounds like we could benefit from having a "JSON file API" for vcpkg, similar to what we have for CMake, whereby external tools could query vcpkg about things like list of files to be cached as sources or binaries for a port.

johnwason:
It appears you can use the actions/cache as a node API to load and store caches. See this action for an example: https://github.com/hendrikmuhs/ccache-action

LegalizeAdulthood:
> It appears you can use the actions/cache as a node API to load and store caches.

Yes, but it's up to you to identify which sets of files are to be cached together with the low-level cache action.

With vcpkg what you want is to cache all the files related to a dependency and it's associated hash (e.g. triplet plus maybe a hash of the port directory contents).  This is why I was thinking it would be beneficial to have a "JSON file API" to vcpkg.  vcpkg knows what files go with what dependency and if these file lists were available externally to a caching tool, then CI caching mechanisms could be decoupled from vcpkg entirely.

johnwason:
An option to save a text file with one vcpkg cache filename per line alongside --dry-run would probably be enough for my use case.

autoantwort:
You can use something like this: https://github.com/lukka/run-cmake/issues/152#issuecomment-2833729132

PR #1669, opened 2025-05-03
Fix cross-device rename_or_delete

dg0yt:
Fixes 10.1 seconds delay for each file write operation across devices, e.g. for `FilesWriteBinaryProvider`.
Fixes https://github.com/microsoft/vcpkg/issues/45130.

### DIFF
--- a/src/vcpkg/base/files.cpp
+++ b/src/vcpkg/base/files.cpp
@@ -2031,6 +2031,7 @@ namespace vcpkg
     {
         this->rename(old_path, new_path, ec);
         using namespace std::chrono_literals;
+        std::error_code local_ec;
         for (const auto& delay : {10ms, 100ms, 1000ms, 10000ms})
         {
             if (!ec)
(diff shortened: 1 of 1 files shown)


src/vcpkg/base/files.cpp
```diff
@@ -2031,14 +2031,15 @@ namespace vcpkg
...
+        std::error_code local_ec;
         for (const auto& delay : {10ms, 100ms, 1000ms, 10000ms})
         {
             if (!ec)
             {
                 return true;
             }
             else if (ec == std::make_error_condition(std::errc::directory_not_empty) ||
-                     ec == std::make_error_condition(std::errc::file_exists) || this->exists(new_path, ec))
+                     ec == std::make_error_condition(std::errc::file_exists) || this->exists(new_path, local_ec))
```
dg0yt, line 2042: Overwriting `ec` here broke early `return false` for `std::errc::cross_device_link`.

dg0yt: BTW the forced formatting didn't let me put the trailing `this->exists` on a new line were it would be much more obvious...

BillyONeal (approved):
Thanks!

PR #1684, opened 2025-05-13
Add platform support for Solaris and illumos

trisk:
This adds support for building and using vcpkg-tool on Solaris and illumos systems. The intent is to get the rest of vcpkg working. The majority of the work is in the `vcpkg::get_user_mac_hash` implementation using interfaces available in Solaris 8 and later. Tests pass with clang 18 on illumos/amd64.


trisk:

vicroms (approved):

PR #1738, opened 2025-07-29
Add `x-azcopy` and `x-azcopy-sas` binary caching sources

vicroms:
Adds two new binary sources that interact with Azure Storage blob containers using the `azcopy` executable.

* `x-azcopy-sas,<url>,<sas>[,<rw>]`: Intended as a drop-in replacement for `azblob`, it uses SAS tokens for authentication.
* `x-azcopy,<url>[,<rw>]`:  Intended for non-SAS authentication methods like Microsoft Entra ID; requires preloading credentials or/and the `AZCOPY_AUTO_LOGIN_TYPE`.

Internally, we use `azcopy copy --from-to BlobLocal {destination} --include-path {files}` argument to download multiple files in a single `azcopy` job, this improves performance and brings it more in line with `azblob`, that uses a single `curl` invocation to download a batch of files. Upload happens one file at a time in the background.

**Small test**

```
x-azblob:
    Restored 5 package(s) from HTTP servers in 3.8 s. Use --debug to see more details.

x-azcopy:
    Restored 5 package(s) from Azure Storage in 1.5 min. Use --debug to see more details.

x-azcopy-sas:
    Restored 5 package(s) from Azure Storage in 4.7 s. Use --debug to see more details.
```

**Large test**

```
azcopy-sas (with --include-path):
    Restored 141 package(s) from Azure Storage in 1.5 min. Use --debug to see more details.

azblob:
    Restored 141 package(s) from HTTP servers in 3 min. Use --debug to see more details.
```

Related:

- [X] End-to-end tests:
  -  [Test PR](https://github.com/microsoft/vcpkg/pull/46705) [CI results](https://dev.azure.com/vcpkg/public/_build/results?buildId=119182&view=results)
  - Large blobs: https://github.com/microsoft/vcpkg/pull/45461, results: https://dev.azure.com/vcpkg/public/_build/results?buildId=118645&view=results
- [X] Documentation (PR: https://github.com/MicrosoftDocs/vcpkg-docs/pull/511)


src/vcpkg/binarycaching.cpp
```diff
@@ -1814,6 +1904,93 @@ namespace
                 handle_readwrite(
                     state->upkg_templates_to_get, state->upkg_templates_to_put, std::move(upkg_template), segments, 4);
             }
+            else if (segments[0].second == "x-azcopy")
+            {
+                // Scheme: x-azcopy,<baseurl>,<sas>[,<readwrite>]
```
vicroms:
```suggestion
                // Scheme: x-azcopy,<baseurl>[,<readwrite>]
```

src/vcpkg/binarycaching.cpp
```diff
@@ -1814,6 +1995,16 @@ namespace
                 handle_readwrite(
                     state->upkg_templates_to_get, state->upkg_templates_to_put, std::move(upkg_template), segments, 4);
             }
+            else if (segments[0].second == "x-azcopy")
+            {
+                // Scheme: x-azcopy,<baseurl>[readwrite>]
+                handle_azcopy_segments("azcopy", false);
```
vicroms, line 2001:
```suggestion
                handle_azcopy_segments("x-azcopy", false);
```

src/vcpkg/binarycaching.cpp
```diff
@@ -1814,6 +1995,16 @@ namespace
...
+            }
+            else if (segments[0].second == "x-azcopy-sas")
+            {
+                // Scheme: x-azcopy,<baseurl>,<sas>[,<readwrite>]
+                handle_azcopy_segments("azcopy-sas", true);
```
vicroms, line 2006:
```suggestion
                handle_azcopy_segments("x-azcopy-sas", true);
```

include/vcpkg/metrics.h
```diff
@@ -17,6 +17,7 @@ namespace vcpkg
         AssetSource,
         BinaryCachingAws,
         BinaryCachingAzBlob,
+        BinaryCachingAzCopy,
```
BillyONeal, line 20: Probably want a separate one for the SAS thing?

src/vcpkg/binarycaching.cpp
```diff
@@ -925,7 +925,7 @@ namespace
         {
         }
 
-        static std::string make_object_path(const std::string& prefix, const std::string& abi)
+        virtual std::string make_object_path(const std::string& prefix, const std::string& abi) const
```
BillyONeal: Can you explain why this is being made virtual? Being static means all existing callers are expecting to call it non-virtually, and the path to the zip file, which this returns, should not change between different providers. (After all, it is a goal for the different providers to share the same zip if at all possible)

src/vcpkg/binarycaching.cpp
```diff
@@ -1176,6 +1206,72 @@ namespace
...
+            }
+
+            return msg::format_error(msgLaunchingProgramFailed, msg::tool_name = Tools::AZCOPY)
+                .append_raw(' ')
+                .append_raw(maybe_output.error().to_string());
+        }
+
+        ExpectedL<RestoreResult> download_file(StringView url, const Path& archive) const override
+        {
+            auto maybe_result = stat(url);
```
BillyONeal: We should be avoiding this extra network auth cycle and network RTT if at all possible.

src/vcpkg/binarycaching.cpp
```diff
@@ -1416,6 +1512,91 @@ namespace
         void handle_segments(std::vector<std::pair<SourceLoc, std::string>>&& segments)
         {
             Checks::check_exit(VCPKG_LINE_INFO, !segments.empty());
+
+            auto handle_azcopy_segments = [&](StringView azcopy_source, bool requires_sas) {
```
BillyONeal: That the entire first half of this function depends on the bool suggests that it should not be one function, but two. Also, it looks like the lambda could be avoided if `state` gets passed?

src/vcpkg/binarycaching.cpp
```diff
@@ -1416,6 +1512,91 @@ namespace
...
+                    if (segments.size() < min_segments)
+                    {
+                        return add_error(
+                            msg::format(msgInvalidArgumentRequiresBaseUrlAndToken, msg::binary_source = azcopy_source),
+                            segments[0].first);
+                    }
+
+                    if (segments.size() > max_segments)
+                    {
+                        return add_error(msg::format(msgInvalidArgumentRequiresTwoOrThreeArguments,
```
BillyONeal: It looks like there was an attempt to parameterize this with min_segments and max_segments but the error message shows that these are effectively disjoint

src/vcpkg/binarycaching.cpp
```diff
@@ -1416,6 +1512,91 @@ namespace
...
+
+                const size_t rw_segment_idx = max_segments - 1;
+                handle_readwrite(state->azcopy_read_templates,
+                                 state->azcopy_write_templates,
+                                 std::move(url_template),
+                                 segments,
+                                 rw_segment_idx);
+
+                // We count azcopy and azcopy-sas as the same provider
+                state->binary_cache_providers.insert("azcopy");
```
BillyONeal: It seems like a lot of this is the same as azblob and should be extracted into a function they both use.

src/vcpkg/binarycaching.cpp
BillyONeal: I submitted a PR with changes resolving all comments herein about handle_azcopy_segments: https://github.com/vicroms/vcpkg-tool/pull/1

src/vcpkg/binarycaching.cpp
BillyONeal: It seems like this is trying to cram azcopy into the generic `ObjectStorageProvider` which is common to all the aws-alikes but not necessarily here.

Could this be avoided by implementing ZipXxxProvider(s) for azcopy instead?

src/vcpkg/binarycaching.cpp
```diff
@@ -988,7 +990,7 @@ namespace
         {
         }
 
-        static std::string make_object_path(const std::string& prefix, const std::string& abi)
+        virtual std::string make_object_path(const std::string& prefix, const std::string& abi) const
```
vicroms:
```suggestion
        static std::string make_object_path(const std::string& prefix, const std::string& abi)
```

vicroms:
Back to draft while the `x64-osx` upload failures are being investigated: https://dev.azure.com/vcpkg/public/_build/results?buildId=118621&view=logs&j=7b75bd19-17d3-53d4-00fd-23f1a49a8ba4&t=c0c57043-5abe-5a13-5ef1-aa66ea261ab3&l=95

src/vcpkg/binarycaching.cpp
```diff
@@ -1020,34 +1020,243 @@ namespace
...
+        static std::vector<std::vector<std::string>> batch_azcopy_args(const std::vector<std::string>& abis,
+                                                                       const size_t fixed_len)
+        {
+            constexpr size_t ABI_ENTRY_LEN = 68; // 64 for SHA256 + 4 for ".zip"
+            const size_t available_len = Command::maximum_allowed - fixed_len;
+
+            // Not enough space for even one entry
+            if (available_len < ABI_ENTRY_LEN)
+            {
+                return {};
```
BillyONeal: Should this be `Checks::unreachable`?

src/vcpkg/binarycaching.cpp
```diff
@@ -1020,34 +1020,243 @@ namespace
...
+                }
+                current_batch.push_back(abis[i] + ".zip");
+                current_len += entry_len;
+            }
+            if (!current_batch.empty())
+            {
+                batches.push_back(std::move(current_batch));
+            }
+            return batches;
+        }
```
BillyONeal: Because this construction is not value-dependent on the ABI hash (we know each entry will be exactly ABI_ENTRY_LEN except either the first/last one) we could make this a lot simpler by just calculating that figure and dividing by it. How about something like:

```suggestion
        // Batch the azcopy arguments to fit within the maximum allowed command line length.
        static std::vector<std::vector<std::string>> batch_azcopy_args(const std::vector<std::string>& abis,
                                                                       const size_t fixed_len)
        {
            constexpr ptrdiff_t ABI_ENTRY_LEN = 68; // 64 for SHA256 + 4 for ".zip"
            const ptrdiff_t available_len = static_cast<ptrdiff_t>(Command::maximum_allowed) - fixed_len;

            // Not enough space for even one entry
            if (available_len < ABI_ENTRY_LEN)
            {
                return {};
            }

            const size_t entries_per_batch =
                1 + // the first entry with no ;
                (available_len - ABI_ENTRY_LEN) / (ABI_ENTRY_LEN + 1); // all remaining entries with a ;

            auto first = abis.begin();
            const auto last = abis.end();
            std::vector<std::vector<std::string>> batches;
            while (first != last)
            {
                auto end_of_batch = first + (std::min)(last - first, entries_per_batch);
                std::vector<std::string> current_batch;
                while (++first != end_of_batch) {
                    current_batch.emplace_back(abi + ".zip");
                }
                batches.emplace_back(std::move(current_batch));
                first = end_of_batch;
            }

            return batches;
        }
```

src/vcpkg/binarycaching.cpp
```diff
@@ -1020,34 +1020,243 @@ namespace
...
+
+        // Check for valid cache file names. The format is `<abi>.zip`. WHere `abi` is a 64-character hex string
+        static bool is_cache_file(StringView filename)
+        {
+            static constexpr size_t ABI_LENGTH = 64;
+
+            if (filename.size() != ABI_LENGTH + 4) return false;
+            for (size_t idx = 0; idx < ABI_LENGTH && ParserBase::is_hex_digit(filename[idx]); ++idx)
```
BillyONeal: That can't have been right, it just stops at the non-hexit rather than failing

```suggestion
            for (size_t idx = 0; idx < ABI_LENGTH; ++idx)
```

src/vcpkg/binarycaching.cpp
```diff
@@ -1020,34 +1020,243 @@ namespace
...
+                msg::println_warning(LocalizedString::from_raw(output->output));
+                return {};
+            }
+
+            std::vector<std::string> abis;
+            for (auto&& line : Strings::split(output->output, '\n'))
+            {
+                if (line.empty()) continue;
+                // `azcopy list` output uses format `<filename>; Content Length: <size>`, we only need the filename
+                auto parts = Strings::split(line, ';');
```
BillyONeal: Do we know that `<filename>` can't contain `;`?

Instead of using `split` for this, since we only ever look at the first block, could we use find instead?

```c++
auto first_part_end = std::find(line.begin(), line.end(); ';');
if (first_part_end != line.end()) {
    StringView abifile = Strings::trim(StringView{first_part_end, line.end()});
    if (is_cache_file(abifile))
    {
        // remove ".zip" extension
        abis.emplace_back(abifile.substr(0, abifile.size() - 4).to_string());
    }
}
```

src/vcpkg/binarycaching.cpp
```diff
@@ -1020,34 +1020,243 @@ namespace
...
+                }
+            }
+            return abis;
+        }
+
+        void acquire_zips(View<const InstallPlanAction*> actions,
+                          Span<Optional<ZipResource>> out_zip_paths) const override
+        {
+            std::vector<CacheAvailability> precheck_results{actions.size(), CacheAvailability::unknown};
+            precheck(actions, precheck_results);
```
BillyONeal: Do we need this extra network RTT or would it be better to just try and see what we get?

src/vcpkg/binarycaching.cpp
```diff
@@ -1020,34 +1020,243 @@ namespace
...
+                    cmd_execute_and_capture_output(Command{base_cmd}.string_arg(Strings::join(";", batch)));
+                // We don't return on a failure because the command may have
+                // only failed to restore some of the requested packages.
+                if (!maybe_output.has_value())
+                {
+                    msg::println_warning(maybe_output.error());
+                }
+            }
+
+            for (auto&& file : m_fs.get_files_recursive(tmp_downloads_location, VCPKG_LINE_INFO))
```
BillyONeal:
```suggestion
            for (auto&& file : m_fs.get_files_non_recursive(tmp_downloads_location, VCPKG_LINE_INFO))
```

?

src/vcpkg/binarycaching.cpp
```diff
@@ -1020,34 +1020,243 @@ namespace
...
-        std::string make_object_path(const std::string& prefix, const std::string& abi) const override
+        LocalizedString restored_message(size_t count,
+                                         std::chrono::high_resolution_clock::duration elapsed) const override
         {
-            return Strings::replace_all(prefix, "{sha}", abi);
+            return msg::format(
+                msgRestoredPackagesFromAzureStorage, msg::count = count, msg::elapsed = ElapsedTime(elapsed));
         }
+
+        Path m_buildtrees;
```
BillyONeal, line 1182: No change requested here because HttpGetBinaryProvider already does this, but it seems kinda strange to me that we use buildtrees for the temporary zip on 'get' but packages for the temporary zip on 'push'. It seems like we should use packages for both?

src/vcpkg/binarycaching.cpp
```diff
@@ -1020,34 +1020,243 @@ namespace
...
+                                  .string_arg(archive)
+                                  .string_arg(url);
+
+            return flatten(cmd_execute_and_capture_output(upload_cmd), Tools::AZCOPY);
+        }
+
+        size_t push_success(const BinaryPackageWriteInfo& request, MessageSink& msg_sink) override
+        {
+            if (!request.zip_path) return 0;
+            const auto& zip_path = *request.zip_path.get();
```
BillyONeal:
```suggestion
            const auto& zip_path = request.zip_path.value_or_exit(VCPKG_LINE_INFO);
```

This thing asks for zips. If we get here and there aren't zips something bad has happened and we need to die.

BillyONeal (changes requested):
The only comment I'm 'request changes' over is the one on line 1078 where the for loop seems clearly broken. The other things are nitpicks.

src/vcpkg/binarycaching.cpp
BillyONeal: I don't know why GitHub's diff is so confused :(

src/vcpkg/binarycaching.cpp
vicroms: I can try benchmarking this. It used to be slower when we did single package per invocation (because of deriving from `IObjectStorageProvider`).

src/vcpkg/binarycaching.cpp
vicroms: Files get downloaded to `{tmp_downloads_location}/{container name}/`. I didn't find a way to force the destination not to contain a subdirectory, so this was done to avoid parsing out the container name from the URL. I can see this becoming an issue if there are multiple `azcopy` sources given that we don't clean `tmp_downloads_location`.

src/vcpkg/binarycaching.cpp
vicroms: Ah yes, I had written this as a one-liner `for` but felt the code was easier to read without trying to be clever about it and forgot to remove the condition.

src/vcpkg/binarycaching.cpp
BillyONeal: I would be happy with just a comment explaining that that is faster because it enables a bulk test instead of per-request tests if you'd prefer that.

src/vcpkg/binarycaching.cpp
BillyONeal: Makes sense!

src/vcpkg/binarycaching.cpp
BillyONeal: > StringView abifile = Strings::trim(StringView{first_part_end, line.end()});

Megaderp I did this backwards sorry. Should be:

> StringView abifile = Strings::trim(StringView{line.begion(), first_part_end});

I think

BillyONeal (approved):

src/vcpkg/binarycaching.cpp
```diff
@@ -1020,6 +1020,234 @@ namespace
...
+            static constexpr size_t ABI_LENGTH = 64;
+
+            if (filename.size() != ABI_LENGTH + 4) return false;
+            for (size_t idx = 0; idx < ABI_LENGTH; ++idx)
+            {
+                if (!ParserBase::is_hex_digit(filename[idx]))
+                {
+                    return false;
+                }
+            }
```
Thomas1664: Can this be replaced with std::all_of?

src/vcpkg/binarycaching.cpp
```diff
@@ -1030,6 +1033,203 @@ namespace
         std::shared_ptr<const IObjectStorageTool> m_tool;
     };
 
+    struct AzCopyStorageProvider : ZipReadBinaryProvider
+    {
+        AzCopyStorageProvider(
+            ZipTool zip, const Filesystem& fs, const Path& buildtrees, AzCopyUrl&& az_url, const Path& tool)
```
BillyONeal, line 1039: Copy/move antipattern on `ZipTool` seems bad to me but the other providers are already doing it so no change requested.

src/vcpkg/binarycaching.cpp
```diff
@@ -2930,3 +3243,34 @@ FeedReference vcpkg::make_nugetref(const InstallPlanAction& action, StringView p
...
+    std::vector<std::vector<std::string>> batches;
+    while (first != last)
+    {
+        auto end_of_batch = first + std::min(static_cast<size_t>(last - first), entries_per_batch);
+        std::vector<std::string> current_batch;
+        while (first != end_of_batch)
+        {
+            current_batch.emplace_back(*(first++));
+        }
+        batches.emplace_back(std::move(current_batch));
```
BillyONeal:
```suggestion
        batches.emplace_back(first, end_of_batch);
```

All shall love the range constructor and despair

all shall love me and despair by Galadriel from Lord of the Rings

BillyONeal (approved):

src/vcpkg/binarycaching.cpp
vicroms: I think this would be greatly improved by something like @autoantwort's #911 but out of the scope of this PR.

PR #1747, opened 2025-08-03
Add support for tvOS and watchOS platforms

wkornewald:
This is taking the same approach as here: https://github.com/microsoft/vcpkg-tool/pull/1681/files


wkornewald:
@trisk @BillyONeal Do you have time for this tiny PR? Then I could prepare the next PR in vcpkg.

BillyONeal (approved):
Thanks!

ras0219-msft (approved):

PR #1752, opened 2025-08-12
percent-escape square brackets in URL

### LINKED ISSUE #1751: Issue submission URL is not clickable
On Windows 11, running vcpkg version 2025-04-01-9c604254140797833b6f76908435c9fcbf09920e in the Developer PowerShell for VS.

If there is a build error, some helpful text appears with a URL which automatically creates a new GitHub issue. Unfortunately clicking this link doesn't open the correct location. The text looks like

```
You can submit a new issue at:
  https://github.com/microsoft/vcpkg/issues/new?title=[openslide]+Build+error+on+x64-windows&body=Copy%20issue%20body%20from%20E%3A%2Fsource%2FWinUIAppDemo%2Fvcpkg_installed%2Fvcpkg%2Fissue_body.md
```

When Terminal parses the text to create a clickable link it stops at `[`, so the link leads to `https://github.com/microsoft/vcpkg/issues/new?title=` which is incomplete.

It looks like there should be a bit more percent-escaping applied, here

https://github.com/microsoft/vcpkg-tool/blob/7c1c14801f5f6865a3f1eede68b6538ff0669b7c/src/vcpkg/commands.build.cpp#L2015-L2021

and

https://github.com/microsoft/vcpkg-tool/blob/7c1c14801f5f6865a3f1eede68b6538ff0669b7c/src/vcpkg/commands.build.cpp#L1878-L1885

By replacing '[' and ']' with %5B and %5D respectively, or by putting the title text through `Strings::percent_encode` as well as the body.

pedro-w:
There is a bit of repetition in the code but I opted to keep changes to a minimum for fixing this very minor bug.

Fixes #1751

### DIFF
--- a/src/vcpkg/commands.build.cpp
+++ b/src/vcpkg/commands.build.cpp
@@ -1877,10 +1877,9 @@ namespace vcpkg
 
     static std::string make_gh_issue_open_url(StringView spec_name, StringView triplet, StringView body)
     {
-        return Strings::concat("https://github.com/microsoft/vcpkg/issues/new?title=[",
-                               spec_name,
-                               "]+Build+error+on+",
-                               triplet,
+        auto title = fmt::format("[{}] build error on {}", spec_name, triplet);
+        return Strings::concat("https://github.com/microsoft/vcpkg/issues/new?title=",
+                               Strings::percent_encode(title),
                                "&body=",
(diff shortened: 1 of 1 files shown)


BillyONeal (approved):
Thanks!

PR #1787, opened 2025-09-14
Support toolset v145 (Visual Studio 2026)

huangqinjin:


BillyONeal (approved):
Please note that this approval doesn't mean 'official support' for that version yet, if that makes sense. Thanks for the help!

PR #1791, opened 2025-09-22
Exit when json::parse_string failed, avoid double valid string check

miyanyan:
There are two points to call `json::parse_string`
https://github.com/microsoft/vcpkg-tool/blob/46b8cce32be200765c7137203a6f9bcd81a80406/src/vcpkg/base/json.cpp#L910-L928
https://github.com/microsoft/vcpkg-tool/blob/46b8cce32be200765c7137203a6f9bcd81a80406/src/vcpkg/base/json.cpp#L1025-L1039

The current implementation of `json::parse_string` doesn't exit on invalid strings. So when the string is invalid: 
* `json::parse_value` exits but performs redundant string validation
* `json::parse_kv_pair` continues processing even with an invalid string, which is unnecessary.

### DIFF
--- a/src/vcpkg/base/json.cpp
+++ b/src/vcpkg/base/json.cpp
@@ -336,6 +336,12 @@ namespace vcpkg::Json
         val.underlying_ = std::make_unique<ValueImpl>(ValueKindConstant<VK::String>(), std::move(s));
         return val;
     }
+    Value Value::string_valid(std::string&& s) noexcept
+    {
+        Value val;
+        val.underlying_ = std::make_unique<ValueImpl>(ValueKindConstant<VK::String>(), std::move(s));
+        return val;
+    }
     Value Value::array(Array&& arr) noexcept
     {
         Value val;
@@ -699,6 +705,7 @@ namespace vcpkg::Json
                 }
 
                 add_error(msg::format(msgUnexpectedEOFMidString));
+                vcpkg::Checks::msg_exit_with_message(VCPKG_LINE_INFO, msgInvalidString);
                 return res;
             }
 
@@ -1036,7 +1043,7 @@ namespace vcpkg::Json
                 {
                     case '{': return parse_object();
                     case '[': return parse_array();
-                    case '"': return Value::string(parse_string());
+                    case '"': return Value::string_valid(parse_string());
                     case 'n':
                     case 't':
                     case 'f': return parse_keyword();
--- a/include/vcpkg/base/json.h
+++ b/include/vcpkg/base/json.h
@@ -130,6 +130,7 @@ namespace vcpkg::Json
         static Value integer(int64_t i) noexcept;
         static Value number(double d) noexcept;
         static Value string(std::string&& s) noexcept;
+        static Value string_valid(std::string&& s) noexcept;
         template<class StringLike, std::enable_if_t<std::is_constructible_v<StringView, const StringLike&>, int> = 0>
         static Value string(const StringLike& s) noexcept
         {


src/vcpkg/base/json.cpp
```diff
@@ -699,6 +705,7 @@ namespace vcpkg::Json
                 }
 
                 add_error(msg::format(msgUnexpectedEOFMidString));
+                vcpkg::Checks::msg_exit_with_message(VCPKG_LINE_INFO, msgInvalidString);
```
BillyONeal, line 708: This code is quite clearly trying to *return* the error message rather than terminate, I don't think elevating an error into termination is appropriate, particularly when this can happen from whatever user supplied input rather than a bug in the tool.

src/vcpkg/base/json.cpp
```diff
@@ -336,6 +336,12 @@ namespace vcpkg::Json
         val.underlying_ = std::make_unique<ValueImpl>(ValueKindConstant<VK::String>(), std::move(s));
         return val;
     }
+    Value Value::string_valid(std::string&& s) noexcept
```
BillyONeal, line 339: Unless there is a fairly extreme perf cost from the check on 330 I'm not really seeing a good reason to add this API. If there is an extreme perf reason, perhaps we should consider altering how that assertion works rather than adding a new API footgun like this.

src/vcpkg/base/json.cpp
miyanyan: Currently there are only two functions call `json::parse_string`, when the string is invalid:
* `json::parse_value` -> `Value::string` -> `json::parse_string`, it will also terminate in `Value::string`
* `json::parse_object` -> `json::parse_kv_pair` -> `json::parse_string`, it will not terminate, but continue to create a invalid kv_pair in  `json::parse_kv_pair`, and insert it in `json::parse_object`

> I don't think elevating an error into termination is appropriate

I agree with this, but seems `json::parse_value` and `json::parse_object` don't behave the same

src/vcpkg/base/json.cpp
miyanyan: Yeah, it's not much necessary. I am investigating why sometimes `vcpkg search` is slow and then find this `json::parse_value` function

When I pref  `vcpkg search a`, the `json::parse_value` function only costs about 5%-10%

image

BillyONeal: I'm guessing you're using manifest mode? In that case the reason search is awfully slow is that we end up launching a git process per port to extract the tree sha. I had some work in progress to use `git mktree` to replace N git invocations with 2 git invocations but I haven't had time to get back to that.

Also unfortunately child process launch time etc. won't be detected by a sampling based profiler like this since it only cares about CPU time in the process in question.

BillyONeal:
Closed and reopened to force re-merge to pick up the macOS build fix

src/vcpkg/base/json.cpp
miyanyan: I'm using the classic mode, and most of the time `vcpkg search` takes only < 500ms, but occasionally it's slow, up to 5 seconds I guess

BillyONeal: Did you take your profile during one of the times where it was slow?

miyanyan: No, currently I can't reproduce when build vcpkg from source

"slow" only occured when using offical vcpkg in classic mode, and not occured on all of my PCs

miyanyan: image

I can reproduce when **the first time** I compile after I reboot, the search method costs about 5s

BillyONeal: Given that the vast majority of the time in this chart is reading the files from disk, and you can only reproduce things being slow the first time when things aren't going to be cached in system memory, I think the real problem you are experiencing is disk I/O related which a sampling profiler like this can't help you find.

miyanyan: Yeah, I found that the slow one is HDD, and vcpkg reads 5783 files then reach the HDD limit

PR #1837, opened 2025-11-06
Fix ci test changing VCPKG_ROOT/installed

dg0yt:
Drive-by fix from work with the `ci` test: Pass the test-time install root via `@commonArgs` (from `end-to-end-tests-prelude.ps1`).


dg0yt:
I see now that the key change is also included in https://github.com/microsoft/vcpkg-tool/pull/1822.

PR #1864, opened 2025-11-21
binarycaching: implement x-aws-config,trace-ops

ghost:
# Description

## The issue

When a developer perform CMake configuration on a complex project using AWS S3 as a cache backend, `vcpkg` is apparently stuck for many minutes.

In reality it is just doing I/O with AWS S3 to download one by one pre-built packages.

## The solution

This contribution add option `x-aws-config,trace-ops` to the `x-aws` caching backend to obtain a feedback when the tool is downloading caches from an S3 bucket.

The solution is also useful for debugging purposes and it is 100% optional.

### DIFF
--- a/include/vcpkg/binarycaching.h
+++ b/include/vcpkg/binarycaching.h
@@ -168,6 +168,7 @@ namespace vcpkg
         std::vector<std::string> aws_read_prefixes;
         std::vector<std::string> aws_write_prefixes;
         bool aws_no_sign_request = false;
+        bool aws_trace_ops = false;
 
         std::vector<std::string> cos_read_prefixes;
         std::vector<std::string> cos_write_prefixes;
--- a/src/vcpkg/binarycaching.cpp
+++ b/src/vcpkg/binarycaching.cpp
@@ -1329,8 +1329,11 @@ namespace
 
     struct AwsStorageTool : IObjectStorageTool
     {
-        AwsStorageTool(const ToolCache& cache, MessageSink& sink, bool no_sign_request)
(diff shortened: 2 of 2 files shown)


BillyONeal:
I don't really like this mechanism; that the NuGet backend has that funny 'interactive' setting that globally applies across the entire providers expression is a mistake we should not repeat IMO.

However, I do think our console I/O for interacting with binary caches sucks and we should be telling more than we do. When I overhauled how download console I/O works one of the main priorities is "every time before we touch the network we need to tell the user we are doing that so if we get stuck they know why" but binary caching didn't get the same treatment yet.

Related:

* https://github.com/microsoft/vcpkg-tool/pull/1565
* https://github.com/microsoft/vcpkg-tool/pull/1580
* https://github.com/microsoft/vcpkg-tool/pull/1618

ghost:
I get your point and I agree with your view. Any access to the network should be traced somehow.
- I will keep this PR open for people requiring a temporary solution (the same as us) while waiting you for a more elegant implementation of this entire module.

BillyONeal:
> * I will keep this PR open for people requiring a temporary solution (the same as us) while waiting you for a more elegant implementation of this entire module.

Just keep your additional output and remove the option to control it?

ghost:
- This is obsolete now.

BillyONeal:
To be clear, just adding the output rather than trying to have a configure setting for it should have worked?

nickdademo:
> * This is obsolete now.

Which PR made this one obsolete?

PR #1889, opened 2025-12-19
Manually revert 87adc3f "Fix default features control by top level manifest (#1331)"

Thomas1664:
Fixes https://github.com/microsoft/vcpkg/issues/48892

I think the root cause of the issue was that the default features of the top level manifest were initially unknown and the graph was then stuck resolving the circular dependency between the individual features. This issue is now avoided because all features of the top level manifest are now already known beforehand. 

I also added an e2e test based on the manifest from the original bug report. Note that I removed the boost-test dependency. I verified that the original bug could still be reproduced.

### DIFF
--- a/azure-pipelines/end-to-end-tests-dir/manifests.ps1
+++ b/azure-pipelines/end-to-end-tests-dir/manifests.ps1
@@ -156,7 +156,6 @@ $vcpkgJson = @{
   'features' = @{
     'core-lib' = @{
       'description' = 'Dependencies for core component';
-      'dependencies' = @( 'zlib' );
     };
     'math' = @{
       'description' = 'Dependencies for math component';


Thomas1664:
Please note that this PR would again cause the bug that #1331 was trying to fix: https://github.com/microsoft/vcpkg/issues/35694#issuecomment-1912511325

The failure mode there is that if you install a feature A of the top-level manifest that depends on another feature B of the top level manifest and feature B depends on a port with default features, vcpkg would ignore that you disabled default features when referencing that dependency in feature B. To get the intended behavior, you would need to also disable default features in feature A when referencing feature B - even if you don't have any default features in the top level manifest.

IMO, the original PR did not fix this issue. It just wasn't visible because the top level manifest wouldn't (fully) participate in the dependeny graph.

Thomas1664:
@BillyONeal I wouldn't recommend merging this as is because this would pull in more additional dependencies for manifest users which might be unexpected and cause problems.

Instead, the underlying issue with  overwriting default features false  in nested top level manifest features should be fixed. Unfortunately, I don't have time to address this. Feel free to push to this PR to make these changes.

dg0yt:
A complete revert misses the point, in particular given the tests that came with #1331. The only behavioral change is much smaller, https://github.com/microsoft/vcpkg-tool/pull/1331/commits/4285366943ef74a655698f690877b2d94f7c59c1.

Thomas1664:
Closing in favor of #1890

PR #1890, opened 2025-12-20
Test/fix internal dependencies in top level manifest.

dg0yt:
Fixes https://github.com/microsoft/vcpkg/issues/48892.
Implements https://github.com/microsoft/vcpkg/issues/48892#issuecomment-3676612425

This seems to be the first test that actually covers details of handling the top-level manifest, i.e. the request is installing the given top level manifest, not unrelated ports from the registry.  
Edit: The difference is how the "source control file" is passed to `create_versioned_install_plan`.

The PR now includes the e2e test from #1889 because it helps to understand the construction of the test in `dependencies.cpp`.

### DIFF
--- a/src/vcpkg/dependencies.cpp
+++ b/src/vcpkg/dependencies.cpp
@@ -1642,14 +1642,10 @@ namespace vcpkg
         Optional<VersionedPackageGraph::PackageNode&> VersionedPackageGraph::require_package(const PackageSpec& spec,
                                                                                              const std::string& origin)
         {
-            // Implicit defaults are disabled if spec is requested from top-level spec.
-            const bool default_features_mask = origin != m_toplevel.name();
-
             auto it = m_graph.find(spec);
             if (it != m_graph.end())
             {
                 it->second.origins.insert(origin);
-                it->second.default_features &= default_features_mask;
                 return *it;
             }
 
@@ -1704,8 +1700,9 @@ namespace vcpkg
                 }
             }
 
-            it->second.default_features = default_features_mask;
+            // Implicit defaults are disabled if spec is requested from top-level spec.
             // Note that if top-level doesn't also mark that reference as `[core]`, defaults will be re-engaged.
+            it->second.default_features = origin != m_toplevel.name();
             it->second.requested_features.insert(FeatureNameCore.to_string());
 
             require_scfl(*it, it->second.scfl, origin);


src/vcpkg/dependencies.cpp
```diff
@@ -1574,6 +1574,8 @@ namespace vcpkg
                         {
                             auto deps = scfl->source_control_file->find_dependencies_for_feature(f.name);
                             if (!deps) continue;
+                            auto inserted = ref.second.requested_features.emplace(f.name).second;
+                            if (!inserted) continue;
```
dg0yt: This is adapted from the following function, `require_port_feature`.

Thomas1664: It might make more sense to do this before we go find the dependencies of the feature because there's no point in finding them if we don't do anything with them.

src/vcpkg/dependencies.cpp
```diff
@@ -1642,14 +1642,10 @@ namespace vcpkg
...
                                                                                              const std::string& origin)
         {
-            // Implicit defaults are disabled if spec is requested from top-level spec.
-            const bool default_features_mask = origin != m_toplevel.name();
-
             auto it = m_graph.find(spec);
             if (it != m_graph.end())
             {
                 it->second.origins.insert(origin);
-                it->second.default_features &= default_features_mask;
```
dg0yt, line 1652: This reverts a change from #1331 and is the fix for the livelock: Do not reset `default_features` to false.

 - The toplevel request (`ConstraintFrame`) could include multiple dependencies pointing to the toplevel manifest's default features.  
 - `resolve_stack()` calls `require_port_defaults()` for each of these dependencies.  
 - The `default_features` flag of the solution data node (`it->second`) is what prevents the creation of multiple new `resolve_stack` items (`ConstraintFrame`).

src/vcpkg/dependencies.cpp
```diff
@@ -1704,8 +1700,9 @@ namespace vcpkg
                 }
             }
 
-            it->second.default_features = default_features_mask;
+            // Implicit defaults are disabled if spec is requested from top-level spec.
             // Note that if top-level doesn't also mark that reference as `[core]`, defaults will be re-engaged.
+            it->second.default_features = origin != m_toplevel.name();
```
dg0yt, line 1705: This is not a revert but retains the actual fix from #1331: Fixing the initialization of the flag.

There are three places which engage the actual default features:
 - For the input to the `vcpkg install` command, default features are initialy unrolled in `commands.install.cpp`, subject to command line switch.
 - For dependencies requested from the toplevel manifest (name), the value is `false`. Actual default features can be engaged by `require_port_defaults()`, subject to explicit dependencies in the `resolve_stack` items (`ConstraintFrame`).
 - For dependencies requested from other manifests, , the value is `true`. The actual default features are engaged by `require_scfl()` called three lines later.

dg0yt:
Now it should be the right fix, rectifying changes from #1331.

I removed cleanup and optimization from this PR. I would submit these usefull improvements after this PR. In addition, I think that `PackageNode` must become a struct with properly named members instead of a `std::pair` with non-tell `first` and `second` members.

azure-pipelines/end-to-end-tests-dir/manifests.ps1
```diff
@@ -235,3 +235,42 @@ The following packages will be built and installed:
     vcpkg-empty-port:
 "@
 Throw-IfNonContains -Expected $expected -Actual $output
+
+# This e2e tests is mirred in the dependencies.cpp test.
```
BillyONeal, line 239: FWIW, if we can test it with a unit test I don't think an e2e test is strictly needed but I'm not going to say no to more tests :)

src/vcpkg/dependencies.cpp
BillyONeal: Q: Why isn't this `|=`?
A: Because we know it was already false/unset because if we had not inserted a new node, the `if (it != m_graph.end())` branch would have been entered.

No change requested.

BillyONeal (approved):
Thanks!

azure-pipelines/end-to-end-tests-dir/manifests.ps1
dg0yt: Without the e2e test, you miss one the "three places which engage the actual default features", command.install.cpp.

PR #1891, opened 2025-12-22
dependencies.cpp optimizations

dg0yt:
Collected while working on #1890.
Removes unused code.
Makes `batch_load_vars()` handle all dependencies from the current `ConstraintFrame` stack item. In particular, extends the initial batch when the stack is empty and the current frame already constains multiple dependencies from the top level manifest. OTOH it may have no effect at all because unbatched processing already collected the vars in `commands.install.cpp`.

### DIFF
--- a/src/vcpkg/dependencies.cpp
+++ b/src/vcpkg/dependencies.cpp
@@ -1725,8 +1725,7 @@ namespace vcpkg
         {
             for (auto&& dep : deps)
             {
-                if (!dep.platform.is_empty() &&
-                    !dep.platform.evaluate(m_var_provider.get_or_load_dep_info_vars(m_toplevel, m_host_triplet)))
+                if (!dep.platform.is_empty() && !evaluate(m_toplevel, dep.platform))
                 {
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


src/vcpkg/dependencies.cpp
```diff
@@ -1485,13 +1485,15 @@ namespace vcpkg
...
         {
-            auto vars = m_var_provider.get_dep_info_vars(spec);
+            auto vars = m_var_provider.get_dep_info_vars(frame.spec);
             if (!vars)
             {
-                // We want to batch as many dep_infos as possible, so look ahead in the stack
-                std::unordered_set<PackageSpec> spec_set = {spec};
+                // We want to batch as many dep_infos as possible, so look ahead in the frame and stack
+                std::unordered_set<PackageSpec> spec_set = {frame.spec};
+                for (auto&& d : frame.deps)
```
BillyONeal, line 1495: I hate the no `{}`s but that's what the other code in this function is doing so no change requested.

src/vcpkg/dependencies.cpp
```diff
@@ -1665,42 +1667,38 @@ namespace vcpkg
                 it->second.overlay_or_override = true;
                 it->second.scfl = p_overlay;
             }
+            else if (const auto over_it = m_overrides.find(spec.name()); over_it != m_overrides.end())
```
BillyONeal, line 1670: GitHub's diff is freaking awful as always. This is just deleting the unused `Version ver` and changing `else { if ` into `else if`

No change requested.

PR #1896, opened 2026-01-09
Install files in parallel, redux

BillyONeal:
Alternative to: https://github.com/microsoft/vcpkg-tool/pull/1256

I created a test port that just writes 10'000 files:

```console
set(INCLUDE_DIR "${CURRENT_PACKAGES_DIR}/include/${PORT}")
file(MAKE_DIRECTORY "${INCLUDE_DIR}")

foreach(i RANGE 1 10000)
    set(header_name "header${i}")
    file(WRITE "${INCLUDE_DIR}/${header_name}.h"
"// Generated header ${header_name}
#pragma once
constexpr int ${header_name}_value = ${i};
")
endforeach()

file(MAKE_DIRECTORY "${CURRENT_PACKAGES_DIR}/share/${PORT}")
vcpkg_install_copyright(FILE_LIST "${CMAKE_CURRENT_LIST_DIR}/copyright")
```

Testing before/after this change for `vcpkg install testing-headers` shows a meaningful performance improvement; I'm guessing due to antivirus involvement or similar

parallelism results

This is on AMD RYZEN AI MAX+ PRO 395 w/ Radeon 8060S

### DIFF
--- a/src/vcpkg/commands.install.cpp
+++ b/src/vcpkg/commands.install.cpp
@@ -204,7 +204,7 @@ namespace vcpkg
                     listfile_lines.push_back(list_listfile_line);
                     break;
                 }
-                case FileType::none: break; // .DS_Store or error case
+                case FileType::none: break; // skip or error case
                 default: Checks::unreachable(VCPKG_LINE_INFO); break;
             }
         }
@@ -217,10 +217,13 @@ namespace vcpkg
                     const auto& target = target_regular_files[idx];
                     if (fs.exists(target, IgnoreErrors{}))
                     {
-                        std::lock_guard<std::mutex> lock(console_mutex);
-                        msg::println_warning(msgOverwritingFile, msg::path = target);
+                        {
+                            std::lock_guard<std::mutex> lock(console_mutex);
+                            msg::println_warning(msgOverwritingFile, msg::path = target);
+                        } // unlock
+
                         fs.remove_all(target, IgnoreErrors{});
-                    } // unlock
+                    }
 
                     std::error_code ec;
                     fs.create_hard_link(source_files[idx], target, ec);


src/vcpkg/commands.install.cpp
```diff
@@ -147,61 +183,66 @@ namespace vcpkg
...
-                                           ec.message());
-                            use_hard_link = false;
-                        }
-                    }
-                    if (!use_hard_link)
+                    target_regular_files[idx] = std::move(target_file);
+                    listfile_lines.push_back(list_listfile_line);
+                    break;
+                }
+                case FileType::symlink:
```
BillyONeal, line 190: GitHub's diff is awful. This line should go with 185 on the 'red' side.

BillyONeal: Obligatory cult of Beyond Compare recruitment flyer:

a better diff of this function

src/vcpkg/commands.install.cpp
```diff
@@ -182,26 +204,44 @@ namespace vcpkg
                     listfile_lines.push_back(list_listfile_line);
                     break;
                 }
-                case FileType::symlink:
-                case FileType::junction:
+                case FileType::none: break; // .DS_Store or error case
```
BillyONeal:
```suggestion
                case FileType::none: break; // skip or error case
```

src/vcpkg/commands.install.cpp
```diff
@@ -182,26 +204,44 @@ namespace vcpkg
...
+            proximate_files.size(), [&statuses, &target_regular_files, &source_files, &fs, &console_mutex](size_t idx) {
+                if (statuses[idx] == FileType::regular)
                 {
+                    const auto& target = target_regular_files[idx];
                     if (fs.exists(target, IgnoreErrors{}))
                     {
+                        std::lock_guard<std::mutex> lock(console_mutex);
                         msg::println_warning(msgOverwritingFile, msg::path = target);
-                    }
+                        fs.remove_all(target, IgnoreErrors{});
```
BillyONeal, line 225: This remove_all can be moved out of the critical region

PR #1897, opened 2026-01-12
Switch default Ninja minimal version to a reasonable value

cqundefine:
The previous value (3.5.1) doesn't make much sense and it being that broke systems like FreeBSD which didn't have a tool download link.

This also comes with a question, while this solves the problem for now, what would be a more reasonable solution to keep the minimum versions in sync for all those platforms (like RISC-V Linux, *BSD, Solaris and more) that will most probably never have all download links added, should there a be separate JSON file listing minimal versions, should all those platforms get spoofed to x64 Linux for version checks or maybe something entirely else.

### DIFF
--- a/src/vcpkg/tools.cpp
+++ b/src/vcpkg/tools.cpp
@@ -488,7 +488,7 @@ namespace vcpkg
         virtual bool is_abi_sensitive() const override { return false; }
         virtual StringView tool_data_name() const override { return Tools::NINJA; }
         virtual std::vector<StringView> system_exe_stems() const override { return {Tools::NINJA}; }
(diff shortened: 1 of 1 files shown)


Osyotr:
> Wow, 8 year old copy pasto (3.5.1 is the minimal git version) from [e6442b0](https://github.com/microsoft/vcpkg-tool/commit/e6442b0716ea56da896549a09856a331a0ac1a53)

Which means that this code is probably unused and should be removed.

BillyONeal:
> Which means that this code is probably unused and should be removed.

It was unused because most platforms have a different value filled in from `vcpkg-tools.json`. But there was no ninja entry for FreeBSD there.

Issue #1900, opened 2026-01-19
vcpkg-tool inconsistent behaviour when using response file


TobiasFunk:
Hey i recognized that since i introduced VCPKG_HASH_ADDITIONAL_FILES in my vcpkg environment that my cache does not work any more reliably

i have a vcpkg response file containing several entries. i build it in x64 and x86, but as we use innosetup we also need to install some dependencies in x64 also in x86

so my response file looks like this (simplified):
[vcpkg.txt](https://github.com/user-attachments/files/24724135/vcpkg.txt)
```
gtest
gtest:our-x86-windows-static
```

this should ensure that we install in x86 gtest:x86 and in x64 gtest:x64 and gtest:x86
than i have these triplet files:
our-x86-windows-static.cmake:
```
include(${CMAKE_CURRENT_LIST_DIR}/shared.cmake)

set(VCPKG_TARGET_ARCHITECTURE x86)
set(VCPKG_CRT_LINKAGE static)
set(VCPKG_LIBRARY_LINKAGE static)

set(VCPKG_HASH_ADDITIONAL_FILES
    ${VCPKG_HASH_ADDITIONAL_FILES}
    "${CMAKE_CURRENT_LIST_DIR}/shared.cmake")
```

our-x64-windows-static.cmake:
```
include(${CMAKE_CURRENT_LIST_DIR}/shared.cmake)

set(VCPKG_TARGET_ARCHITECTURE x64)
set(VCPKG_CRT_LINKAGE static)
set(VCPKG_LIBRARY_LINKAGE static)

set(VCPKG_HASH_ADDITIONAL_FILES
    ${VCPKG_HASH_ADDITIONAL_FILES}
    "${CMAKE_CURRENT_LIST_DIR}/shared.cmake")
```

shared.cmake:
```
set(VCPKG_DISABLE_COMPILER_TRACKING TRUE)
set(VCPKG_ENV_PASSTHROUGH "_SupportedMinimalOs")

set(VCPKG_ENV_PASSTHROUGH_UNTRACKED
    "VCPKG_HASH_ADDITIONAL_FILES;BUILD_NUMBER;COMPUTERNAME;NAME;BUILD_URL;"
)
```

when i execute now:
cd d:\src\test
git clone https://github.com/microsoft/vcpkg.git
vcpkg\bootstrap-vcpkg.bat
[...]

PR #1901, opened 2026-01-20
fix double adding of additional_files_0...

TobiasFunk:
Should fix https://github.com/microsoft/vcpkg-tool/issues/1900 but i'm not sure if i would break an additional feature (this is why i first draft the PR)

As already commented in the ticket:

There are two places where hash_additional_files are added to abi_tag_entries:

Lines 1425-1435 (INCORRECT): Inside the port_dir_cache.get_lazy() lambda, the code adds entries directly to abi_tag_entries:

```
for (size_t i = 0; i < abi_info.pre_build_info->hash_additional_files.size(); ++i)
{
    // ...
    abi_tag_entries.emplace_back(
        fmt::format("additional_file_{}", i),
        Hash::get_file_hash(fs, file, Hash::Algorithm::Sha256).value_or_exit(VCPKG_LINE_INFO));
}
```

Lines 1474-1488 (CORRECT): After retrieving the cache entry, the code adds the same entries again:

```
for (auto& filestr : abi_info.pre_build_info->hash_additional_files)
{
    // ...
    abi_tag_entries.emplace_back(fmt::format("additional_file_{}", i++), hash);
}
```

### Why This Causes Inconsistent Behavior

First package built (cache miss): The lambda at lines 1409-1468 executes, and line 1432 incorrectly adds additional_file_* to abi_tag_entries. Then line 1486 adds them again → double entries.

Second package with same port_dir (cache hit): The lambda doesn't execute (cache hit), so only line 1486 runs → single entry.

### DIFF
--- a/src/vcpkg/commands.build.cpp
+++ b/src/vcpkg/commands.build.cpp
@@ -1419,20 +1419,8 @@ namespace vcpkg
                 port_dir_cache_entry.files = std::move(rel_port_files);
             }
             const auto& rel_port_files = port_dir_cache_entry.files;
-            // Technically the pre_build_info is not part of the port_dir cache key, but a given port_dir is only going
-            // to be associated with 1 port
-            for (size_t i = 0; i < abi_info.pre_build_info->hash_additional_files.size(); ++i)
-            {
-                const auto& file = abi_info.pre_build_info->hash_additional_files[i];
-                if (file.is_relative() || !fs.is_regular_file(file))
-                {
-                    Checks::msg_exit_with_message(
-                        VCPKG_LINE_INFO, msgInvalidValueHashAdditionalFiles, msg::path = file);
-                }
-                abi_tag_entries.emplace_back(
-                    fmt::format("additional_file_{}", i),
-                    Hash::get_file_hash(fs, file, Hash::Algorithm::Sha256).value_or_exit(VCPKG_LINE_INFO));
-            }
+            // Note: hash_additional_files are triplet-specific, not port-specific,
+            // so they are processed outside this cache lambda (see lines 1474-1488)
 
             for (const Path& rel_port_file : rel_port_files)
             {


src/vcpkg/commands.build.cpp
```diff
@@ -1419,20 +1419,8 @@ namespace vcpkg
...
-                {
-                    Checks::msg_exit_with_message(
-                        VCPKG_LINE_INFO, msgInvalidValueHashAdditionalFiles, msg::path = file);
-                }
-                abi_tag_entries.emplace_back(
-                    fmt::format("additional_file_{}", i),
-                    Hash::get_file_hash(fs, file, Hash::Algorithm::Sha256).value_or_exit(VCPKG_LINE_INFO));
-            }
+            // Note: hash_additional_files are triplet-specific, not port-specific,
+            // so they are processed outside this cache lambda (see lines 1474-1488)
```
Neumann-A, line 1423: should still run through per port customization though. Maybe add a testcase which shows the issue.

TobiasFunk: i think this comment is a bit misleading, i will remove it

this is my first change in this project, and i feel a bit uncomfortable to do the change by myself as i have not a complete picture in the project, for me the PR was more about to show the issue, but this was going wrong in this case :D

So i understand your comment. Per port customizations should be untouched by the change. Adding a test is hard as nearly no function is accessible for the test..

The problem is actual really simple. 
When using a response file you can install the same port for different triplets

gtest
gtest:our-x86-windows-static

the first entry use the triplet from vcpkg command, the second the port mentioned in the response file. If the triplet has now a VCPKG_HASH_ADDITIONAL_FILES variable, the first entry was adding the VCPKG_HASH_ADDITIONAL_FILES to the 
port_dir_cache_entry & to the abi_tag_entries +1422 to +1423
-> as i'm not 100% sure perhaps it is better to actual add it to port_dir_cache_entry.abi_entries 

[between +1472 to ](https://github.com/microsoft/vcpkg-tool/pull/1901/files/246de0c45c48cba1c634654ed15b9a860375807f#diff-a93efc9f4d36565c420ffe731eb4aafb23b3f5f300db52302f898c4a64884a94L1486)

this code is doing this as well, so actual the action for the first port is doubled (can be easily retested with some port have VCPKG_HASH_ADDITIONAL_FILES)

you will get additional_file_0 xxx 2 times, always

the second port now runs the code again in same vcpkg.exe instance, as the port_dir_cache_entry is already filled, the removed code was skipped and was resulting in additional_file_0 xxx once

This is a problem with my cache
as now 

vcpkg install gtest:our-x64-windows-static --overlay-triplet=xxx 
vcpkg install gtest:our-x86-windows-static --overlay-triplet=xxx 

and
vcpkg install --triplet=our-x64-windows-static --overlay-triplet=xxx @responefile 

results in different abi for gtest:our-x86-windows-static

BillyONeal:
It looks like the double adding was added by accident as part of https://github.com/microsoft/vcpkg-tool/pull/802 ; probably from merge conflict hell since that PR went on for so long.

BillyONeal:
Closing due to being replaced by #1914 . Thanks for your contribution!

PR #1909, opened 2026-02-05
Add manifest support to depend-info command

dg0yt:
Resolves https://github.com/microsoft/vcpkg/issues/36520.

Adapting code from the install command.

Not including the top-level manifest in the graphs. This omits useful information, but could be added later.

### DIFF
--- a/src/vcpkg/commands.depend-info.cpp
+++ b/src/vcpkg/commands.depend-info.cpp
@@ -476,6 +476,8 @@ namespace vcpkg
         const auto unsupported_port_action = UnsupportedPortAction::Warn;
 
         ActionPlan action_plan;
+        // this needs to be kept alive because the ActionPlan has pointers into it
+        std::unique_ptr<IOverlayProvider> oprovider;
         if (manifest)
         {
             const CreateInstallPlanOptions create_options{
@@ -500,7 +502,7 @@ namespace vcpkg
                 extended_overlay_port_directories.builtin_overlay_port_dir.emplace(paths.builtin_ports_directory());
             }
 
-            auto oprovider =
+            oprovider =
                 make_manifest_provider(fs, extended_overlay_port_directories, manifest->path, std::move(manifest_scf));
             action_plan = create_versioned_install_plan(*verprovider,
                                                         *baseprovider,


dg0yt:
TBD: `error: unexpected switch: --allow-unsupported`

src/vcpkg/commands.depend-info.cpp
```diff
@@ -248,6 +255,8 @@ namespace vcpkg
         return s;
     }
 
+    // These command metadata must share "critical" values (switches, number of arguments).
```
dg0yt, line 258: The comment is copied, but actually I do not apply the same "number of arguments". Before I change the comment, maintainers might want to review the comment/pattern in `command.install.cpp`.

BillyONeal, line 258: This is copying the design of `install` in that the behavior totally changes based on whether we found a vcpkg.json. I'm pretty sure at this point we regret `install` having that behavior and the existence of a visible manifest creating such "spooky action at a distance" and we should have added `install-manifest` or something like that instead. Particularly given that very few customers use `vcpkg install` to install a manifest's dependencies rather than using CMake or MSBuild bindings where the command name passed to our exe is meaningless.

I can see an argument that we should keep doing what install was doing regardless for consistency with what we've already shipped.

src/vcpkg/commands.depend-info.cpp
```diff
@@ -399,31 +420,184 @@ namespace vcpkg
...
-        StatusParagraphs status_db;
         PackagesDirAssigner packages_dir_assigner{paths.packages()};
-        auto action_plan = create_feature_install_plan(
-            provider,
-            var_provider,
-            specs,
-            status_db,
-            packages_dir_assigner,
-            {nullptr, host_triplet, UnsupportedPortAction::Warn, UseHeadVersion::No, Editable::No});
+        const auto unsupported_port_action = Util::Sets::contains(options.switches, SwitchAllowUnsupported)
```
BillyONeal: Since there does appear to be a lot of copy/pasta here do you think anything is extractable/sharable between this and install?

BillyONeal:
> Changing behavior in classic mode for unsupported triplets, by introducing the `--allow-unsupported` switch. Alternative: Do not add the switch, just warn in both modes.

The existing behavior is to always warn because the entire command itself behaves like a dry run: it doesn't mutate anything so the potential time wastage concern we are trying to mitigate with `"supports"` doesn't apply here. So I would vote to keep the "just warn" behavior.

src/vcpkg/commands.depend-info.cpp
dg0yt: FTR I just play with isolated `vcpkg install`/`vcpkg export` to make my actual project CI build with cached exported artifacts. I'm not convinced that 40 min of `cmake` configuration are good UX.

From the `depend-info` perspective, users want it to behave it like `install`. So alternatives patterns are out-of-scope for this PR.

src/vcpkg/commands.depend-info.cpp
dg0yt: Maybe yes. Not sure about effects on diagnostics. And I don't want to break the `install` command ;-)

dg0yt:
> So I would vote to keep the "just warn" behavior.

I came to the same conclusion.

But I'm unhappy about the fact that the warning appears first in the output. I would always want to see it, but it is two pages away in the console output for my use cases with GDAL or Qt.

src/vcpkg/commands.depend-info.cpp
```diff
@@ -399,31 +420,204 @@ namespace vcpkg
...
+                    return dep.constraint.type != VersionConstraintKind::None;
+                }))
+            {
+                get_global_metrics_collector().track_define(DefineMetric::ManifestVersionConstraint);
+            }
+
+            if (!manifest_core.overrides.empty())
+            {
+                get_global_metrics_collector().track_define(DefineMetric::ManifestOverrides);
+            }
```
dg0yt: What I hadn't copied (introduced) to depend-info is: telemetry. How do you want to deal with it?

The functional result of the shared code is an install plan. Telemetry is a side effect. Here, the telemetry reports calculated intermediate results, so it can't be moved out of the shared code easily. The side effect could be optional, subject to an optional metrics_collector parameter.

dg0yt: @BillyONeal I would like to hear feedback for how to deal with telemetry. If it is easier for you to review a particular choice, you can say so.

BillyONeal: I don't think we should track telemetry here; my understanding is that the intent of that thing is to note which ports we know about being installed and this isn't an install

vicroms: Agree with Billy, we don't need to track telemetry for `depend-info` beyond that the command was used.

dg0yt: So choose the "optional metrics_collector parameter" implementation?

dg0yt: And which translation unit should take the shared code? It is neither pure command.install nor pure commands.depend-info nor pure dependencies - it is glue between the actual command's options and `create_versioned_install_plan`.

BillyONeal: I'm confused, I don't see any shared code here? Can't we just delete these two ifs?

dg0yt: @BillyONeal you asked for deduplication in https://github.com/microsoft/vcpkg-tool/pull/1909#discussion_r2801568686. I elaborated the duplicated code so that it is marked, and it only differs in the name of the result variable. The logical next step is to do the actual deduplication by moving it to separate function.

BillyONeal: I was thinking something closer to https://github.com/dg0yt/vcpkg-tool/pull/2

src/vcpkg/commands.depend-info.cpp
```diff
@@ -398,32 +419,205 @@ namespace vcpkg
...
+            auto baseprovider = make_baseline_provider(*registry_set);
+
+            auto extended_overlay_port_directories = paths.overlay_ports;
+            if (add_builtin_ports_directory_as_overlay)
+            {
+                extended_overlay_port_directories.builtin_overlay_port_dir.emplace(paths.builtin_ports_directory());
+            }
+
+            auto oprovider =
+                make_manifest_provider(fs, extended_overlay_port_directories, manifest->path, std::move(manifest_scf));
```
BillyONeal: This causes use-after-free because `action_plan` ends up having pointers into it. (I only found this out when tests failed on my local box and I fixed in https://github.com/dg0yt/vcpkg-tool/pull/2 )

PR #1910, opened 2026-02-10
Make z-applocal command cross-platform for PE dependency analysis

zynfly:
## Summary

Remove the `_WIN32` compilation guards from the `z-applocal` command, enabling it to compile and run on macOS and Linux hosts for cross-platform Windows PE dependency analysis.

## Motivation

This change follows from the review discussion on [microsoft/vcpkg#49890](https://github.com/microsoft/vcpkg/pull/49890), where @BillyONeal suggested:

> Maybe we should finally just wire up https://github.com/microsoft/vcpkg-tool/blob/main/src/vcpkg/commands.z-applocal.cpp one of these days :/

The `z-applocal` command's core logic — PE file parsing via `cofffilereader` and DLL dependency copying — has no actual Windows API dependency. The COFF/PE reader is already cross-platform. By removing the `_WIN32` restriction, cross-compilation workflows (e.g., `host=macOS/Linux, target=x64-mingw-dynamic`) can use the native C++ implementation instead of the PowerShell script path (`applocal.ps1` + `objdump`), which is more reliable and has no external dependencies.

### `include/vcpkg/commands.z-applocal.h`
- Remove outer `#if defined(_WIN32)` / `#endif` guards

### `src/vcpkg/commands.z-applocal.cpp`
- Remove outer `#if defined(_WIN32)` / `#endif` guards
- **Replace `MutantGuard`** (Windows named mutex via `CreateMutexW` / `WaitForSingleObject`) with **`ExclusiveFileLock`** — the existing cross-platform file lock infrastructure (`flock()` on POSIX, `CreateFileW` on Windows). Lock files are placed in the vcpkg temp directory (`$TMPDIR/vcpkg/` or `%TEMP%\vcpkg\`) using the same hash-based naming pattern as the original (`"vcpkg-applocal-" + SHA256(target_dir)`), avoiding any residual files in user build output directories.
- **tlog writing**: Wrap UTF-16 tlog output in `#if defined(_WIN32)` with a UTF-8 fallback for non-Windows (tlog is MSBuild-specific; cross-platform CMake builds won't pass `--tlog-file`)

### `src/vcpkg/commands.cpp`
- Move `CommandZApplocalMetadata` registration out of `#if defined(_WIN32)` block (`CommandZUploadMetricsMetadata` remains Windows-only)

## What is NOT changed
- `cofffilereader.h` / `cofffilereader.cpp` — already cross-platform, no `_WIN32` guards
- `CMakeLists.txt` — uses GLOB, no changes needed
- `vcpkg_copy_tool_dependencies.cmake` — wiring the CMake side to invoke `z-applocal` instead of `applocal.ps1` is a separate follow-up

## Testing

Compiled and tested on **3 platforms** with zero compilation errors:

| Platform | Compiler | Build | Command | PE Parsing | Unit Tests |
|----------|----------|-------|---------|------------|------------|
| macOS arm64 | AppleClang 17 | 229/229 ✅ | ✅ | ✅ | 518/519 passed |
| Linux arm64 (Ubuntu 22.04) | GCC 11.4 | 223/223 ✅ | ✅ | ✅ | 519/519 passed |
| Linux amd64 (Ubuntu 22.04) | GCC 11.4 | 223/223 ✅ | ✅ | ✅ | 519/519 passed |

The single failure on macOS is the `download_files` test (`downloads.cpp:191`) which received an HTTP 502 from the network. This is unrelated to this change and passes normally on Linux.

**Functional test**: Used a real PE32 DLL (`vcpkg-msvc-2013/release/test_dll.dll`) on all three platforms. The `z-applocal` command correctly parsed the PE Import Table and identified `MSVCR120.dll` and `KERNEL32.dll` as imported dependencies. Lock files are placed in the vcpkg temp directory — user build output directories remain clean.

### DIFF
--- a/src/vcpkg/commands.z-applocal.cpp
+++ b/src/vcpkg/commands.z-applocal.cpp
@@ -89,14 +89,18 @@ namespace
                            const Path& installed_bin_dir,
                            const Path& installed,
                            bool is_debug,
+#if defined(_WIN32)
                            WriteFilePointer&& tlog_file,
+#endif // ^^^ _WIN32
                            WriteFilePointer&& copied_files_log)
             : m_fs(fs)
             , m_deployment_dir(deployment_dir)
             , m_installed_bin_dir(installed_bin_dir)
             , m_installed(installed)
             , m_is_debug(is_debug)
+#if defined(_WIN32)
             , m_tlog_file(std::move(tlog_file))
+#endif // ^^^ _WIN32
             , m_copied_files_log(std::move(copied_files_log))
             , m_openni2_installed(m_fs.exists(m_installed / "bin/OpenNI2/openni2deploy.ps1", VCPKG_LINE_INFO))
             , m_azurekinectsdk_installed(
@@ -521,22 +525,17 @@ namespace
                     format_filesystem_call_error(ec, "copy_file", {source, target, "CopyOptions::update_existing"}));
             }
 
+#if defined(_WIN32)
             if (m_tlog_file)
             {
-#if defined(_WIN32)
                 const auto as_utf16 = Strings::to_utf16(source);
                 Checks::check_exit(VCPKG_LINE_INFO,
                                    m_tlog_file.write(as_utf16.data(), sizeof(wchar_t), as_utf16.size()) ==
                                        as_utf16.size());
                 static constexpr wchar_t native_newline = L'\n';
                 Checks::check_exit(VCPKG_LINE_INFO, m_tlog_file.write(&native_newline, sizeof(wchar_t), 1) == 1);
-#else  // ^^^ _WIN32 / !_WIN32 vvv
-                const auto& native = source.native();
-                Checks::check_exit(VCPKG_LINE_INFO,
-                                   m_tlog_file.write(native.c_str(), 1, native.size()) == native.size());
-                Checks::check_exit(VCPKG_LINE_INFO, m_tlog_file.put('\n') == '\n');
-#endif // ^^^ !_WIN32
             }
+#endif // ^^^ _WIN32
 
             if (m_copied_files_log)
             {
@@ -554,7 +553,9 @@ namespace
         Path m_installed_bin_dir;
         Path m_installed;
         bool m_is_debug;
+#if defined(_WIN32)
         WriteFilePointer m_tlog_file;
+#endif // ^^^ _WIN32
         WriteFilePointer m_copied_files_log;
         std::unordered_set<std::string> m_searched;
         bool m_openni2_installed;
@@ -569,7 +570,9 @@ namespace
     constexpr CommandSetting SETTINGS[] = {
         {SwitchTargetBinary, msgCmdSettingTargetBin},
         {SwitchInstalledBinDir, msgCmdSettingInstalledDir},
+#if defined(_WIN32)
         {SwitchTLogFile, msgCmdSettingTLogFile},
+#endif // ^^^ _WIN32
         {SwitchCopiedFilesLog, msgCmdSettingCopiedFilesLog},
     };
 } // unnamed namespace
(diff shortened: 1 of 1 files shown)


src/vcpkg/commands.z-applocal.cpp
```diff
@@ -32,34 +32,6 @@ namespace
         return fs.open_for_write(entry->second, VCPKG_LINE_INFO);
     }
 
-    struct MutantGuard
```
BillyONeal, line 35: This should remain an NT mutant on Windows; for this purpose mutants provide a number of advantages over using files, for example, they are automatically removed by the operating system when a process terminates or crashes holding the mutant while the file solution can force a user to fix afterwards.

zynfly: Okay, I'll revise it later.

zynfly:
I have revised the code, please review it. @BillyONeal

zynfly:
Conflicts resolved.

src/vcpkg/commands.z-applocal.cpp
```diff
@@ -101,6 +104,7 @@ namespace
             , m_magnum_installed(m_fs.exists(m_installed / "bin/magnum/magnumdeploy.ps1", VCPKG_LINE_INFO) ||
                                  m_fs.exists(m_installed / "bin/magnum-d/magnumdeploy.ps1", VCPKG_LINE_INFO))
             , m_qt_installed(m_fs.exists(m_installed / "plugins/qtdeploy.ps1", VCPKG_LINE_INFO))
+            , m_temp_dir(m_fs.create_or_get_temp_directory(VCPKG_LINE_INFO))
```
BillyONeal, line 112:
```suggestion
#if !defined(_WIN32)
            , m_temp_dir(m_fs.create_or_get_temp_directory(VCPKG_LINE_INFO))
#endif // ^^^ !_WIN32
```

src/vcpkg/commands.z-applocal.cpp
```diff
@@ -543,6 +559,7 @@ namespace
         bool m_azurekinectsdk_installed;
         bool m_magnum_installed;
         bool m_qt_installed;
+        Path m_temp_dir;
```
BillyONeal, line 566:
```suggestion
#if !defined(_WIN32)
        Path m_temp_dir;
#endif // ^^^ !_WIN32
```

src/vcpkg/commands.z-applocal.cpp
```diff
@@ -512,12 +521,19 @@ namespace
...
+#if defined(_WIN32)
                 const auto as_utf16 = Strings::to_utf16(source);
                 Checks::check_exit(VCPKG_LINE_INFO,
                                    m_tlog_file.write(as_utf16.data(), sizeof(wchar_t), as_utf16.size()) ==
                                        as_utf16.size());
                 static constexpr wchar_t native_newline = L'\n';
                 Checks::check_exit(VCPKG_LINE_INFO, m_tlog_file.write(&native_newline, sizeof(wchar_t), 1) == 1);
+#else  // ^^^ _WIN32 / !_WIN32 vvv
+                const auto& native = source.native();
+                Checks::check_exit(VCPKG_LINE_INFO,
```
BillyONeal: The whole point of 'tlog's is that they are always UTF-16 because that's the form MSBuild wants. I think this should be left alone. Customers who want the UTF-8 version can just use 'copied files log'

zynfly: Thanks for the explanation! I agree that tlog is a Windows/MSBuild-specific concept and should stay UTF-16.
Should I just remove the #else branch for tlog writing in this PR, or go further and make the entire m_tlog_file member and --tlog-file parameter Windows-only with #if defined(_WIN32)?

BillyONeal: Either one is OK with me. The reason I was a stickler about the temp file one is that that incurred an extra filesystem operation.

zynfly: resolved it.

PR #1911, opened 2026-02-10
[tools] Fix missing executable permission on downloaded tools for Unix

zynfly:
## Summary

On macOS/Linux, tools downloaded by `download_tool()` in `src/vcpkg/tools.cpp` can lack the executable bit (`0644` instead of `0755`), causing `Permission denied` (exit code 126) when vcpkg immediately attempts to run the tool to determine its version.

This affects:
- **Raw binary downloads** (no archive) — e.g., `coscli` on macOS/Linux always gets `0644` from HTTP download
- **Some archive extractions** — e.g., `powershell-core` `.tar.gz` loses execute permission during the extract-to-temp-then-rename process

The PowerShell Core cross-platform tool support was recently introduced in https://github.com/microsoft/vcpkg/pull/49910, adding `powershell-core` entries for macOS and Linux to `vcpkg-tools.json`. This makes the permission issue more visible, as `vcpkg fetch powershell-core` now fails on these platforms without this fix.

The CMake-side equivalent (`vcpkg_find_acquire_program.cmake`) already handles this correctly via `FILE_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE ...`, but the C++ `download_tool()` path had no such handling.

## Fix

Add `chmod(exe_path, 0755)` before `return exe_path` in `download_tool()`, covering both the raw-binary and archive code paths in a single call. The helper is gated behind `#if !defined(_WIN32)` and uses `Debug::println` for non-fatal failure reporting.

Changes are confined to a single file (`src/vcpkg/tools.cpp`, +18 lines).

## Test Results

Tested on macOS arm64 by comparing `vcpkg fetch <tool>` between the unpatched and patched binaries after clearing tool caches:

| Tool | Download Type | Unpatched | Patched |
|------|--------------|-----------|---------|
| coscli | raw binary | `0644` — Permission denied | `0755` — OK |
| powershell-core | .tar.gz | `0644` — Permission denied | `0755` — OK |
| node | .tar.gz | `0755` — OK (preserved) | `0755` — OK |
| azcopy | .zip | `0755` — OK (preserved) | `0755` — OK |

- No regressions on tools that already had correct permissions (chmod is idempotent)
- Unit tests: 518/519 passed (1 pre-existing network-dependent failure unrelated to this change)

## Consistency

This follows the same "chmod at placement time" pattern used throughout the vcpkg ecosystem:
- `vcpkg_find_acquire_program.cmake` — `file(COPY ... FILE_PERMISSIONS OWNER_EXECUTE ...)`
- `bootstrap.sh` — `chmod +x` before `mv`
- `applocal.py` — `chmod 755` after copying dependencies

### DIFF
--- a/src/vcpkg/tools.cpp
+++ b/src/vcpkg/tools.cpp
@@ -1052,6 +1052,7 @@ namespace vcpkg
                 return nullopt;
             }
 
+            fs.set_executable(context, exe_path);
             return exe_path;
         }
 
--- a/include/vcpkg/base/files.h
+++ b/include/vcpkg/base/files.h
@@ -334,6 +334,8 @@ namespace vcpkg
 
         virtual bool last_write_time(DiagnosticContext& context, const Path& target, int64_t new_time) const = 0;
 
+        virtual bool set_executable(DiagnosticContext& context, const Path& target) const = 0;
+
         using ReadOnlyFilesystem::current_path;
         virtual void current_path(const Path& new_current_path, std::error_code&) const = 0;
         void current_path(const Path& new_current_path, LineInfo li) const;
--- a/src/vcpkg/base/files.cpp
+++ b/src/vcpkg/base/files.cpp
@@ -4051,6 +4051,31 @@ namespace vcpkg
 #endif // ^^^ !_WIN32
         }
 
+        virtual bool set_executable(DiagnosticContext& context, const Path& target) const override
+        {
+#if defined(_WIN32)
+            // On Windows, executability is determined by file extension, not permissions.
+            (void)context;
+            (void)target;
+            return true;
+#else  // ^^^ _WIN32 // !_WIN32 vvv
+            if (::chmod(target.c_str(), 0755) != 0)
+            {
+                auto local_errno = errno;
+                context.report(
+                    DiagnosticLine{DiagKind::Error,
+                                   target,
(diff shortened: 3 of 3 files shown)


src/vcpkg/tools.cpp
```diff
@@ -22,11 +22,26 @@
...
+#endif // !_WIN32
+
 #include <fmt/ranges.h>
 
 namespace
 {
     using namespace vcpkg;
+
+#if !defined(_WIN32)
+    static void set_unix_executable_permission(const Path& exe_path)
```
BillyONeal: Please add

```
        virtual bool set_executable(DiagnosticContext& context, const Path& target) const = 0;
```

to `vcpkg::Filesystem` in `include/files.h` instead.

src/vcpkg/tools.cpp
```diff
@@ -1052,6 +1067,9 @@ namespace vcpkg
                 return nullopt;
             }
 
+#if !defined(_WIN32)
+            set_unix_executable_permission(exe_path);
+#endif // !_WIN32
```
BillyONeal:
```suggestion
            fs.set_executable(context, exe_path);
```

BillyONeal (changes requested):

BillyONeal (approved):
Thank you!

PR #1925, opened 2026-02-19
Disable metrics for vcpkg-in-vcpkg.

BillyONeal:
This should fix the crash we observed trying to do a tool release where vcpkg install invokes vcpkg x-download invokes vcpkg z-upload-metrics https://github.com/microsoft/vcpkg/pull/50068#issuecomment-3918024268


BillyONeal:
I *believe* this is a fix demonstration: https://dev.azure.com/vcpkg/public/_build/results?buildId=127447&view=logs&j=f79cfdd7-47a8-597f-8f57-dc3e21a8f2ad&t=73c5513e-c47c-57a4-379a-581f7e429f13

PR #1932, opened 2026-02-24
[telemetry] Add --use-test-data to z-upload-metrics

vicroms:
This let us send dummy data for testing purposes that is easily identifiable.

An alternative would be to check in a dummy payload file and use that as the payload-file parameter instead.


BillyONeal:
How does this differ from https://github.com/microsoft/vcpkg-tool/blob/main/src/vcpkg/commands.z-preregister-telemetry.cpp ?

PR #1933, opened 2026-02-26
Group feature test results by port.

BillyONeal:
I'm going through ~1200 failures in https://github.com/microsoft/vcpkg/pull/45331 and the diagnostics are ordered by the size of the `ActionPlan` even though generally baseline settings for a given port must be considered together which is inconvenient.

/cc @autoantwort

### DIFF
--- a/src/vcpkg/commands.test-features.cpp
+++ b/src/vcpkg/commands.test-features.cpp
@@ -41,11 +41,6 @@ namespace
     };
 
     using DiagnosticVec = std::vector<SpecDiagnostic>;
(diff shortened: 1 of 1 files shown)


src/vcpkg/commands.test-features.cpp
```diff
@@ -905,6 +949,8 @@ namespace vcpkg
             msg::println();
         }
 
+        binary_cache.wait_for_async_complete_and_join(); // make sure "all feature tests passed" is the last line
```
BillyONeal, line 947: This is technically a drive-by change.

JavierMatosD (approved):

PR #1935, opened 2026-03-03
Accept tags in portsdiff and z-changelog.

BillyONeal:
In https://github.com/microsoft/vcpkg-tool/pull/1842/ I broke portsdiff and z-changelog because the previous behavior checked only if git cat-file -t succeeded, not whether the target was actually a commit. By "fixing" that to actually check for commit that breaks tag inputs.

Using -e and ^{commit} was suggested by gpt-5.3-codex but it was unable to make the other changes.

Demonstration:

```console
PS C:\Dev\vcpkg> .\vcpkg.exe z-changelog 2026.01.16
error: Invalid commit id: 2026.01.16
PS C:\Dev\vcpkg> C:\Dev\vcpkg-tool\out\build\Win-x64-Debug-WithArtifacts\vcpkg.exe z-changelog bogus
error: Invalid commit id: bogus
note: "C:\Program Files\Git\cmd\git.exe" -C "C:\Dev\vcpkg" -c core.autocrlf=false cat-file -e "bogus^{commit}" failed with exit code 128
fatal: Not a valid object name bogus^{commit}

PS C:\Dev\vcpkg> C:\Dev\vcpkg-tool\out\build\Win-x64-Debug-WithArtifacts\vcpkg.exe z-changelog 2026.01.16
#### Total port count: 2765
#### Total port count per triplet (tested): LINK
.
.
.
```

(Also with color:

the same console output above color screenshot

)

### DIFF
--- a/include/vcpkg/base/git.h
+++ b/include/vcpkg/base/git.h
@@ -93,11 +93,6 @@ namespace vcpkg
                           const Path& destination,
                           StringView treeish);
 
-    Optional<bool> git_check_is_commit(DiagnosticContext& context,
-                                       const Path& git_exe,
-                                       GitRepoLocator locator,
-                                       StringView git_commit_id);
-
     Optional<std::string> git_merge_base(DiagnosticContext& context,
                                          const Path& git_exe,
                                          GitRepoLocator locator,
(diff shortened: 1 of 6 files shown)


BillyONeal:
As an aside all the git operations really should get the "include console output on failure" treatment but I only had to fix one here.

vicroms (approved):

PR #1936, opened 2026-03-03
Allow Terrapin Retrieval Tool env vars.

BillyONeal:
Terrpain is an internal Microsoft network isolation tool frequently used with `x-script` asset caching settings. Allow environment variables used to pass authentication data like which tentant to use to it.


dfederm:
I'm the internal customer who needs this :)

Note: tried to confirm from within a port file, but couldn't download cmake to even start executing the port :(

vicroms (approved):

Issue #1939, opened 2026-03-11
vcpkg install fails behind proxy with schannel error 0x80092012


sdn9002:
I am behind corporate Trend Micro proxy (TMWS) on Windows 11 24H2.
HTTP_PROXY, HTTPS_PROXY and NO_PROXY environment variables are configured correctly.
With the latest version of vcpkg, I get an error when running vcpkg install command.

C:\vcpkg>vcpkg install fmt
Computing installation plan...
A suitable version of cmake was not found (required v3.31.10).
Downloading https://github.com/Kitware/CMake/releases/download/v3.31.10/cmake-3.31.10-windows-x86_64.zip -> cmake-3.31.10-windows-x86_64.zip
error: curl operation failed with error code 35 (SSL connect error).
error: Not a transient network error, won't retry download from https://github.com/Kitware/CMake/releases/download/v3.31.10/cmake-3.31.10-windows-x86_64.zip
note: If you are using a proxy, please ensure your proxy settings are correct.

In the OS event log I have error from schannel - The certificate received from the remote server has not validated correctly. The error code is 0x80092012. The TLS connection request has failed. The attached data contains the server certificate.
The SSPI client process is vcpkg-2026-03-04 (PID: 27172).

Trend Micro proxy (TMWS) is generating "on-the-fly" certificates for GitHub that contain no revocation information at all.
Because there is no CRL (Certificate Revocation List) or OCSP URL inside the certificate, Windows schannel library fails.

vcpkg must ignore missing
---------------- Certificate CDP ----------------
No URLs "None" Time: 0 (null)

and not "hard fail". It should continue as the OpenSSL backend for curl does.
The version of vcpkg.exe from a few weeks ago does not have this problem, and install command works correctly.
The 5.21 MB version of vcpkg.exe works correctly, but the latest 6.27 MB version does not.

BillyONeal:
We can't use the OpenSSL backend because it doesn't obey a system's certificate configuration. If schannel won't validate the certificate, then browsers on this system should be broken too. How is that working?

sdn9002:
Most browsers use Chromium, which does not use Schannel. 
Even if Schannel is used for certain functions, I assume the necessary flags have been set so that the absence of a Certificate Revocation List does not cause a hard failure. 
[...]

PR #1950, opened 2026-03-19
Fix for schannel missing certificate revocation list errors with proxy server.

sdn9002:
#1939 
Fix for schannel missing certificate revocation list errors with proxy server.


BillyONeal:
This weakens security as described in the linked issue so we would be unable to merge this as is. An opt in switch may be acceptable.

PR #1980, opened 2026-04-08
Add new baseline-diff command

autoantwort:
If we use x-update-baseline we wanted to know which packages gets updated for our manifest file.


src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+} // unnamed namespace
+
+namespace vcpkg
+{
+    constexpr CommandMetadata CommandBaselineDiffMetadata{
+        "x-baseline-diff",
+        msgCmdBaselineDiffSynopsis,
+        {"vcpkg x-baseline-diff <old_commit_sha> <newer_commit_sha>",
+         "vcpkg x-baseline-diff $(git rev-parse 2026.02.27) $(git rev-parse 2026.03.18)"},
```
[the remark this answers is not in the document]
autoantwort: I must say that i am always very confused if such examples gets translated. Also only old and new would be translated here (The people that uses git also understand git without translations...).

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,182 @@
...
+            auto verprovider = make_versioned_portfile_provider(*registry_set);
+            auto baseprovider = make_baseline_provider(*registry_set);
+
+            auto extended_overlay_port_directories = paths.overlay_ports;
+
+            auto oprovider = make_manifest_provider(fs,
+                                                    extended_overlay_port_directories,
+                                                    manifest->path,
+                                                    std::make_unique<SourceControlFile>(manifest_scf->clone()));
+            PackagesDirAssigner packages_dir_assigner{paths.packages()};
```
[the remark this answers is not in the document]
autoantwort: This is intended, since the use a baseline we never need the overlay port provider which would ignore the baseline.

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,183 @@
...
+        return false;
+    }
+
+    void check_for_valid_sha(StringView sha)
+    {
+        if (sha.size() < 7 || !Util::all_of(sha, ParserBase::is_hex_digit_lower))
+        {
+            Checks::msg_exit_with_error(VCPKG_LINE_INFO, msgInvalidCommitId, msg::commit_sha = sha);
+        }
+    }
```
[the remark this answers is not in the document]
autoantwort: This is wrong. The sha is only used for `git show`

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,183 @@
...
+            auto& vec = action.request_type == RequestType::USER_REQUESTED ? user_requested : transitive;
+            if (oldIter == versions.end())
+            {
+                vec.push_back(fmt::format("{}: new: {}", action.spec.name(), newVersion));
+            }
+            else if (oldIter->second != newVersion)
+            {
+                vec.push_back(fmt::format("{}: {} -> {}", action.spec.name(), oldIter->second, newVersion));
+            }
+        }
```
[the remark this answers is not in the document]
autoantwort: I considered this, but decided against it to reduce noise.

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,183 @@
...
+            auto oldIter = versions.find(action.spec);
+            auto newVersion = action.version.to_string();
+            auto& vec = action.request_type == RequestType::USER_REQUESTED ? user_requested : transitive;
+            if (oldIter == versions.end())
+            {
+                vec.push_back(fmt::format("{}: new: {}", action.spec.name(), newVersion));
+            }
+            else if (oldIter->second != newVersion)
+            {
+                vec.push_back(fmt::format("{}: {} -> {}", action.spec.name(), oldIter->second, newVersion));
```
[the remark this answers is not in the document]
autoantwort: To my understanding it is very unlikely that we have different versions for host and target triplets, therefore I didn't include the triplet to reduce noise.

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,183 @@
...
+            plan[i] = create_versioned_install_plan(*verprovider,
+                                                    *baseprovider,
+                                                    *oprovider,
+                                                    var_provider,
+                                                    dependencies,
+                                                    manifest_core.overrides,
+                                                    toplevel,
+                                                    packages_dir_assigner,
+                                                    create_options)
+                          .value_or_exit(VCPKG_LINE_INFO);
```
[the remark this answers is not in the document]
autoantwort: You get them if you install a plan, I thing the warnings are not needed here.

src/vcpkg/commands.baseline-diff.cpp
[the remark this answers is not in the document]
BillyONeal: In general if there was "English" in the example we made it be translated, and if there was not we did not. So for instance, `remove` has one that is translated (`vcpkg remove <package name>...` because "package name" is translatable) and two that are not:

https://github.com/microsoft/vcpkg-tool/blob/4a0ae3efd85e7501e7c3a0e905cc9c2ada80269c/src/vcpkg/commands.remove.cpp#L174

So I believe in this case the one that says `<old_commit_sha>` should get translated but the one with `rev-parse` should not. (However I'm not sure we should show the rev-parse example at all as that syntax makes shell assumptions)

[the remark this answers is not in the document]
BillyONeal: I think we should just accept the tags here rather than needing the user to say rev-parse? (I fixed a similar bug in https://github.com/microsoft/vcpkg-tool/pull/1935 )

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,183 @@
...
+                msg::write_unlocalized_text(Color::none, line);
+                msg::write_unlocalized_text(Color::none, "\n");
+            }
+            msg::write_unlocalized_text(Color::none, "\n");
+            return true;
+        }
+        return false;
+    }
+
+    void check_for_valid_sha(StringView sha)
```
BillyONeal, line 58: Can this use check_commit_exists (from git.h) instead?

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,183 @@
...
+    static constexpr CommandMultiSetting multisettings[] = {
+        {SwitchXFeature, msgHelpTxtOptManifestFeature},
+    };
+
+    bool print_lines(LocalizedString&& header, std::vector<std::string>&& lines)
+    {
+        if (!lines.empty())
+        {
+            Util::sort_unique_erase(lines);
+            msg::println(header.append_raw(':'));
```
BillyONeal, line 46:
```suggestion
    bool print_lines(const msg::MessageT<>& header, std::vector<std::string>&& lines)
    {
        if (lines.empty())
        {
            return false;
        }
        
            Util::sort_unique_erase(lines);
            msg::print(msg::format(header).append_raw(":\n"));
```

That is: early return on the `false` case, and pass `msg::MessageT<>` so we don't bother formatting at all in the no lines case.

I could have swore we already had a utility function doing something like this floating around but I'm not finding it...

BillyONeal: I think I'm remembering `format_plan_block` in dependencies.cpp which isn't directly applicable here.

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,183 @@
...
+                                        Triplet host_triplet)
+    {
+        const auto* manifest = paths.get_manifest();
+        if (manifest == nullptr)
+        {
+            Checks::msg_exit_with_message(VCPKG_LINE_INFO, msgMissingManifestFile);
+        }
+        auto options = args.parse_arguments(CommandBaselineDiffMetadata);
+        check_for_valid_sha(options.command_arguments.at(0));
+        check_for_valid_sha(options.command_arguments.at(1));
```
BillyONeal, line 95:
```suggestion
        check_for_valid_sha(options.command_arguments[0]);
        check_for_valid_sha(options.command_arguments[1]);
```

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,183 @@
...
+        auto features = get_manifest_features(options, manifest_core, var_provider, toplevel, host_triplet);
+
+        auto dependencies = get_manifest_dependencies(*manifest_scf, features);
+
+        ActionPlan plan[2];
+        for (int i = 0; i < 2; ++i)
+        {
+            if (auto default_reg = configuration.config.default_reg.get())
+            {
+                default_reg->baseline = options.command_arguments.at(i);
```
BillyONeal, line 117:
```suggestion
                default_reg->baseline = options.command_arguments[i];
```

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,183 @@
...
+        auto manifest_scf = parse_manifest_scf_or_exit(*manifest, paths, is_default_builtin);
+
+        const auto& manifest_core = *manifest_scf->core_paragraph;
+        PackageSpec toplevel{manifest_core.name, default_triplet};
+        auto features = get_manifest_features(options, manifest_core, var_provider, toplevel, host_triplet);
+
+        auto dependencies = get_manifest_dependencies(*manifest_scf, features);
+
+        ActionPlan plan[2];
+        for (int i = 0; i < 2; ++i)
```
BillyONeal, line 113:
```suggestion
        for (size_t i = 0; i < 2; ++i)
```

to avoid any sugned/unsigned mismatch warnings from vector.

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,183 @@
...
+            auto verprovider = make_versioned_portfile_provider(*registry_set);
+            auto baseprovider = make_baseline_provider(*registry_set);
+
+            auto extended_overlay_port_directories = paths.overlay_ports;
+
+            auto oprovider = make_manifest_provider(fs,
+                                                    extended_overlay_port_directories,
+                                                    manifest->path,
+                                                    std::make_unique<SourceControlFile>(manifest_scf->clone()));
+            PackagesDirAssigner packages_dir_assigner{paths.packages()};
```
BillyONeal, line 137: :sigh: We really need to have a "prepare for build" step and separate types for  "I have a packages dir" and friends.

No change requested.

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,183 @@
...
+} // unnamed namespace
+
+namespace vcpkg
+{
+    constexpr CommandMetadata CommandBaselineDiffMetadata{
+        "x-baseline-diff",
+        msgCmdBaselineDiffSynopsis,
+        {"vcpkg x-baseline-diff <old_commit_sha> <newer_commit_sha>",
+         "vcpkg x-baseline-diff $(git rev-parse 2026.02.27) $(git rev-parse 2026.03.18)"},
+        Undocumented,
```
BillyONeal, line 75: I think we need a docs page for this.

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,183 @@
...
+                vec.push_back(fmt::format("{}: new: {}", action.spec.name(), newVersion));
+            }
+            else if (oldIter->second != newVersion)
+            {
+                vec.push_back(fmt::format("{}: {} -> {}", action.spec.name(), oldIter->second, newVersion));
+            }
+        }
+        bool any_changes = false;
+        any_changes |= print_lines(msg::format(msgDirectDependencies), std::move(user_requested));
+        any_changes |= print_lines(msg::format(msgTransitiveDependencies), std::move(transitive));
```
BillyONeal, line 175:
```suggestion
        any_changes |= print_lines(msgDirectDependencies, std::move(user_requested));
        any_changes |= print_lines(msgTransitiveDependencies, std::move(transitive));
```

(to go with the change to `print_lines` above)

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,183 @@
...
+        {
+            if (auto default_reg = configuration.config.default_reg.get())
+            {
+                default_reg->baseline = options.command_arguments.at(i);
+            }
+            else
+            {
+                RegistryConfig synthesized_registry;
+                synthesized_registry.kind = JsonIdBuiltin.to_string();
+                synthesized_registry.baseline = options.command_arguments.at(i);
```
BillyONeal, line 123:
```suggestion
                synthesized_registry.baseline = options.command_arguments[i];
```

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,183 @@
...
+        nullptr,
+    };
+
+    void command_baseline_diff_and_exit(const VcpkgCmdArguments& args,
+                                        const VcpkgPaths& paths,
+                                        Triplet default_triplet,
+                                        Triplet host_triplet)
+    {
+        const auto* manifest = paths.get_manifest();
+        if (manifest == nullptr)
```
BillyONeal, line 89: I'm not sure I agree with this design: Because the user is explicitly passing in multiple baseline SHAs, it seems to really be more of a classic mode command as other registries aren't accounted for.

It seems like what you actually wanted is effectively this same diff but before/after what `x-update-baseline` does?

If this stays manifest only, it seems like one of the SHAs should be the one already in the manifest?

src/vcpkg/commands.baseline-diff.cpp
BillyONeal: I see we can't do that in the general case because this kinda just blindly overwrites the baseline with a SHA and we don't necessarily have a git repo present for that.

However that makes this kinda broken because filesystem registries' baselines are not SHAs. We should only be checking that the input is a SHA when we know that registry has that constraint.

BillyONeal (changes requested):
I think the general *idea* of such a command makes sense.

Most of what is here is nitpicks but the one about 'why manifest only' and 'what about non-git registries' are kinda foundational.

Should we have output like this printed by `x-update-baseline` instead of/in addition to what you've done here?

This needs at least one end to end test.

Thanks for the new command submission!

src/vcpkg/commands.baseline-diff.cpp
BillyONeal: I applied this in https://github.com/autoantwort/vcpkg-tool/pull/2

src/vcpkg/commands.baseline-diff.cpp
BillyONeal: I applied this in https://github.com/autoantwort/vcpkg-tool/pull/2

BillyONeal (commented):
(paraphrased by @BillyONeal )
@vicroms: Anything that makes baselines more helpful and less obscure is a good thing
@AugP: This concept sounds like a good idea and I have no objections to it

src/vcpkg/commands.baseline-diff.cpp
autoantwort: > It seems like what you actually wanted is effectively this same diff but before/after what x-update-baseline does? 

Yes. I wanted to know what x-update-baseline does mean to me in terms of port versions, or what it would mean.

autoantwort: > If this stays manifest only, it seems like one of the SHAs should be the one already in the manifest? 

At least if there is no second one. You are right, this would be more developer friendly.

src/vcpkg/commands.baseline-diff.cpp
autoantwort: The code is registry.h/cpp could really benefit from some comments. What is a `LockFile` and `LockFile::lockdata` maps what to what?

include/vcpkg/base/message-data.inc.h
```diff
@@ -341,6 +341,10 @@ DECLARE_MESSAGE(BaselineConflict,
                 "",
                 "Specifying vcpkg-configuration.default-registry in a manifest file conflicts with built-in "
                 "baseline.\nPlease remove one of these conflicting settings.")
+DECLARE_MESSAGE(BaselineDiffNoChange,
+                (),
+                "commit is a git commit here",
+                "No package that would have been installed was updated between the two commits")
```
vicroms:
```suggestion
                "No changes to installable packages between baselines")
```

I'm not sure if `installable` conveys the meaning but I'd prefer a terse description.

include/vcpkg/base/message-data.inc.h
```diff
@@ -599,6 +603,11 @@ DECLARE_MESSAGE(CmdAddVersionOptOverwriteVersion, (), "", "Overwrites git-tree o
 DECLARE_MESSAGE(CmdAddVersionOptSkipFormatChk, (), "", "Skips the formatting check of vcpkg.json files")
 DECLARE_MESSAGE(CmdAddVersionOptSkipVersionFormatChk, (), "", "Skips the version format check")
 DECLARE_MESSAGE(CmdAddVersionOptVerbose, (), "", "Prints success messages rather than only errors")
+DECLARE_MESSAGE(CmdBaselineDiffSynopsis,
+                (),
+                "",
+                "Computes the set of packages installed for different baselines and compares the versions that "
+                "would be installed")
```
vicroms:
```suggestion
                "Computes version changes to installable packages between baselines")
```

include/vcpkg/base/message-data.inc.h
```diff
@@ -2180,6 +2190,7 @@ DECLARE_MESSAGE(MissingDependency,
                 (msg::spec, msg::package_name),
                 "",
                 "Package {spec} is installed, but dependency {package_name} is not.")
+DECLARE_MESSAGE(MissingManifestFile, (), "", "Could not find manifest file (vcpkg.json file)")
```
vicroms:
```suggestion
DECLARE_MESSAGE(MissingManifestFile, (), "", "Couldn't find a manifest file (vcpkg.json file)")
```

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,184 @@
...
+            auto oldIter = versions.find(action.spec);
+            auto newVersion = action.version.to_string();
+            auto& vec = action.request_type == RequestType::USER_REQUESTED ? user_requested : transitive;
+            if (oldIter == versions.end())
+            {
+                vec.push_back(fmt::format("{}: new: {}", action.spec.name(), newVersion));
+            }
+            else if (oldIter->second != newVersion)
+            {
+                vec.push_back(fmt::format("{}: {} -> {}", action.spec.name(), oldIter->second, newVersion));
```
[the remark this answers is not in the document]
autoantwort: Not necessary here, you can not have an overrides per triplet

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,184 @@
...
+            }
+        }
```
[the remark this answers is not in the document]
autoantwort: Imho this is not that interesting. I mainly want to know which versions changed.

include/vcpkg/base/message-data.inc.h
autoantwort: Maybe installed? Because the packages listed in the vcpkg.json are normally installed.

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,315 @@
...
+                if (!is_git_sha(old_baseline))
+                {
+                    Checks::msg_exit_with_error(VCPKG_LINE_INFO, msgInvalidCommitId, msg::commit_sha = old_baseline);
+                }
+                if (!is_git_sha(new_baseline))
+                {
+                    Checks::msg_exit_with_error(VCPKG_LINE_INFO, msgInvalidCommitId, msg::commit_sha = new_baseline);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
autoantwort: we only accept lower case shas

src/vcpkg/commands.baseline-diff.cpp
autoantwort: https://github.com/MicrosoftDocs/vcpkg-docs/pull/584

src/vcpkg/commands.baseline-diff.cpp
```diff
@@ -0,0 +1,309 @@
...
+                    return baseline.to_string();
+                }
+                auto maybe_resolved = git_resolve_to_full_sha(console_diagnostic_context, *git_exe, locator, baseline);
+                Checks::msg_check_exit(
+                    VCPKG_LINE_INFO, maybe_resolved.has_value(), msgInvalidGitRef, msg::value = baseline);
+                return *maybe_resolved.get();
+            };
+            old_baseline = resolve_baseline_ref(old_baseline_ref);
+            new_baseline = resolve_baseline_ref(new_baseline_ref);
+        }
```
BillyONeal, line 207: I don't understand why we are trying to look at a local git root when this is a manifest mode command. What the local git situation is is totally irrelevant to manifest mode.

BillyONeal:
>>It seems like what you actually wanted is effectively this same diff but before/after what x-update-baseline does?
>
>Yes. I wanted to know what x-update-baseline does mean to me in terms of port versions, or what it would mean.

The problem with this is that `x-update-baseline` will update all registries in the manifest and this only cares about the default one. Only talking about the default registry means we could be showing an inaccurate view of what changes.

Inside `x-update-baseline`, we have this data for every registry we are updating and can show you the diff between the old world order and the new world order. Since the thing you seem to be actually trying to get out of this is what that diff looks like, should we fix `x-update-baseline` to print that rather than trying to add a new difficult-to-specify command like this?

I think this kind of "compare these two SHAs" should be a classic mode universe thing (which is exactly what `portsdiff` already intends to do) and the kind of comparison you're trying to do here would be better to compare 2 different manifests instead.

That said, I have some sympathy for "99% of people only use one registry so maybe this does what they actually want most of the time" but it sounds like even you really want this from x-update-baseline instead?

BillyONeal:
I believe I implemented the outcome you were trying to get to here in https://github.com/microsoft/vcpkg-tool/pull/2057

PR #1982, opened 2026-04-08
file lock: Only print wait message once (Fixes #1913) 

### LINKED ISSUE #1913: vcpkg-tool spams "waiting to take filesystem lock on..."
Since #1842 `waiting to take filesystem lock on...` gets printed every second instead of once. 
If we now build our VS Solution its get printed **7871** times. 
Since a real fix (#362) has not been merged since years which would speed up our build significantly (if it was printed 7871 times that means that the processes were waiting for 7871/3600 = 2.1 hours), can the occurrence of the message reduced again?

autoantwort:
Fixes #1913


BillyONeal (commented):
Hmmm... I'm worried about this situation looking like we are hung, but I understand not wanting it to be every second.

Could we do something like exponential backoff with a maximum of, say, 10 minutes?

autoantwort:
> Could we do something like exponential backoff with a maximum of, say, 10 minutes?

Yes. Maybe I will first look into using something like https://github.com/microsoft/vcpkg/pull/41605 to fix the underlying problem

autoantwort:
closing in favor of https://github.com/microsoft/vcpkg/pull/51348

PR #1984, opened 2026-04-09
Accept libcurl.so as a provider of curl.

podsvirov:


BillyONeal (approved):

BillyONeal:
Thanks!

podsvirov:
@BillyONeal should we update [UnableToFindCurl](https://github.com/microsoft/vcpkg-tool/blob/179a69bb1971c4cf51ff593ab2354b45d96edf49/include/vcpkg/base/message-data.inc.h#L2879-L2883) message too?

BillyONeal:
Yeah that's probably a good idea.

PR #2004, opened 2026-04-30
Add OHOS (OpenHarmony / HarmonyOS) platform support

jobor:
This adds the `ohos` identifier for the HamonyOS / OpenHarmony platform. 

See https://github.com/microsoft/vcpkg/issues/51302 and https://github.com/microsoft/vcpkg/pull/51470


BillyONeal:
Do you have a screenshot or a photo of the end-to-end scenario working (e.g. `vcpkg install zlib` on OHOS)? We're generally happy to merge platforms someone is willing to add but because we don't know how to validate this ourselves and there are multiple parts like the toolchain that need to be in place for things to work. I *think* you covered all the pieces but if someone cares enough to add this hopefully they can actually try it :)

You might also want to write a PR adding this to the community supported documentation over in https://github.com/MicrosoftDocs/vcpkg-docs/blob/45b52d047aadf711d377f4e5ae6665282bb21876/vcpkg/concepts/supported-hosts.md?plain=1#L122

jobor:
> Do you have a screenshot or a photo of the end-to-end scenario working (e.g. `vcpkg install zlib` on OHOS)? We're generally happy to merge platforms someone is willing to add but because we don't know how to validate this ourselves and there are multiple parts like the toolchain that need to be in place for things to work. I _think_ you covered all the pieces but if someone cares enough to add this hopefully they can actually try it :)

Sure, here's a screenshot of installing zlib with the arm64-ohos triplet:
Screenshot_20260502_121702

That's with exactly this branch.

If you want, I could also provide vcpkg installation logs from Qt's CI which covers much more libraries, albeit with a few modified ports.
 
> You might also want to write a PR adding this to the community supported documentation over in https://github.com/MicrosoftDocs/vcpkg-docs/blob/45b52d047aadf711d377f4e5ae6665282bb21876/vcpkg/concepts/supported-hosts.md?plain=1#L122

Oh right! Done here: https://github.com/MicrosoftDocs/vcpkg-docs/pull/575

jobor:
@BillyONeal, is anything else required from our side?

BillyONeal (dismissed):
Sorry for the delay, I was in CUDA update hell and then VS update hell :)

Do we need to add something to:

https://github.com/microsoft/vcpkg/blob/3855d4d6b52353ee2e2f73044327e59d01a18c84/scripts/bootstrap.sh#L84

jobor:
> Sorry for the delay, I was in CUDA update hell and then VS update hell :)

Welcome back from these depths! :)

> Do we need to add something to:
> 
> https://github.com/microsoft/vcpkg/blob/3855d4d6b52353ee2e2f73044327e59d01a18c84/scripts/bootstrap.sh#L84

No, that's not needed. Apart from the tools that are already checked, one only needs the HarmonyOS SDK. And that one needs to be pointed at via an environment variable.

BillyONeal (approved):

vicroms (approved):

PR #2014, opened 2026-05-12
x-download: more output, error if no asset cache configured for --store

autoantwort:
Also allows `--store` and `--skip-sha`.
Currently one must call `vcpkg hash <file>` and use that output in `vcpkg x-download <file> --store --sha512 <sha>`.
This is redundant. This change directly allows `vcpkg x-download <file> --store --skip-sha512` which computed the sha itself and prints it.


autoantwort:
@BillyONeal This is ready to review :)

include/vcpkg/base/message-data.inc.h
```diff
@@ -2809,7 +2816,12 @@ DECLARE_MESSAGE(StartingFeatureTest,
                 (msg::value, msg::feature_spec),
                 "{value} is the position in the run, for example '1/4'",
                 "Feature Test [{value}] {feature_spec}")
-DECLARE_MESSAGE(StoreOptionMissingSha, (), "", "--store option is invalid without a sha512")
+DECLARE_MESSAGE(StoreAssetShaWithoutAssetCache,
+                (msg::sha, msg::path),
+                "",
+                "requested store of asset cache entry {sha} -> {path}, but no asset cache write URL is configured")
```
BillyONeal:
```suggestion
                "requested store of asset cache entry {sha} -> {path}, but no writable asset cache is configured")
```

This way the message is correct for backends that don't necessarily have a "url". (I think the current asset cache ones all do but some of the binary cache backends do not)

src/vcpkg/base/downloads.cpp
```diff
@@ -1213,16 +1213,24 @@ namespace vcpkg
...
-            return store_to_asset_cache(
-                context, raw_upload_url, sanitized_upload_url, asset_cache_settings.m_write_headers, file_to_put);
+            if (store_to_asset_cache(
+                    context, raw_upload_url, sanitized_upload_url, asset_cache_settings.m_write_headers, file_to_put))
+            {
+                context.statusln(
+                    msg::format(msgStoreAssetSuccess, msg::url = sanitized_upload_url, msg::path = file_to_put));
+                return false;
+            }
+            return true;
```
```suggestion
                return true;
            }
            return false;
```

If we're emitting a success status it needs to return success. (Conventionally we use boolean true for success because it results in consistent tests in callers for functions that return a pointer which is nullptr on failure)

BillyONeal (changes requested):

BillyONeal (approved):
Thanks!

PR #2032, opened 2026-05-27
Remove vcpkg-artifacts.

BillyONeal:
This must not be committed until after July 1 2026.


vcpkg-init/vcpkg-init.ps1
```diff
@@ -141,12 +120,9 @@ IF ERRORLEVEL 1 (
 
 SET Z_POWERSHELL_EXE=
 
-:: Install the doskey
-DOSKEY vcpkg-shell="%VCPKG_ROOT%\vcpkg-cmd.cmd" $*
-
 :: If there were any arguments, also invoke vcpkg with them
 IF "%1"=="" GOTO fin
-CALL "%VCPKG_ROOT%\vcpkg-cmd.cmd" %*
+  "%VCPKG_ROOT%\vcpkg.exe" %*
```
BillyONeal, line 125: I need to do some smoke testing after inserting into VS to confirm that vcpkg still works from the developer command prompt.

BillyONeal:
Some ARM folks contacted me out of band who are interested in taking this over because ARM is the only entity that ended up using this stuff, so we're going to hold off from landing this for a bit.

PR #2036, opened 2026-05-29
Fix collide on shared .partial extraction directory on POSIX

Erik-White:
`extract_archive_to_temp_subdirectory` (in src/vcpkg/archives.cpp) builds the temporary extraction directory as <path>.partial. On Windows a a per-PID suffix added, but onPOSIX every concurrent vcpkg process targets the same directory.

I believe this is the root cause of the issue that I have been seeing in vcpkg: https://github.com/microsoft/vcpkg/issues/50051
The directory clash is causing ninja to write to to stderr while waiting to take filesystem lock, then the `--x-stderr-status flag `makes vcpkg return exit 1 on ANY stderr output.

I tried to add a minimal test case to cover this but it's a little tricky since it involves concurrency

### DIFF
--- a/azure-pipelines/end-to-end-tests-dir/fetch.ps1
+++ b/azure-pipelines/end-to-end-tests-dir/fetch.ps1
@@ -115,3 +115,27 @@ if (-not $IsMacOS -and -not $IsLinux) {
         throw 'Couldn''t run resulting ninja'
     }
 }
+
+# Concurrent `vcpkg fetch ninja` invocations must all succeed: each process
+# extracts into a per-PID partial directory before atomically renaming into
+# place. Regression test for microsoft/vcpkg#50051.
+$env:VCPKG_DOWNLOADS = Join-Path $TestingRoot 'concurrent-fetch-downloads'
+Remove-Item -Recurse -Force $env:VCPKG_DOWNLOADS -ErrorAction SilentlyContinue
+
+$jobs = @()
+for ($i = 0; $i -lt 8; $i++) {
+    $jobs += Start-Job -ScriptBlock {
+        param($vcpkg, $vcpkgArgs)
+        & $vcpkg @vcpkgArgs
+        $LASTEXITCODE
+    } -ArgumentList @($VcpkgExe, ($commonArgs + @("fetch", "ninja", "--x-stderr-status")))
+}
+
+$results = $jobs | Wait-Job | Receive-Job
+$jobs | Remove-Job
+$failed = @($results | Where-Object { $_ -ne 0 })
+if ($failed.Count -gt 0) {
+    throw "Concurrent vcpkg fetch ninja: $($failed.Count) of 8 invocations failed (regression of microsoft/vcpkg#50051)"
+}
+
+Remove-Item env:VCPKG_DOWNLOADS
--- a/azure-pipelines/end-to-end-tests-dir/fetch.ps1
+++ b/azure-pipelines/end-to-end-tests-dir/fetch.ps1
@@ -118,24 +118,34 @@ if (-not $IsMacOS -and -not $IsLinux) {
 
 # Concurrent `vcpkg fetch ninja` invocations must all succeed: each process
 # extracts into a per-PID partial directory before atomically renaming into
-# place. Regression test for microsoft/vcpkg#50051.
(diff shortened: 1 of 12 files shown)


src/vcpkg/archives.cpp
```diff
@@ -250,16 +250,24 @@ namespace vcpkg
...
     }
 
+    Path extraction_partial_path(const Path& to_path)
+    {
+        Path partial = to_path + ".partial";
+#if defined(_WIN32)
+        partial += "." + std::to_string(GetCurrentProcessId());
+#else
+        partial += "." + std::to_string(getpid());
+#endif
```
BillyONeal:
```suggestion
#if defined(_WIN32)
        const auto pid = GetCurrentProcessId();
#else
        const auto pid = getpid();
#endif
        Path partial = to_path + fmt::format(".partial.{}", pid);
```

or perhaps this pid function should be extracted to system.h?

src/vcpkg-test/archives.cpp
```diff
@@ -16,3 +20,18 @@ TEST_CASE ("Testing guess_extraction_type", "[z-extract]")
...
+// Regression test for microsoft/vcpkg#50051: concurrent vcpkg processes must not collide
+// on the same `.partial` extraction directory. The per-PID suffix is what keeps them apart.
+TEST_CASE ("extraction_partial_path is per-process", "[z-extract]")
+{
+    using namespace vcpkg;
+#if defined(_WIN32)
+    const auto pid = std::to_string(GetCurrentProcessId());
+#else
+    const auto pid = std::to_string(getpid());
+#endif
```
BillyONeal: Another use of this suggests that it should be extracted.

Alternately, `extraction_partial_path` should accept the pid as an input to make it testable. If we actually want this test at all; there's an argument that the only test for a function like this is the same as its implementation...

azure-pipelines/end-to-end-tests-dir/fetch.ps1
```diff
@@ -115,3 +115,37 @@ if (-not $IsMacOS -and -not $IsLinux) {
         throw 'Couldn''t run resulting ninja'
     }
 }
+
+# Concurrent `vcpkg fetch ninja` invocations must all succeed: each process
+# extracts into a per-PID partial directory before atomically renaming into
+# place. Regression test for microsoft/vcpkg#50051. The archive is served from a
+# local asset cache so the eight concurrent jobs stay offline; they share a
+# downloads directory so they genuinely race on the extraction.
+$priorDownloads = if (Test-Path env:VCPKG_DOWNLOADS) { $env:VCPKG_DOWNLOADS } else { $null }
```
BillyONeal: I'm fairly certain you don't need to be this paranoid about resetting it to something else; we already intend to reset this here: https://github.com/microsoft/vcpkg-tool/blob/1c6ce3ac837e1697eefe166fbd2dba172790ed49/azure-pipelines/end-to-end-tests.ps1#L121

BillyONeal (commented):
Thank you!

src/vcpkg-test/archives.cpp
Erik-White: I think I'll just remove it since there wasn't really much to test there

src/vcpkg/archives.cpp
Erik-White: I found a simpler way to do it so I have just inlined it now

azure-pipelines/end-to-end-tests-dir/fetch.ps1
Erik-White: I'll just remove that then :+1:

BillyONeal (approved):
LMAO that we already had the helper. Thank you!

Erik-White:
Thank you for the quick response!
What is the usual release flow for vcpkg-tool? Would you be able to cut a release now so i can get the fix in vcpkg?

BillyONeal:
I have to follow https://github.com/microsoft/vcpkg-tool/blob/main/docs/vcpkg_tool_release_process.md -- we aim to do that at least once a month or if there is a particularly impactful change.

Erik-White:
Ok thanks, I'll just have to be patient then :smile:

PR #2049, opened 2026-06-24
Emit a PURL in generated SPDX SBOMs

### LINKED ISSUE #2046: Emit a PURL in generated SPDX SBOMs
vcpkg-tool writes an SPDX SBOM for each installed package at `<installed>/<triplet>/share/<port>/vcpkg.spdx.json`. The `SPDXRef-port` package has no `externalRefs`, so the SBOM contains no PURL or CPE. Scanners match a package to its advisories through those identifiers, so an installed vcpkg package can't be linked to its CVEs. microsoft/vcpkg#39254 reported this against Black Duck.

#1482 added a PURL and CPE to the SBOM but was closed without merging. It was blocked on the `vcpkg` PURL type not existing in the spec yet (see the comments there and package-url/purl-spec#245).

package-url/purl-spec#885 now registers that type, superseding #562 and closing #217. The shape is `pkg:vcpkg/<name>@<version>` with `port_version`, `repository_revision`, `triplet`, and `repository_url` qualifiers, for example `pkg:vcpkg/nlohmann-json@3.10.4?port_version=5`.

Once #885 is merged, vcpkg-tool can add an `externalRefs` entry of type `purl` to `SPDXRef-port`, built from the port name, version, and port-version it already records,plus the triplet and registry baseline it builds against. A CPE reference could follow but isn't needed for tools that match on PURL. I'm happy to open the SBOM-side PR once the PURL type lands.

References:
- package-url/purl-spec#885: registers the vcpkg PURL type (supersedes package-url/purl-spec#562, closes package-url/purl-spec#217)
- #1482: previous attempt at this change, closed unmerged
- microsoft/vcpkg#39254: original report (vcpkg SBOMs don't match in Black Duck)

JamieMagee:
Adds a `purl` externalRef to `SPDXRef-port` so scanners can link installed packages to their CVEs, e.g. `pkg:vcpkg/zlib@1.0?port_version=5&triplet=x64-linux` (plus `vcs_url` for the port git-tree and `repository_url` for non-default registries). Follows package-url/purl-spec#885.

Closes #2046.

### DIFF
--- a/include/vcpkg/base/contractual-constants.h
+++ b/include/vcpkg/base/contractual-constants.h
@@ -133,6 +133,12 @@ namespace vcpkg
     inline constexpr StringLiteral SpdxDocumentNamespace = "documentNamespace";
     inline constexpr StringLiteral SpdxDownloadLocation = "downloadLocation";
     inline constexpr StringLiteral SpdxElementId = "spdxElementId";
+    inline constexpr StringLiteral SpdxExternalReferenceCategory = "referenceCategory";
+    inline constexpr StringLiteral SpdxExternalReferenceCategoryPackageManager = "PACKAGE_MANAGER";
+    inline constexpr StringLiteral SpdxExternalReferenceLocator = "referenceLocator";
+    inline constexpr StringLiteral SpdxExternalReferenceType = "referenceType";
+    inline constexpr StringLiteral SpdxExternalReferenceTypePurl = "purl";
+    inline constexpr StringLiteral SpdxExternalRefs = "externalRefs";
     inline constexpr StringLiteral SpdxFileName = "fileName";
     inline constexpr StringLiteral SpdxGenerates = "GENERATES";
     inline constexpr StringLiteral SpdxLicenseConcluded = "licenseConcluded";
--- a/src/vcpkg-test/spdx.cpp
+++ b/src/vcpkg-test/spdx.cpp
@@ -450,7 +450,14 @@ TEST_CASE ("spdx maximum serialization", "[spdx]")
       "copyrightText": "NOASSERTION",
       "summary": "summary",
       "description": "description",
-      "comment": "This is the port (recipe) consumed by vcpkg."
+      "comment": "This is the port (recipe) consumed by vcpkg.",
+      "externalRefs": [
+        {
+          "referenceCategory": "PACKAGE_MANAGER",
+          "referenceType": "purl",
+          "referenceLocator": "pkg:vcpkg/zlib@1.0?port_version=5&triplet=arm-uwp&vcs_url=git%3A%2F%2Fsome-vcs-url"
+        }
+      ]
     },
     {
       "name": "zlib:arm-uwp",
@@ -604,7 +611,14 @@ TEST_CASE ("spdx minimum serialization", "[spdx]")
       "licenseConcluded": "NOASSERTION",
       "licenseDeclared": "NOASSERTION",
       "copyrightText": "NOASSERTION",
-      "comment": "This is the port (recipe) consumed by vcpkg."
+      "comment": "This is the port (recipe) consumed by vcpkg.",
+      "externalRefs": [
+        {
+          "referenceCategory": "PACKAGE_MANAGER",
+          "referenceType": "purl",
+          "referenceLocator": "pkg:vcpkg/zlib@1.0?triplet=arm-uwp"
+        }
+      ]
     },
     {
       "name": "zlib:arm-uwp",
@@ -719,7 +733,14 @@ TEST_CASE ("spdx concat resources", "[spdx]")
       "licenseConcluded": "NOASSERTION",
       "licenseDeclared": "NOASSERTION",
       "copyrightText": "NOASSERTION",
-      "comment": "This is the port (recipe) consumed by vcpkg."
+      "comment": "This is the port (recipe) consumed by vcpkg.",
+      "externalRefs": [
+        {
+          "referenceCategory": "PACKAGE_MANAGER",
+          "referenceType": "purl",
+          "referenceLocator": "pkg:vcpkg/zlib@1.0?triplet=arm-uwp"
+        }
+      ]
     },
     {
       "name": "zlib:arm-uwp",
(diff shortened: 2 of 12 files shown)


BillyONeal:
https://github.com/package-url/purl-spec/pull/245#issuecomment-2689044366 <-- Has this been addressed by anyone associated with the PURL project?

JamieMagee:
@BillyONeal [Philippe Ombredanne](https://github.com/pombredanne) picked it up on that thread last July: https://github.com/package-url/purl-spec/pull/245#issuecomment-3122041025. That became https://github.com/package-url/purl-spec#885.
 
It isn't merged yet, but I attended the PURL community meeting earlier today, and the maintainers signaled strong intent to move forward with the proposal. The meeting notes haven't been uploaded yet, but they'll be here: https://github.com/package-url/community/tree/main/meetings/purl-community
 
On provenance, the spec defines `repository_url`, `repository_revision`, and `triplet` qualifiers, so a registry port can record where it came from and at which baseline. Overlay ports and cross-machine reproducibility stay imperfect, and I don't think a URL fixes either. Happy to ask Philippe to weigh in here directly if that helps.

src/vcpkg-test/spdx.cpp
```diff
@@ -750,6 +771,46 @@ TEST_CASE ("spdx concat resources", "[spdx]")
...
+    SourceControlFileAndLocation scfl;
+    auto& scf = *(scfl.source_control_file = std::make_unique<SourceControlFile>());
+    auto& cpgh = *(scf.core_paragraph = std::make_unique<SourceParagraph>());
+    cpgh.name = "zlib";
+    cpgh.version = Version{"", 1};
+
+    InstallPlanAction ipa(
+        spec, scfl, packages_dir_assigner, RequestType::USER_REQUESTED, UseHeadVersion::No, Editable::No, {}, {}, {});
+    auto& abi = *(ipa.abi_info = AbiInfo{}).get();
+    abi.package_abi = "ABIHASH";
```
BillyONeal, line 799: This not being an actual SHA grinds my gears a bit but other tests here appear to be already doing that. No change requested.

BillyONeal:
Hi there! thanks again for your submission, I'm finally getting to it right now. I think it's a good starting point but I want to submit a PR against your PR that fills in more of the bits.

[Add new PURL type: vcpkg by JamieMagee · Pull Request #885 · package-url/purl-spec](https://github.com/package-url/purl-spec/pull/885/changes#diff-63de65c9f074a2fa787b880a41e26ec8c91f829dc36010c2dffa33a5486d153fR38)

* The bits there refer to `repository_revision` to be a baseline SHA, but that's not actually a meaningful bit when we're talking about an individual port. For instance, the port might not be at that baseline SHA version. To vcpkg the thing that *really* matters about ports is their git tree SHA rather than anything about the overall repo.
* The "Examples" bit there has a bunch of values like `repository_url=` which aren't in the qualifiers table above. Is there a shared PURL-wide list of fields we should be considering filling out, or is that just a missing bit from the table?

JamieMagee:
@BillyONeal On the qualifiers table: `repository_url` isn't a vcpkg-specific qualifier, which is why it's not listed there. PURL has a set of shared "known qualifiers" that apply to every type, documented in [common-qualifiers.md](https://github.com/package-url/purl-spec/blob/master/docs/common-qualifiers.md): `checksum`, `download_url`, `file_name`, `repository_url`, `vcs_url`, and `vers`. `repository_url` is the standard way to point at a non-default registry, so the examples use it for the azure-sdk fork and the `file://` overlay. It's shared, not missing from the table.

On `repository_revision`: The baseline is a registry-wide commit, and it doesn't pin an individual port, because the same port version is reachable from many baselines. The bit that identifies a specific port recipe is its git tree SHA, and we already compute it. vcpkg writes it as the SPDX download location today, e.g. `git+https://github.com/Microsoft/vcpkg@<git-tree>` (see `builtin_git_tree_spdx_location` in `paragraphs.cpp`).

That string is already in `vcs_url` shape, so I'd lean toward carrying the git tree there instead of in `repository_revision`. It needs no spec change, and it keeps `repository_revision` meaning what the spec says, a commit/baseline, rather than overloading it with a tree object. If we'd rather give the git tree its own named qualifier, that's a small follow-up to the type definition, but I don't think it's needed.

src/vcpkg-test/spdx.cpp
```diff
@@ -719,7 +733,14 @@ TEST_CASE ("spdx concat resources", "[spdx]")
...
       "licenseDeclared": "NOASSERTION",
       "copyrightText": "NOASSERTION",
-      "comment": "This is the port (recipe) consumed by vcpkg."
+      "comment": "This is the port (recipe) consumed by vcpkg.",
+      "externalRefs": [
+        {
+          "referenceCategory": "PACKAGE_MANAGER",
+          "referenceType": "purl",
+          "referenceLocator": "pkg:vcpkg/zlib@1.0"
+        }
```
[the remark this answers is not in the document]
JamieMagee: `PACKAGE_MANAGER` is correct for JSON. The v2.2.1 spdx-schema.json enum is `["OTHER", "SECURITY", "PACKAGE_MANAGER"]`; the hyphenated `PACKAGE-MANAGER` is the tag-value spelling, not JSON.

src/vcpkg-test/spdx.cpp
```diff
@@ -604,7 +611,14 @@ TEST_CASE ("spdx minimum serialization", "[spdx]")
...
```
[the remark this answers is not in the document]
JamieMagee: See other comment.

src/vcpkg-test/spdx.cpp
```diff
@@ -450,7 +450,14 @@ TEST_CASE ("spdx maximum serialization", "[spdx]")
...
       "summary": "summary",
       "description": "description",
-      "comment": "This is the port (recipe) consumed by vcpkg."
+      "comment": "This is the port (recipe) consumed by vcpkg.",
+      "externalRefs": [
+        {
+          "referenceCategory": "PACKAGE_MANAGER",
+          "referenceType": "purl",
+          "referenceLocator": "pkg:vcpkg/zlib@1.0?port_version=5"
+        }
```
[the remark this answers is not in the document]
JamieMagee: See other comment.

include/vcpkg/base/contractual-constants.h
```diff
@@ -133,6 +133,12 @@ namespace vcpkg
     inline constexpr StringLiteral SpdxDocumentNamespace = "documentNamespace";
     inline constexpr StringLiteral SpdxDownloadLocation = "downloadLocation";
     inline constexpr StringLiteral SpdxElementId = "spdxElementId";
+    inline constexpr StringLiteral SpdxExternalReferenceCategory = "referenceCategory";
+    inline constexpr StringLiteral SpdxExternalReferenceCategoryPackageManager = "PACKAGE_MANAGER";
+    inline constexpr StringLiteral SpdxExternalReferenceLocator = "referenceLocator";
+    inline constexpr StringLiteral SpdxExternalReferenceType = "referenceType";
+    inline constexpr StringLiteral SpdxExternalReferenceTypePurl = "purl";
+    inline constexpr StringLiteral SpdxExternalRefs = "externalRefs";
```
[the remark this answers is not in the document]
JamieMagee: See other comment.

src/vcpkg/spdx.cpp
```diff
@@ -334,6 +334,56 @@ std::string vcpkg::create_spdx_sbom(const InstallPlanAction& action,
...
+
+            qualifiers.push_back(
+                fmt::format("triplet={}", Strings::percent_encode(action.spec.triplet().canonical_name())));
+
+            // spdx_location is the SPDX PackageDownloadLocation: a VCS URL that pins the port's
+            // git-tree, which identifies the exact port recipe.
+            if (!spdx_location.empty())
+            {
+                qualifiers.push_back(fmt::format("vcs_url={}", Strings::percent_encode(spdx_location)));
+            }
```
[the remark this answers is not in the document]
JamieMagee: `vcs_url` is a standard PURL qualifier (it's in the spec's common-qualifiers, defined to hold an SPDX download location, which is what `spdx_location` is), so it isn't ad-hoc.

The `@<sha>` is a git-tree, not a commit. purl-spec#885 defines `repository_revision` as the registry baseline commit, so a tree SHA doesn't belong there. Carrying the git-tree in `vcs_url` is deliberate; @BillyONeal and I settled on it in the thread above.

`repository_url` is emitted only for non-default registries. The default microsoft/vcpkg (Builtin) is the type's `default_repository_url`, so it's omitted.

src/vcpkg-test/spdx.cpp
```diff
@@ -750,6 +771,116 @@ TEST_CASE ("spdx concat resources", "[spdx]")
...
+
+    InstallPlanAction ipa(
+        spec, scfl, packages_dir_assigner, RequestType::USER_REQUESTED, UseHeadVersion::No, Editable::No, {}, {}, {});
+    auto& abi = *(ipa.abi_info = AbiInfo{}).get();
+    abi.package_abi = "ABIHASH";
+
+    const auto sbom = create_spdx_sbom(ipa, {}, {}, {}, {}, "now", "https://test-document-namespace", {});
+    CHECK(sbom.find("pkg:vcpkg/zlib?port_version=1&triplet=arm-uwp") != std::string::npos);
+    CHECK(sbom.find("pkg:vcpkg/zlib@") == std::string::npos);
+}
```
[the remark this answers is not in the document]
JamieMagee: Done. Converted these to parse the SBOM and assert on `packages[0].externalRefs[0].referenceLocator` via a small `port_purl` helper, instead of substring matching.

BillyONeal:

src/vcpkg/spdx.cpp
```diff
@@ -334,6 +334,56 @@ std::string vcpkg::create_spdx_sbom(const InstallPlanAction& action,
...
+            if (cpgh.version.port_version != 0)
+            {
+                qualifiers.push_back(fmt::format("port_version={}", cpgh.version.port_version));
+            }
+
+            StringView spdx_location = scfl.spdx_location;
+            if (scfl.kind == PortSourceKind::Git && spdx_location.starts_with("git+"))
+            {
+                // spdx_location has the form "git+<repository_url>@<git-tree>". The default microsoft/vcpkg
+                // registry is Builtin, so only non-default registries reach this branch.
```
BillyONeal: That's not quite correct: someone can form that URI that points at the default registry. I think it would be better to preserve what the user said if they supplied a registry URL. Ideally we just record this when forming the scfl rather than trying to parse it back out after the fact.

src/vcpkg/spdx.cpp
```diff
@@ -334,6 +334,56 @@ std::string vcpkg::create_spdx_sbom(const InstallPlanAction& action,
         if (!cpgh.summary.empty()) obj.insert(JsonIdSummary, Strings::join("\n", cpgh.summary));
         if (!cpgh.description.empty()) obj.insert(JsonIdDescription, Strings::join("\n", cpgh.description));
         obj.insert(JsonIdComment, "This is the port (recipe) consumed by vcpkg.");
+        {
+            std::string purl = fmt::format("pkg:vcpkg/{}", Strings::percent_encode(action.spec.name()));
+            if (!cpgh.version.text.empty())
```
BillyONeal, line 339: Did you actually see this be empty in testing or are you just being defensive about it?

src/vcpkg/spdx.cpp
```diff
@@ -334,6 +334,56 @@ std::string vcpkg::create_spdx_sbom(const InstallPlanAction& action,
...
+            {
+                purl += fmt::format("@{}", Strings::percent_encode(cpgh.version.text));
+            }
+
+            // PURL qualifiers, emitted in canonical (alphabetical) key order.
+            std::vector<std::string> qualifiers;
+            if (cpgh.version.port_version != 0)
```
BillyONeal: 1. We should only claim anything about port-version if we claimed something about version, so this should be in the block ending on 342.
2. Should we consider always including port_version even when it is 0? It's an affirmation that we know that it is 0 while its absence could imply someone talking about any port-version.

src/vcpkg/spdx.cpp
```diff
@@ -334,6 +334,56 @@ std::string vcpkg::create_spdx_sbom(const InstallPlanAction& action,
...
+                }
+
+                StringView repository_url = spdx_location.substr(4, revision_sep - 4);
+                qualifiers.push_back(fmt::format("repository_url={}", Strings::percent_encode(repository_url)));
+            }
+
+            qualifiers.push_back(
+                fmt::format("triplet={}", Strings::percent_encode(action.spec.triplet().canonical_name())));
+
+            // spdx_location is the SPDX PackageDownloadLocation: a VCS URL that pins the port's
```
BillyONeal, line 359: Hmmm this is not technically true for us; it's a lot more "advisory" than that. For instance, you can clone vcpkg, and edit a port, and will emit `git+https://github.com/Microsoft/vcpkg#ports/<that port name>` in the location but that doesn't mean you could have ever reached that port content at that location.

I wonder if we should try to detect this. Maybe it doesn't matter.

src/vcpkg/spdx.cpp
```diff
@@ -334,6 +334,56 @@ std::string vcpkg::create_spdx_sbom(const InstallPlanAction& action,
...
+            // git-tree, which identifies the exact port recipe.
+            if (!spdx_location.empty())
+            {
+                qualifiers.push_back(fmt::format("vcs_url={}", Strings::percent_encode(spdx_location)));
+            }
+
+            if (!qualifiers.empty())
```
BillyONeal: We always have triplet so this if is dead

BillyONeal (commented):
I fixed some of the comments in https://github.com/JamieMagee/vcpkg-tool/pull/1 but I'm curious as to your thoughts on the cases where we don't have a git tree or proof that the port's content is on the remote.

src/vcpkg/spdx.cpp
JamieMagee: Defensive. `Version::text` can legitimately be empty (`Version::parse("#1")` gives empty text with `port_version` 1), which is the case Copilot flagged upthread. It's covered by the `spdx purl omits empty version` test.

src/vcpkg/spdx.cpp
JamieMagee: I'd lean toward omitting it at 0. Component Detection already emits vcpkg purls that way (`port_version` dropped when it's 0), so matching that keeps one canonical shape rather than two, and absence reads as 0 by convention. Happy to always emit it if you'd rather be explicit, but then we diverge from the purls already in the wild.

src/vcpkg/spdx.cpp
JamieMagee: We record the registry URL as `spdx_repository_url` when building the scfl instead of parsing it back out. `is_builtin_git_registry_url` also covers the case you raised, a git registry pointed at the default repo, so it doesn't get a redundant `repository_url`.

src/vcpkg/spdx.cpp
JamieMagee: Gone.

JamieMagee:
@BillyONeal I merged your PR, plus a small follow-up that drops the unused `NoAssertionTag` (every `PortLocation` goes through the `spdx_location`/`spdx_repository_url` constructor now).

On the advisory point: agreed, the location can't prove the content is reachable, and it probably shouldn't try to. `vcs_url`, `download_url`, and `repository_url` are all best-effort locators in PURL, not content hashes.

We do have a clean signal for when it's stronger, though. `builtin_git_tree_spdx_location` emits `git+...@<git-tree-sha>`, and the tree SHA pins the exact recipe. `builtin_port_spdx_location` emits `git+...#ports/<name>`, which is only a path and proves nothing, which is the edit-a-local-clone case you hit. Overlay and filesystem ports already emit no location.

So we could set `vcs_url` only for the `@<sha>` form and skip the `#ports/<name>` one, so it never claims more than it can back up. What do you think?

src/vcpkg/spdx.cpp
BillyONeal: Yeah sorry I wrote this comment before starting any of the changes that went into my PR against your PR. :)

BillyONeal:
> So we could set `vcs_url` only for the `@<sha>` form and skip the `#ports/<name>` one, so it never claims more than it can back up. What do you think?

The problem is that in the usual/expected case the builtin registry uses the path form, not the git sha form, so it would limit the utility of this feature.

(That is to say, checking out a git tree is relatively expensive, so we try really hard to just use the path on disk already there, and in such cases we only claim that we know the path, not the git tree)

This is what I meant by

> I wonder if we should try to detect this.

above.

Is it acceptable if for now we don't have much to claim there? I'm nervous to add git ops and similar which can be relatively expensive to produce a file I don't know if most of customers will read...

JamieMagee:
Okay, in that case I'm fine not claiming more than the path for now. It's the same string we already write as the SPDX `downloadLocation`, so `vcs_url` isn't a new claim, it just mirrors the location the SBOM already states. The flows that resolve a git tree (versioned/registry installs) still emit the `@<sha>` form, so we get the stronger locator wherever it's already cheap. If a scanner ever needs the guarantee, we can revisit.

So no code change, current behaviour stays. Is that okay?

BillyONeal (approved):

BillyONeal:
> So no code change, current behaviour stays. Is that okay?

Indeed. Thanks for your help!

PR #2052, opened 2026-06-29
[z-applocal] Only log dependency deployment details with --verbose

ricardoofnl:
Addresses microsoft/vcpkg#52550.

Since `vcpkg z-applocal` became the default DLL deployment path (microsoft/vcpkg#52315), builds emit a lot of noise: a "deploying dependencies" header for every target and dependency, plus a "skipped, up to date" line for every DLL that is already present.

The legacy `applocal.ps1` logged all of this via `Write-Verbose`, which MSBuild hides by default, so the old behavior was quiet. This restores parity by demoting those informational messages to debug-only output (`Debug::g_debugging`), while keeping the "copied file" line  which reports an actual deployment at normal verbosity.

Result: the first build still shows DLLs that are actually deployed; subsequent builds are quiet.

The existing end-to-end test (`z-applocalcpp.ps1`) only asserts on the "copied file" message, which is unchanged, so it continues to pass.

### DIFF
--- a/include/vcpkg/base/message-data.inc.h
+++ b/include/vcpkg/base/message-data.inc.h
@@ -910,6 +910,10 @@ DECLARE_MESSAGE(
     (),
     "",
     "Copies a binary's dependencies from the installed tree to where that binary's location for app-local deployment")
+DECLARE_MESSAGE(CmdZApplocalOptVerbose,
+                (),
+                "",
+                "Prints which dependencies are being processed and which are already up to date")
 DECLARE_MESSAGE(CmdZExtractExample1,
                 (),
                 "This is a command line, only the parts in <>s should be localized",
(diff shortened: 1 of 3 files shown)


BillyONeal:
I think this needs to be a new switch rather than using the intended-for-vcpkg-developers-only `--debug` (but otherwise I agree with the sentiment)

ricardoofnl:
> I think this needs to be a new switch rather than using the intended-for-vcpkg-developers-only `--debug` (but otherwise I agree with the sentiment)

can u review it again?

BillyONeal (approved):
I'm happy with this but @vicroms was touching this area recently so I'd like him to confirm.

vicroms (approved):

BillyONeal:
Thanks for the fix!

PR #2056, opened 2026-07-05
Change `GITHUB_SERVER_URL` to `GITHUB_API_URL` to allow submission of dependency graph data again

### LINKED ISSUE #2055: Dependency Graph upload not working
As stated in this [issue](https://github.com/microsoft/vcpkg/issues/52749) the upload of the dependency graph is failing. In the comment there is a proposed solution for the issue.

rwindegger:
This PR fixes #2055 and allows to upload dependency graph again on the public GitHub instance.

The `VcpkgCmdArguments` for `GITHUB_SERVER_URL` was removed as it was only used in the context where the submission of the dependency graph is happening. Instead the `GITHUB_API_URL` environment variable is used. The usage is adjusted as for GHES and github.com the `GITHUB_API_URL` contains the right URL for API requests so there is no need to append `/api/v3` anymore.


rwindegger:

BillyONeal (approved):
Thanks!

PR #2067, opened 2026-07-13
Add `"min-version"` to `vcpkg-tools.json`.

### LINKED ISSUE #2021: --x-abi-tools-use-exact-versions: numeric version comparison accepts suffixed versions, causing ABI mismatch
## Summary

When `--x-abi-tools-use-exact-versions` is used, vcpkg's version matching for ABI-sensitive tools uses a numeric comparison (`std::array<int, 3>`) that strips non-numeric suffixes. This causes it to accept a system tool whose version string differs from the one in `vcpkg-tools.json`, but the raw (non-normalized) version string is then recorded in the ABI hash — leading to binary cache misses across machines with different cmake builds.

## Reproduction

- System cmake reports `cmake version 4.2.3-msvc3` (Visual Studio 18's bundled cmake)
- `vcpkg-tools.json` specifies cmake version `"4.2.3"`
- `--x-abi-tools-use-exact-versions` is passed

**Expected behavior:** vcpkg should reject the system cmake because `"4.2.3-msvc3" != "4.2.3"`, download its own cmake 4.2.3, and record `cmake 4.2.3` in the ABI — matching CI and other machines.

**Actual behavior:** `parse_tool_version_string("4.2.3-msvc3")` returns `{4, 2, 3}`, which equals `{4, 2, 3}` from `vcpkg-tools.json`. The system cmake is accepted (no download). But the ABI records the raw string `cmake 4.2.3-msvc3`, causing cache misses against machines that have vanilla cmake `4.2.3`.

## Root cause in code

1. `CMakeProvider::get_version()` ([tools.cpp L462-483](https://github.com/microsoft/vcpkg-tool/blob/main/src/vcpkg/tools.cpp#L462-L483)) extracts the raw version string `"4.2.3-msvc3"` from `cmake --version` output.

2. `find_first_with_sufficient_version()` ([tools.cpp L937-954](https://github.com/microsoft/vcpkg-tool/blob/main/src/vcpkg/tools.cpp#L937-L954)) parses this into `std::array<int, 3>{4, 2, 3}` via `parse_tool_version_string()`, which strips the `-msvc3` suffix.

3. The exact version check ([tools.cpp L1140-1153](https://github.com/microsoft/vcpkg-tool/blob/main/src/vcpkg/tools.cpp#L1140-L1153)) compares numerically:
   ```cpp
   if (exact_version) {
       return actual_version[0] == min_version[0] && actual_version[1] == min_version[1] &&
              actual_version[2] == min_version[2];
   }
   ```
   `{4, 2, 3} == {4, 2, 3}` → match. System cmake is accepted.

4. But the raw version string is stored ([tools.cpp L954](https://github.com/microsoft/vcpkg-tool/blob/main/src/vcpkg/tools.cpp#L954)): `PathAndVersion{candidate, *version}` → `"4.2.3-msvc3"` ends up in the ABI hash.

## Suggested fix

When `exact_version` is true, the version comparison should be a **string comparison** against the version from `vcpkg-tools.json`, not a numeric comparison. `"4.2.3-msvc3" != "4.2.3"` would correctly reject the system cmake and force a download of the canonical version.

A string comparison is more appropriate here because:
- The flag is named "exact versions" — it should mean exact, not "numerically equivalent after stripping suffixes"
- A suffixed cmake (e.g. `-msvc3`) may have different behavior than the vanilla release
- The whole point of the flag is to guarantee ABI consistency across machines

## Impact

When the system cmake has a suffixed version string that matches numerically (e.g. `4.2.3-msvc3` vs `4.2.3`), all port ABI hashes differ between that machine and one using vanilla cmake — completely defeating binary caching. Every port must be rebuilt from source locally, even though the binaries are identical. For ports like QtWebEngine (which bundles Chromium), this can mean hours of unnecessary rebuilding.

The bug is latent — it only manifests when a system tool's version happens to be numerically equal to the version in `vcpkg-tools.json` but differs in its full version string. Visual Studio 18 bundles cmake `4.2.3-msvc3`, which now matches the `4.2.3` specified in the current `vcpkg-tools.json`.

## Environment

- vcpkg-tool version: `2026-04-08-e0612b42ce44e55a0e630f2ee9d3c533a63d8bc1`
- System cmake: `4.2.3-msvc3` (Visual Studio 18)
- OS: Windows

### LINKED ISSUE #52430: Using `--x-abi-tools-use-exact-versions` leads to extraction of `powershell-core` archive over and over again
**Describe the bug**
When using `--x-abi-tools-use-exact-versions`, `vcpkg` is downloading/extracting `powershell-core` archive over and over again on each `install` run.
Also, the extraction location does not reflect the actual version of the tool, so currently, `PowerShell-7.6.1-win-x64.zip` is downloaded but extracted to `downloads/tools/powershell-core-7.5.4-windows`.

**Environment**
- OS: Windows

**To Reproduce**
1. ./vcpkg install XYZ --x-abi-tools-use-exact-versions
2. same again

**Expected behavior**
* `powershell-core` archive is only downloaded and extracted the first time and simply used as is in the next runs.
* Tool archive gets extracted to a directory that reflects it's actual version instead of the expected minimum version. In the current case, it shall be `downloads/tools/powershell-core-7.6.1-windows`.

**Logs**
```
...
A suitable version of powershell-core was not found (required v7.5.4).
Extracting powershell-core...
...
```

**Additional context**
If the [`version` node in `scripts/vcpkg-tools.json`](https://github.com/microsoft/vcpkg/blob/58613d018f126ccaa9dac9a02b11142cb0f4f4ea/scripts/vcpkg-tools.json#L314) determines a minimum required version (as @BillyONeal stated [here](https://github.com/BillyONeal/vcpkg/pull/7#issuecomment-4710338620)), it should be named like that (e.g. `minimumVersion`). The `version` node shall reflect the downloaded version so `--x-abi-tools-use-exact-versions` doesn't get confused.

BillyONeal:
Before this change, "version" is intended to be the minimum version we will fetch from the system, but we also use it in the path for downloads which is confusing. Moreover, when combined with `--x-abi-tools-use-exact-versions`, the exact version we are looking for is not the downloaded one so we get stuck in the loop described in https://github.com/microsoft/vcpkg/issues/52430 .

"Resolves" https://github.com/microsoft/vcpkg/issues/52225 although that was not a bug at the time.
Resolves https://github.com/microsoft/vcpkg/issues/52430 .

Also fixed https://github.com/microsoft/vcpkg-tool/issues/2021 since I was in this area anyway.

GPT 5.6-Sol was used in preparing these changes.


BillyONeal:

include/vcpkg/tools.h
```diff
@@ -75,6 +83,100 @@ namespace vcpkg
                                               Path tools,
                                               RequireExactVersions abiToolVersionHandling);
 
+    struct ToolVersion
+    {
+        std::array<int, 3> cooked = {0, 0, 0}; // e.g. 24.8  or 2.7.4 or 1.0.0
```
Thomas1664, line 88: Did GPT come up with this name 😂?

BillyONeal: No, that was added in https://github.com/microsoft/vcpkg-tool/pull/1553 . I don't know which of the 3 of us credited in that PR actually came up with the name but it seems most likely to be Victor (@vicroms) as it is not present in Daniel's previous PR of which that was a continuation.

I could have sworn once upon a time that the Standard used this language but https://eel.is/c++draft/over.literal only contains "raw".

I think it is a good name.

vicroms (dismissed):

vicroms (approved):

PR #2070, opened 2026-07-14
Update vcpkg-tool-release-process.

BillyONeal:
This doc drifted a bit.

* `src/ConfigData/Packages/Redist/Setup.props` is now automatically updated by `VcpkgInsertionUtility`
* `2022` is out of date in several places.
* Artifacts activation in VS was removed in VS 2026 so remove that from the smoke test. (See also https://github.com/microsoft/vcpkg-tool/pull/2032 )

### DIFF
--- a/docs/vcpkg_tool_release_process.md
+++ b/docs/vcpkg_tool_release_process.md
@@ -3,8 +3,9 @@
 This document describes the acceptance criteria / process we use when doing a vcpkg-tool update,
 such as https://github.com/microsoft/vcpkg/pull/23757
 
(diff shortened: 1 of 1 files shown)


docs/vcpkg_tool_release_process.md
StephanTLavavej, line 29: Pre-existing: IMO it's bad to use a real value for a placeholder. This would be better as YYYY-MM-DD.

davidmrdavid (approved):

StephanTLavavej (approved):

docs/vcpkg_tool_release_process.md
BillyONeal: Don't want to reset over it but will consider it next time this gets updated. (And will probably make the same change elsewhere as I followed this pattern before in many places)

PR #2071, opened 2026-07-16
Standardize binary cache upload failure diagnostics

### LINKED ISSUE #28927: [vcpkg] Let it fails when uploading binary caching failed
```
error: curl failed to put file to https://xxx.com/packages/x64-osx-dynamic/zlib/1.2.13/23bb6610a3e542385afd48ad46945d93065459e7a691911c8ee88183779fc221.zip with exit code 6 and http code 0.
```

I config a http binary caching backend for my vcpkg, and is the only one, which is read-write. 

I'd like to let it fails when uploading failed.

xiaozhuai:
Fixes microsoft/vcpkg#28927

## Background

Binary cache upload failures currently produce provider-specific diagnostics.

For example, HTTP uploads may report:

```
warning: curl failed to PUT file
```
while S3 uploads may report:
```
warning: ... aws s3 cp ... failed with exit code 1
```

This makes it difficult for CI systems to reliably detect binary cache upload failures by parsing the log.

## Changes
Prepend a standardized message to existing binary cache upload failure diagnostics:
```
Binary cache upload failed:
```
For example:
```
warning: Binary cache upload failed: curl failed to PUT file ...
warning: Binary cache upload failed: ... aws s3 cp ... failed with exit code 1
```

The provider-specific diagnostic details are preserved.

## Example

```console
vcpkg install stb \
    --binarysource="http,https://not.found.com/{sha},readwrite" \
    --require-binary-cache-upload

### CODE REVIEW

File: include/vcpkg/binarycaching.h [hunk 1]
```diff
@@ -319,6 +323,8 @@ namespace vcpkg
         BackgroundWorkQueue<ActionToPush> m_actions_to_push;
         BinaryCacheSynchronizer m_synchronizer;
         std::thread m_push_thread;
+        size_t m_failed_uploads = 0;
+        bool m_require_upload_success = false;
```
BillyONeal: It doesn't seem we really need the binary cache to know this. It needs to track whether there were any failures but it really has no need to know about particular switches.

src/vcpkg/commands.test-features.cpp
```diff
@@ -958,13 +958,17 @@ namespace vcpkg
...
         int exit_code;
         if (diagnostics.empty())
         {
-            exit_code = EXIT_SUCCESS;
-            msg::println(msgAllFeatureTestsPassed);
+            exit_code = binary_cache_upload_requirement_satisfied ? EXIT_SUCCESS : EXIT_FAILURE;
+            if (binary_cache_upload_requirement_satisfied)
+            {
+                msg::println(msgAllFeatureTestsPassed);
+            }
```
BillyONeal: This shows a bit of a "hole" in that we don't actually record upload failures as a diagnostic. Rather than doing this in the `diagnostics.empty` block, it should probably be done before, e.g.

```
if (!binary_cache.wait_for_async_complete_and_join() && require_upload_success) {
     diagnostics.add_error(...)
}
```

then this following `diagnostics.empty` block becomes unchanged.

include/vcpkg/vcpkgcmdarguments.h
```diff
@@ -281,6 +282,7 @@ namespace vcpkg
 
         bool dependency_graph_enabled() const { return dependency_graph_feature.value_or(false); }
         bool binary_caching_enabled() const { return binary_caching.value_or(true); }
+        bool require_binary_cache_upload_enabled() const { return require_binary_cache_upload.value_or(false); }
```
BillyONeal: I don't really think this should be a "global" setting like this; the similar settings like `--only-binarycaching` are only on their respective command(s).

BillyONeal (commented):
1. The command suggests that we should stop/terminate if an upload fails, but it seems the only thing you're actually doing is changing the exit code... but I also don't have better name suggestions right now so I can't really suggest that that should be a blocker.
2. Why can't the CI system you're trying to address with this change look for the `error:` ?
3. Should we distinguish between "all uploads to a given endpoint failed" and "an upload to a given endpoint failed" given that the former suggests a configuration problem while the latter may be just general network instability about which a build can do nothing.
4. If/when we do "inclusive" caches like https://github.com/microsoft/vcpkg-tool/pull/1597 how do you think that should work with this?
5. It might be a good idea to do the audit for making the console output for binary caching clearer similar to that requested in https://github.com/microsoft/vcpkg-tool/pull/1864 before trying to change the exit code. I'm concerned about "the exit code is nonzero... WHY?!" being a later user's reaction.

xiaozhuai:
> it seems the only thing you're actually doing is changing the exit code

Yes, that's exactly what CI needs. When this happens, we need to be notified so we can trigger a rerun, rather than just marking everything as successful and causing everyone else's `--only-binarycaching` installs to fail.

> but I also don't have better name suggestions right now

Coming up with a name is hard, isn't it? This name was first suggested by @autoantwort 
I thought about it for a while, but I couldn't come up with anything better.
If you have a name that conveys the meaning more clearly, I'd be happy to change it.

> Why can't the CI system you're trying to address with this change look for the error: ?

This is very unreliable. We implement it by checking the logs for the message `curl failed to PUT file` because we use webdav backend. It worked fine at first, until we recently added a S3 backend.
The error log become
```
warning: /opt/homebrew/bin/aws s3 cp xxx.zip s3://xxx/xxx/xxxxxx.zip failed with exit code 1
```
Therefore, any attempt to match logs is unreliable.

> Should we distinguish between "all uploads to a given endpoint failed" and "an upload to a given endpoint failed" given that the former suggests a configuration problem while the latter may be just general network instability about which a build can do nothing.

I don't think it's necessary—at least for my use case, which is a typical scenario where I build on the central CI and upload the binary cache, while using `--only-binarycaching` everywhere else. It's enough to know that the upload failed; all you need to do is trigger a rerun.

If it’s a configuration error, there’s nothing we can do.
If it’s a network issue—such as a temporary fluctuation—we usually just need to rerun CI. That's what's happening in real world. I admit this kind of fluctuation is rare, but when it does happen, everyone's installation fails—which is terrible.

> If/when we do "inclusive" caches like https://github.com/microsoft/vcpkg-tool/pull/1597 how do you think that should work with this?

I believe that if the `write-back` fails and `--require-binary-cache-upload` is specified, a non-zero error code should also be returned.
Suppose a large organization has divided its binarycaching storage repositories by region to achieve the best user experience.

> It might be a good idea to do the audit for making the console output for binary caching clearer similar to that requested in https://github.com/microsoft/vcpkg-tool/pull/1864 before trying to change the exit code. I'm concerned about "the exit code is nonzero... WHY?!" being a later user's reaction.

When the `--require-binary-cache-upload` option is not provided, the default behavior actually remains unchanged. I agree with you that the output should be modified to clearly inform users that the error is caused by a failed upload while the `--require-binary-cache-upload` option was specified.

BillyONeal:
> This is very unreliable. We implement it by checking the logs for the message `curl failed to PUT file` because we use webdav backend. It worked fine at first, until we recently added a S3 backend. The error log become
> 
> ```
> warning: /opt/homebrew/bin/aws s3 cp xxx.zip s3://xxx/xxx/xxxxxx.zip failed with exit code 1
> ```
> 
> Therefore, any attempt to match logs is unreliable.

Can we fix the output such that it would be reliable?

xiaozhuai:
> Can we fix the output such that it would be reliable?

It's a workaround. Although this can work, I don't think it's better. Of course, if you insist, I'll update this PR.

BillyONeal:
> > Can we fix the output such that it would be reliable?
> 
> It's a workaround. Although this can work, I don't think it's better. Of course, if you insist, I'll update this PR.

I'm making it as a suggestion only because the "user visible" exposure is smaller and I think it's a net improvement for everyone rather than trying to do the particular thing you are trying to do.

Would you object to this being changed to fail/stop rather than continuing to build everything in the future?

xiaozhuai:
> Would you object to this being changed to fail/stop rather than continuing to build everything in the future?

Yes, I do object to this change. I strongly prefer to keep the current 'continue building' approach.
If we really have to do that, then maybe we need another option to enable the `fail-fast`.

In fact, I don’t think a fail-fast strategy makes sense here.

Especially when upload tasks are actually performed asynchronously; if an upload fails, the current log has already rolled over to a later port in the build. This can be very confusing.

Another point is that for users who have configured multiple binary caching sources, one source might experience a temporary outage while others remain operational; in such cases, `fail-fast` is inappropriate. This will become clearer once the `write-back` feature is added to the system. The `write-back` feature works very well in conjunction with the options added in this PR.

BillyONeal:
> Yes, I do object to this change. I strongly prefer to keep the current 'continue building' approach.
> If we really have to do that, then maybe we need another option to enable the `fail-fast`.

To me "require" means "fail immediately" and if you didn't object to that as a potential future change it was going to short circuit a lot of name bikeshedding.

How about `--binary-cache-upload-exit-code` or similar? Given that it only affects the exit code it would be nice to have something about that in there.

Also should this also apply to asset cache uploads? The systems are fairly similar.

vicroms:
@xiaozhuai 

As an alternative to checking the exit code of vcpkg, the tool could produce a `binary-cache-failure-log.txt` file that only gets created/populated when a package upload fails. Then you could check for the existence of that file and fail the CI pipeline.

xiaozhuai:
> To me "require" means "fail immediately"

I admit that "require" is misleading; I'll think of a better name.

> Also should this also apply to asset cache uploads? The systems are fairly similar.

They do look very similar, however, I personally don't think it's necessary to add this option for it. Because the asset cache will automatically fall back to origin url. Even if it ultimately fails, the build will fail instead of quietly continuing as if everything were fine.

xiaozhuai:
> As an alternative to checking the exit code of vcpkg, the tool could produce a binary-cache-failure-log.txt file that only gets created/populated when a package upload fails. Then you could check for the existence of that file and fail the CI pipeline.

IMHO, using exit code is more straightforward.

xiaozhuai:
How about replacing the boolean switch with a value-based option?
`--binarycaching-upload-failure=<ignore|exit-nonzero>`
`ignore` reports failures without affecting the exit code, which is the current behavior.
`exit-nonzero` would continue all builds and pending asynchronous uploads, then return a non-zero exit code if any upload failed.
This avoids implying `fail-fast` behavior and describes how upload failures are treated. A value-based option is also extensible: we could add fail-fast in the future without introducing another command-line option:
`--binarycaching-upload-failure=fail-fast`

autoantwort:
for the bikeshedding: `--exit-nonzero-if-upload-fails` and then maybe also fail if asset cache upload fails.

xiaozhuai:
>  also fail if asset cache upload fails.

IMHO, I don't care about asset cache upload failures, because they don't actually have any side effects.

xiaozhuai:
Reopen in #2105

PR #2072, opened 2026-07-16
Use canonical vcpkg PURLs in dependency snapshots

JamieMagee:
Uses the SPDX formatter for dependency snapshots, replacing `pkg:github/vcpkg/...` with canonical `pkg:vcpkg/...` PURLs. The detector version is now 2.0.0.


include/vcpkg/purl.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
```
BillyONeal: Can we put this into `dependencies.h` and `dependencies.cpp` rather than introducing a whole translation unit for this one function? It seems strongly associated with `InstallPlanAction`

BillyONeal (changes requested):

include/vcpkg/purl.h
JamieMagee: Moved `make_vcpkg_purl()` into `dependencies.h` and `dependencies.cpp`.

BillyONeal (approved):
Thanks!

PR #2073, opened 2026-07-16
Fix dependency snapshot relationships and correlators

JamieMagee:
Marks only user-requested packages as direct.

The old correlator included `GITHUB_RUN_ID`, so it changed on every workflow run. GitHub merged those submissions instead of replacing the previous one, which could leave stale dependencies in the graph. The new correlator uses the workflow, job, and triplets, giving each build configuration a stable identity across reruns.

### DIFF
--- a/src/vcpkg-test/dependencies.cpp
+++ b/src/vcpkg-test/dependencies.cpp
@@ -1,6 +1,8 @@
 #include <vcpkg-test/util.h>
 
 #include <vcpkg/base/contractual-constants.h>
+#include <vcpkg/base/json.h>
+#include <vcpkg/base/messages.h>
 
 #include <vcpkg/commands.set-installed.h>
 #include <vcpkg/dependencies.h>
@@ -2613,7 +2615,7 @@ TEST_CASE ("dependency graph API snapshot uses canonical purls", "[dependencies]
     };
     auto v = VcpkgCmdArguments::create_from_arg_sequence(nullptr, nullptr);
     v.imbue_from_fake_environment(envmap);
-    auto s = create_dependency_graph_snapshot(v, plan);
+    auto s = create_dependency_graph_snapshot(v, plan, Test::X86_WINDOWS, Test::X64_WINDOWS);
 
     CHECK(s.has_value());
     auto obj = *s.get();
@@ -2650,7 +2652,7 @@ TEST_CASE ("dependency graph API snapshot uses canonical purls", "[dependencies]
 
     CHECK(static_cast<int>(version) == 0);
     CHECK(id == "123");
-    CHECK(correlator == "test-build-x64-windows-x86-windows");
+    CHECK(correlator == "test-build-target-x86-windows-host-x64-windows");
     CHECK(sha == "abc123");
     CHECK(ref == "refs/heads/main");
     CHECK(name == "vcpkg");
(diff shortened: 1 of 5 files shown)


src/vcpkg/commands.set-installed.cpp
```diff
@@ -60,12 +61,24 @@ namespace vcpkg
         const auto github_run_id = args.github_run_id.get();
         if (github_ref && github_sha && github_job && github_workflow && github_run_id)
         {
+            std::vector<std::string> triplets;
+            for (const auto& action : action_plan.install_actions)
+            {
+                triplets.push_back(action.spec.triplet().canonical_name());
+            }
+            Util::sort_unique_erase(triplets);
```
BillyONeal: There will only ever be at most 2 triplets, the host and the target one, so making a whole duplicate list to keep 1 or 2 seems overkill.

Also this sorts the triplets which is probably not what you want; it probably would be better to pick a canonical host/target order? For example, on an amd64 Windows machine, you can easily have host/target set to x86/x64 and vice versa, and the host=x64/target=x86 run should probably be considered different than the host=x86/target=x64 one.

src/vcpkg/commands.set-installed.cpp
```diff
@@ -104,7 +117,10 @@ namespace vcpkg
                 const auto& pkg_url = found->second;
                 Json::Object resolved_item;
                 resolved_item.insert(JsonIdPackageUnderscoreUrl, pkg_url);
-                resolved_item.insert(JsonIdRelationship, Json::Value::string(JsonIdDirect));
+                resolved_item.insert(JsonIdRelationship,
+                                     Json::Value::string(action.request_type == RequestType::USER_REQUESTED
+                                                             ? JsonIdDirect
+                                                             : JsonIdIndirect));
```
BillyONeal, line 115: I think we need something of an end to end test with a real manifest here; I'm pretty sure that we do not track `USER_REQUESTED` exactly as you might intend here. Or at least, currently it isn't load bearing on anything other than pretty console output and you're making it matter here.

src/vcpkg/commands.set-installed.cpp
BillyONeal, line 1: I understand this might not have been the problem you initially came here to solve but it is it right that plain `install` (for classic mode customers) makes no attempt to write this back? Someone who does `vcpkg install zlib` in their Action still depends on zlib even if they didn't write it in a `vcpkg.json`.

src/vcpkg-test/dependencies.cpp
```diff
@@ -2583,7 +2583,7 @@ TEST_CASE ("dependency graph API snapshot uses canonical purls", "[dependencies]
     InstallPlanAction install_a({"a", Test::X86_WINDOWS},
                                 scfl_a,
                                 packages_dir_assigner,
-                                RequestType::AUTO_SELECTED,
+                                RequestType::USER_REQUESTED,
```
BillyONeal, line 2588: Can you describe why this change was made? It's not typical for the user requested ones to be "early" in the graph.

src/vcpkg/commands.set-installed.cpp
BillyONeal: Also this only considers `install_actions` but not `already_installed` which seems wrong. (It *might* be the case that this is one of the "there will be no already_installed" cases but if so we might want an assert or similar)

JamieMagee: Changed this to use the target and host triplets passed into the snapshot builder, in `target-...-host-...` order. This avoids inferring them from either `install_actions` or `already_installed`. I also added a test with the roles swapped.

src/vcpkg/commands.set-installed.cpp
JamieMagee: Added a test that parses a real `vcpkg.json`, runs the resulting dependencies through `create_versioned_install_plan()`, and checks the snapshot direct/indirect relationships and dependency edge.

src/vcpkg/commands.set-installed.cpp
JamieMagee: Agreed, classic mode is a gap. I kept it out of this PR because the install plan only covers the current transaction, and the installed database does not retain the original user-requested roots. Submitting that plan would produce an incomplete graph, so classic mode needs a separate design.

src/vcpkg-test/dependencies.cpp
JamieMagee: I changed this to cover both `direct` and `indirect` relationships in the snapshot serializer test. It was not meant to represent the resolver's ordering.

The new parsed-manifest test covers the real plan ordering, and I'll reorder this synthetic plan so the host dependency comes before the user-requested package.

BillyONeal:

BillyONeal (approved):
Thanks!

PR #2074, opened 2026-07-16
Add Git tree gitoids to SPDX SBOMs

JamieMagee:
Upgrades generated SBOMs to SPDX 2.3 and adds `gitoid:tree:sha1` references for versioned registry ports. Registry tree hashes are validated before use.


BillyONeal:

src/vcpkg/registries.cpp
```diff
@@ -35,9 +36,22 @@ namespace
...
     };
+
+    Optional<std::string> GitTreeStringDeserializer::visit_string(Json::Reader& r, StringView sv) const
+    {
+        if (sv.size() == 40 && std::all_of(sv.begin(), sv.end(), ParserBase::is_hex_digit_lower))
+        {
+            return sv.to_string();
+        }
+
+        r.add_generic_error(type_name(), msg::format(msgInvalidGitObjectSha, msg::sha = sv));
```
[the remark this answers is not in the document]
BillyONeal: I disagree with this: localization will still be accurate with "sha"; there's no way one of the translated messages will be wrong if a SHA512 is there rather than a SHA1 (unless we explicitly describe it that way in the english)

include/vcpkg/base/message-data.inc.h
```diff
@@ -1998,6 +1998,10 @@ DECLARE_MESSAGE(InvalidFormatString,
                 (msg::actual),
                 "{actual} is the provided format string",
                 "invalid format string: {actual}")
+DECLARE_MESSAGE(InvalidGitObjectSha,
+                (msg::sha),
+                "{sha} is the invalid Git object SHA",
```
BillyONeal:
```suggestion
                "",
```

Doesn't need explained because {sha} is already a named replacement with docs.

include/vcpkg/base/message-data.inc.h
```diff
@@ -1998,6 +1998,10 @@ DECLARE_MESSAGE(InvalidFormatString,
...
+                "Invalid Git object SHA: {sha}. Expected a 40-character lowercase hexadecimal string.")
```
BillyONeal:
```suggestion
                "invalid git object SHA: {sha}. Expected a 40-character lowercase hexadecimal string.")
```

This thing gets printed (morally) after `<file>: error: ` so it should start with lowercase, and git is conventionally spelled lowercase even if it's a proper noun because Unix.

include/vcpkg/sourceparagraph.h
```diff
@@ -216,11 +216,13 @@ namespace vcpkg
         explicit PortLocation(const Path& port_directory,
                               std::string&& spdx_location,
                               std::string&& spdx_repository_url,
-                              PortSourceKind kind);
+                              PortSourceKind kind,
+                              StringView git_tree = {});
```
BillyONeal:
```suggestion
                              StringView git_tree);
```

If it's important enough that we're including it in output we should be making an explicit choice to provide it or not in all cases. (Forgetting should be a build error, not silently wrong output)

Ditto below.

src/vcpkg/registries.cpp
```diff
@@ -35,9 +36,22 @@ namespace
...
     {
         LocalizedString type_name() const override { return msg::format(msgAGitObjectSha); }
+        Optional<std::string> visit_string(Json::Reader& r, StringView sv) const override;
 
         static const GitTreeStringDeserializer instance;
     };
+
+    Optional<std::string> GitTreeStringDeserializer::visit_string(Json::Reader& r, StringView sv) const
+    {
+        if (sv.size() == 40 && std::all_of(sv.begin(), sv.end(), ParserBase::is_hex_digit_lower))
```
BillyONeal: Please use `is_git_sha` rather than implementing this again.

include/vcpkg/base/message-data.inc.h
JamieMagee: Removed the translator comment and regenerated `locales/messages.json`.

include/vcpkg/base/message-data.inc.h
JamieMagee: Changed this to `invalid git object SHA: ...` and regenerated the message map.

include/vcpkg/sourceparagraph.h
JamieMagee: Removed the default from both constructors. Every call site now passes either the known tree or an explicit empty `StringView{}`.

src/vcpkg/registries.cpp
JamieMagee: Switched this to `is_git_sha()` and removed the now-unused parse include.

JamieMagee: Done.

include/vcpkg/sourceparagraph.h
JamieMagee: Done.

include/vcpkg/base/message-data.inc.h
JamieMagee: Done.

include/vcpkg/base/message-data.inc.h
JamieMagee: Done.

BillyONeal (approved):
Thanks!

PR #2087, opened 2026-07-29
Add privacy and telemetry notices to the help command

vicroms:


azure-pipelines/end-to-end-tests-dir/notices.ps1
```diff
@@ -0,0 +1,35 @@
...
+$topics = Run-VcpkgAndCaptureOutput help topics
+Throw-IfFailed
+foreach ($topic in @('privacy', 'telemetry')) {
+    if (-not $topics.Contains($topic)) {
+        throw "$topic was not listed as a help topic"
+    }
+}
+
+Run-Vcpkg privacy-notice
+Throw-IfNotFailed
```
BillyONeal: Why have a test about something we do not implement?

azure-pipelines/end-to-end-tests-dir/notices.ps1
BillyONeal, line 1: This file tests `vcpkg help`; should it be called `help.ps1`?

azure-pipelines/end-to-end-tests-dir/notices.ps1
```diff
@@ -0,0 +1,35 @@
...
+}
+
+$expectedTelemetry = @'
+vcpkg collects usage data in order to help us improve your experience.
+The data collected by Microsoft is anonymous.
+You can opt-out of telemetry by re-running the bootstrap-vcpkg script with -disableMetrics, passing --disable-metrics to vcpkg on the command line, or by setting the VCPKG_DISABLE_METRICS environment variable.
+
+Read more about vcpkg telemetry at https://learn.microsoft.com/vcpkg/about/privacy
+Read the Microsoft Privacy Statement at https://go.microsoft.com/fwlink/?LinkId=521839
+'@ -replace "`r`n", "`n"
```
BillyONeal: This `-replace` suggests that this file is incorrectly CRLFs

azure-pipelines/end-to-end-tests-dir/notices.ps1
```diff
@@ -0,0 +1,35 @@
...
+
+$expectedPrivacy = "Read the Microsoft Privacy Statement at https://go.microsoft.com/fwlink/?LinkId=521839`n"
+$actualPrivacy = Run-VcpkgAndCaptureOutput help privacy
+Throw-IfFailed
+if ($actualPrivacy -cne $expectedPrivacy) {
+    throw "help privacy output did not match.`nExpected: $expectedPrivacy`nActual: $actualPrivacy"
+}
+
+$expectedTelemetry = @'
+vcpkg collects usage data in order to help us improve your experience.
```
BillyONeal: I'm not sure it makes sense to check the exact output here; it seems like that would make the test brittle against future modifications that are otherwise legitimate. However it does make sense to ensure that `vcpkg help <a topic in the topics list>` prints something other than "I don't know what that is". Should we remove this and just loop through the topics currently gathered on line 26 and assert that running `vcpkg help $topic` does not fail?

BillyONeal (changes requested):
The only thing I'm truly "request changes" for is the CRLFs; the rest of my comments are nitpicks.

BillyONeal (approved):
Thanks!

PR #2088, opened 2026-07-29
Parallelize ObjectStorageProvider downloads

BillyONeal:
This is https://github.com/microsoft/vcpkg-tool/pull/1485 / https://github.com/microsoft/vcpkg-tool/pull/1392 now that we have the DiagnosticContext infrastructure to make it not painful.

Run up to eight external object-storage downloads concurrently while respecting lower VCPKG_MAX_CONCURRENCY values. Each worker writes to its own result slot and FullyBufferedDiagnosticContext; after all workers finish, diagnostics are replayed on the calling thread in action order, avoiding concurrent console I/O.

Add a bounded execute_in_parallel overload without changing existing callers, plus regression coverage for its concurrency ceiling.

### DIFF
--- a/src/vcpkg/binarycaching.cpp
+++ b/src/vcpkg/binarycaching.cpp
@@ -36,6 +36,7 @@ namespace
 {
     // The length of an ABI in the binary cache
     static constexpr size_t ABI_LENGTH = 64;
+    static constexpr size_t OBJECT_STORAGE_DOWNLOAD_CONCURRENCY = 8;
 
     struct ConfigSegmentsParser : ParserBase
     {
@@ -1030,12 +1031,12 @@ namespace
                           View<const InstallPlanAction*> actions,
                           Span<Optional<ZipResource>> out_zip_paths) const override
         {
-            for (size_t idx = 0; idx < actions.size(); ++idx)
-            {
+            std::vector<FullyBufferedDiagnosticContext> diagnostic_contexts(actions.size());
+            execute_in_parallel(actions.size(), OBJECT_STORAGE_DOWNLOAD_CONCURRENCY, [&](size_t idx) {
                 auto&& action = *actions[idx];
                 const auto& abi = action.package_abi_or_exit(VCPKG_LINE_INFO);
                 auto tmp = make_temp_archive_path(m_buildtrees, action.spec, abi);
-                WarningDiagnosticContext wdc{context};
+                WarningDiagnosticContext wdc{diagnostic_contexts[idx]};
                 auto res = m_tool->download_file(wdc, make_object_path(m_prefix, abi), tmp);
                 if (auto cache_result = res.get())
                 {
@@ -1044,6 +1045,11 @@ namespace
                         out_zip_paths[idx].emplace(std::move(tmp), RemoveWhen::always);
                     }
                 }
+            });
+
+            for (auto&& diagnostic_context : diagnostic_contexts)
+            {
+                std::move(diagnostic_context).report_to(context);
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/vcpkg/binarycaching.cpp
Crzyrndm, line 1076: https://github.com/microsoft/vcpkg-tool/pull/1392#issuecomment-2111370045

My test at the time included precheck in the parallelisation. From memory `stat` also made network calls with the S3 backend and this might have been the main accelerator in the cache miss case?

Crzyrndm (commented):
I'm unable to test this week but the change looks fairly similar to the original experiment otherwise

src/vcpkg/binarycaching.cpp
BillyONeal: Precheck is only used for `vcpkg ci` and our CI doesn't use these backends.

Crzyrndm: No issues then

Crzyrndm:
Testing results with 11 dependencies. Timing taken from the log line which looks like `Restored 11 package(s) from AWS in 34 s. Use --debug to see more details.` (34s)

Test run using `./vcpkg/vcpkg install --triplet=x64-windows --x-install-root=build/vcpkg_installed` and deleting the build folder each run. The cache was an AWS bucket set by env
* `$Env:VCPKG_BINARY_SOURCES="clear;x-aws,s3://<bucket>/test1,readwrite"`

## Baseline (Tag `2026.07.29`)

* Cache miss: 13s
  * 13s
  * 13s
* Cached: ~25s
  * 34s (possibly a cold path somewhere?)
  * 24s
  * 26s
  * 24s

## This branch

Build is windows, using the Win-x64-Release preset and vcpkg_root is checked out to the same tag (`2026.07.29`) and the executable is just copied over the previous test

* Cache miss: 
  * 2.7s
  * 2.3s
* Cached: 
  * 6.7s
  * 7.1s
  * 6.5s

## Summary

Compared to results in https://github.com/microsoft/vcpkg-tool/pull/1392#issuecomment-2111370045
* the baseline appears to be slightly slower (13 instead of 9.5, 25 instead of 23) - partially due to 1 extra package, probably lots of different versions involved, etc.
* This PR is ~50% slower than the original on my machine - likely due to different parallelism restriction (8 here, `get_concurrency` there). This is primarily for CI so a more limited value is sensible and is somewhat likely to reverse the change due to low core counts on CI machines

Crzyrndm (approved):

BillyONeal:
> This PR is ~50% slower than the original on my machine - likely due to different parallelism restriction (8 here, get_concurrency there). This is primarily for CI so a more limited value is sensible and is somewhat likely to reverse the change due to low core counts on CI machines

Yeah I was really worried about introducing reliability problems by hammering remotes with like 100 concurrent queries on big build machines tripping DDOS prevention and that sort of thing.

Crzyrndm:
> > This PR is ~50% slower than the original on my machine - likely due to different parallelism restriction (8 here, get_concurrency there). This is primarily for CI so a more limited value is sensible and is somewhat likely to reverse the change due to low core counts on CI machines
> 
> Yeah I was really worried about introducing reliability problems by hammering remotes with like 100 concurrent queries on big build machines tripping DDOS prevention and that sort of thing.

I did see one example where there was a warning printed about rate limiting but it wasn't replicated and still succeeded. 8 seems like a safe-ish default to start with and the worst potential outcome currently is I believe a local build so not too worried about it for now.
* May have been the first test run with parallel requests so potentially some kind of S3 scaling behaviour got triggered? If so, it would have been a cache miss test

vicroms (approved):

PR #2048, opened 2026-06-18
Fix Qt plugins deployment with z-applocal

vicroms:
Fixes: https://github.com/microsoft/vcpkg/issues/52458

PR #2092, opened 2026-07-31
z-applocal: Add Qt 6 plugin deployment support

miyanyan:
I noticed this pr https://github.com/microsoft/vcpkg-tool/pull/2048, but currently z-applocal doesn't handle qt6 plugins, so I let gpt 5.6 sol add this `deployQt6` function, and the azure-pipelines e2e tests are also generated by gpt 5.6 sol.

## Qt 5 / Qt 6 vcpkg Layout Comparison
|Component	| Qt 5	|Qt 6|
| -- | -- |--|
|Release Qt DLLs	|vcpkg_installed/x64-windows/bin/Qt5*.dll	|vcpkg_installed/x64-windows/bin/Qt6*.dll|
|Debug Qt DLLs	|vcpkg_installed/x64-windows/debug/bin/Qt5*d.dll	|vcpkg_installed/x64-windows/debug/bin/Qt6*d.dll|
|Release plugins	|vcpkg_installed/x64-windows/plugins/<type>/*.dll	|vcpkg_installed/x64-windows/Qt6/plugins/<type>/*.dll|
|Debug plugins	|vcpkg_installed/x64-windows/debug/plugins/<type>/*d.dll	|vcpkg_installed/x64-windows/debug/Qt6/plugins/<type>/*d.dll|
|Release QML imports	|vcpkg_installed/x64-windows/qml/	|vcpkg_installed/x64-windows/Qt6/qml/|
|Debug QML imports	|vcpkg_installed/x64-windows/debug/qml/	|vcpkg_installed/x64-windows/debug/Qt6/qml/|
|Module metadata	|Primarily qmake .pri files; deployment uses hardcoded mappings	|vcpkg_installed/x64-windows/share/Qt6/modules/<Module>.json|
|vcpkg deployment marker	| plugins/qtdeploy.ps1, with another copy under debug/plugins	| No equivalent qtdeploy.ps1|
|Installation detection	|Presence of qtdeploy.ps1	|Presence of Qt6/plugins or debug/Qt6/plugins|
|Module-to-plugin mapping	|Hardcoded in the existing Qt 5 deployment logic	| Read from the plugin_types field in the module JSON|

## qt6 metadata in `vcpkg_installed/x64-windows/share/Qt6/modules/<Module>.json`
the key `plugin_types` indicates related plugins
```json
{
    "schema_version": 3,
    "name": "Gui",
    "repository": "qtbase",
    "version": "6.11.1",
    "plugin_types": ["accessiblebridge", "platforms", "platforms/darwin", "xcbglintegrations", "platformthemes", "platforminputcontexts", "generic", "iconengines", "imageformats", "egldeviceintegrations", "wayland-graphics-integration-client", "wayland-decoration-client", "wayland-shell-integration"],
    "qpa": {
        "platforms": ["windows"],
        "default_platform": "windows"
    },
    "platforms": [
        {
            "name": "Windows",
            "version": "10.0.26200",
            "compiler_id": "MSVC",
            "compiler_version": "19.44.35228.0",
            "targets": [
                {
                    "architecture": "x86_64",
                    "abi": "x86_64-little_endian-lp64"
                }
            ]
        }
    ]
}
```
ref: qt6 windeployqt source https://github.com/qt/qtbase/blob/163e23edd42d47a5bf0c1b9b63928e66da9b374e/src/tools/windeployqt/qtmoduleinfo.cpp#L102

### DIFF
--- a/azure-pipelines/e2e-projects/applocal/qt/build.bat
+++ b/azure-pipelines/e2e-projects/applocal/qt/build.bat
@@ -1,0 +1,81 @@
+@echo off
+setlocal
+cd /d %~dp0
+
+call :build_qt 5 "" qt5-release installed\bin installed\plugins
+if errorlevel 1 exit /b 1
+call :build_qt 5 d qt5-debug installed\debug\bin installed\debug\plugins
+if errorlevel 1 exit /b 1
+call :build_qt 6 "" qt6-release installed\bin installed\Qt6\plugins
+if errorlevel 1 exit /b 1
+call :build_qt 6 d qt6-debug installed\debug\bin installed\debug\Qt6\plugins
+if errorlevel 1 exit /b 1
+exit /b 0
+
+:build_qt
+set "qt_major=%~1"
+set "debug_suffix=%~2"
+set "app_dir=%~3"
+set "bin_dir=%~4"
+set "plugins_dir=%~5"
+
+if not exist "%app_dir%" mkdir "%app_dir%"
+if not exist "%bin_dir%" mkdir "%bin_dir%"
+if not exist "%plugins_dir%\generic" mkdir "%plugins_dir%\generic"
+if not exist "%plugins_dir%\imageformats" mkdir "%plugins_dir%\imageformats"
+if not exist "%plugins_dir%\platforms" mkdir "%plugins_dir%\platforms"
+if not exist "%plugins_dir%\platformthemes" mkdir "%plugins_dir%\platformthemes"
+if not exist "%plugins_dir%\printsupport" mkdir "%plugins_dir%\printsupport"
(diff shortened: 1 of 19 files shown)


BillyONeal:
I think not handling this was somewhat intentional because we didn't want to expand the scope of the hardcoded stuff without bound but we never got to designing the feature that was supposed to fix it for good where a port could declare with a JSON document or similar "a dependency on a.dll implies a dependency on plugins/b.dll" or something like that.

We also wanted to have similar handling for the "tools" problem today where qt can get order dependent because different qt packages try to copy dependent DLLs around; also affects Python and friends. It implies needing to not have single ownership for those anymore.

(To be clear, I am not saying that we should not take this, just recording the thought here while I noticed)

Issue #2066, opened 2026-07-13
Configuring two CMake presets using Visual Studio's vcpkg concurrently fails


Cyphall:
When configuring two CMake presets concurrently via CLion, with each CMake preset invoking Visual Studio's vcpkg in manifest mode, one of them randomly fails.

Here is the logs of the failing one:
```
-- Running vcpkg install
Fetching registry information from https://github.com/microsoft/vcpkg (HEAD)...
error: "c:\program files\git\cmd\git.exe" "--git-dir=C:\Users\Cyphall\AppData\Local\vcpkg\registries\git\.git" "--work-tree=C:\Users\Cyphall\AppData\Local\vcpkg\registries\git" -c core.autocrlf=false init failed with exit code 128
fatal: cannot copy 'C:/Program Files/Git/mingw64/share/git-core/templates/hooks/commit-msg.sample' to 'C:/Users/Cyphall/AppData/Local/vcpkg/registries/git/.git/hooks/commit-msg.sample': File exists

while loading baseline version for boost-optional
error: "c:\program files\git\cmd\git.exe" "--git-dir=C:\Users\Cyphall\AppData\Local\vcpkg\registries\git\.git" "--work-tree=C:\Users\Cyphall\AppData\Local\vcpkg\registries\git" -c core.autocrlf=false init failed with exit code 128
fatal: cannot copy 'C:/Program Files/Git/mingw64/share/git-core/templates/hooks/commit-msg.sample' to 'C:/Users/Cyphall/AppData/Local/vcpkg/registries/git/.git/hooks/commit-msg.sample': File exists

while loading baseline version for boost-scope
error: "c:\program files\git\cmd\git.exe" "--git-dir=C:\Users\Cyphall\AppData\Local\vcpkg\registries\git\.git" "--work-tree=C:\Users\Cyphall\AppData\Local\vcpkg\registries\git" -c core.autocrlf=false init failed with exit code 128
fatal: cannot copy 'C:/Program Files/Git/mingw64/share/git-core/templates/hooks/commit-msg.sample' to 'C:/Users/Cyphall/AppData/Local/vcpkg/registries/git/.git/hooks/commit-msg.sample': File exists

while loading baseline version for glfw3
[...]

PR #2095, opened 2026-08-02
Do not replace registry errors with "<port> does not exist"

### LINKED ISSUE #2094: Registry errors are reported as "<port> does not exist", hiding the real failure
### Summary

`VersionedPortfileProviderImpl::entry()` discards the real error returned by `RegistryImplementation::get_port_entry` and replaces it with `{package_name} does not exist`. Any registry-level failure — network, git, filesystem, lock contention — is reported to the user as "this port doesn't exist", which is both wrong and undiagnosable.

### Where

https://github.com/microsoft/vcpkg-tool/blob/60b5cfa92045b3abf92fc5ce81e0c8c8be3e73ac/src/vcpkg/portfileprovider.cpp#L427-L455

```cpp
if (auto reg = m_registry_set.registry_for_port(name))
{
    if (auto entry = reg->get_port_entry(name))
    {
        entry_it = m_entry_cache.emplace(name.to_string(), std::move(entry)).first;
    }
    else
    {
        entry_it = m_entry_cache
                       .emplace(name.to_string(),
                                msg::format(msgPortDoesNotExist, msg::package_name = name))   // <-- error dropped
                       .first;
    }
}
```

`get_port_entry` returns `ExpectedL<std::unique_ptr<RegistryEntry>>` and has three outcomes:

1. **error** — something went wrong talking to the registry
2. **value, null pointer** — the registry genuinely has no such port
3. **value, non-null** — found

Case 2 is already handled further down, in `load_control_file`:

```cpp
if (!ent->get())
{
    return msg::format_error(msgPortDoesNotExist, msg::package_name = version_spec.port_name);
}
```

So the `else` branch above is reached **only** in case 1, and it throws the diagnostic away. It also uses `msg::format` rather than `msg::format_error`, so the resulting line isn't even prefixed with `error:`.

### Why it matters

This masked #2066 completely for us. A concurrent `git init` race in the registries cache (see #2066 / #2093) made `get_port_entry` fail, and all we ever saw was:

```
Fetching registry information from https://github.com/microsoft/vcpkg (HEAD)...
apr does not exist
apr-util does not exist
capstone does not exist
crc32c does not exist
expat does not exist
...
```

No `error:` prefix, no git output, no exit code, nothing. Every port from the default registry is listed as non-existent, which points suspicion at the manifest or the baseline rather than at the registry fetch that actually failed. It took a `GIT_TRACE2_EVENT` capture of a live failure to find the real message, which was:

```
could not lock config file .../registries/git/.git/config: File exists
could not set 'core.repositoryformatversion' to '0'
```

Note that #2066 hit the *same* underlying failure but through `get_baseline_version`, which propagates the error properly — so that reporter saw the actual git error and could describe the problem accurately. The difference is entirely down to this one `else` branch.

### Suggested fix

Propagate the error instead of inventing one:

```cpp
else
{
    entry_it = m_entry_cache.emplace(name.to_string(), std::move(entry).error()).first;
}
```

`ExpectedT::error() &&` already exists, and the cache value type (`ExpectedL<std::unique_ptr<RegistryEntry>>`) is constructible from `LocalizedString`, so this is a drop-in replacement. Case 2 keeps producing `msgPortDoesNotExist` via `load_control_file` as before.

I've opened a PR with this change.

bugale:
Fixes #2094.

`VersionedPortfileProviderImpl::entry()` threw away the error returned by `get_port_entry` and substituted `{package_name} does not exist`, so any registry-level failure was reported as the port not existing.

`get_port_entry` returns `ExpectedL<std::unique_ptr<RegistryEntry>>` with three outcomes:

1. **error** — something went wrong talking to the registry
2. **value, null pointer** — the registry genuinely has no such port
3. **value, non-null** — found

Case 2 is already handled downstream in `load_control_file`:

```cpp
if (!ent->get())
{
    return msg::format_error(msgPortDoesNotExist, msg::package_name = version_spec.port_name);
}
```

so the `else` branch touched here is reached **only** for case 1. It also used `msg::format` rather than `msg::format_error`, so the line wasn't even prefixed with `error:`.

## Before / after

This hid #2066 from us entirely. A concurrent `git init` race in the registries cache made `get_port_entry` fail, and the entire user-visible output was:

```
Fetching registry information from https://github.com/microsoft/vcpkg (HEAD)...
apr does not exist
apr-util does not exist
capstone does not exist
crc32c does not exist
expat does not exist
...
```

No `error:` prefix, no git output — just every port from the default registry declared non-existent, which points suspicion at the manifest or the baseline instead of at the registry fetch that actually failed. Finding the real message required a `GIT_TRACE2_EVENT` capture of a live failure:

```
could not lock config file .../registries/git/.git/config: File exists
could not set 'core.repositoryformatversion' to '0'
```

With this change that message reaches the user directly.

Worth noting: #2066 hit the same underlying failure but reached it through `get_baseline_version`, which propagates errors correctly — so that reporter saw the actual git error. The difference was entirely this one `else` branch.

## The change

```cpp
entry_it = m_entry_cache.emplace(name.to_string(), std::move(entry).error()).first;
```

`ExpectedT::error() &&` already exists and the cache value type is constructible from `LocalizedString`, so this is a drop-in replacement for the discarded `msg::format(...)`. `msgPortDoesNotExist` remains in use for the genuine cases at both other call sites in this file.

### DIFF
--- a/src/vcpkg/portfileprovider.cpp
+++ b/src/vcpkg/portfileprovider.cpp
@@ -436,17 +436,7 @@ namespace vcpkg
                 {
                     if (auto reg = m_registry_set.registry_for_port(name))
                     {
-                        if (auto entry = reg->get_port_entry(name))
-                        {
-                            entry_it = m_entry_cache.emplace(name.to_string(), std::move(entry)).first;
-                        }
-                        else
-                        {
-                            entry_it = m_entry_cache
-                                           .emplace(name.to_string(),
-                                                    msg::format(msgPortDoesNotExist, msg::package_name = name))
-                                           .first;
-                        }
+                        entry_it = m_entry_cache.emplace(name.to_string(), reg->get_port_entry(name)).first;
                     }
                     else
                     {


src/vcpkg/portfileprovider.cpp
```diff
@@ -442,10 +442,7 @@ namespace vcpkg
                         }
                         else
                         {
-                            entry_it = m_entry_cache
-                                           .emplace(name.to_string(),
-                                                    msg::format(msgPortDoesNotExist, msg::package_name = name))
-                                           .first;
+                            entry_it = m_entry_cache.emplace(name.to_string(), std::move(entry).error()).first;
                         }
```
BillyONeal: Is this whole thing now just:

```c++
entry_it = m_entry_cache.emplace(name.to_string(), reg->get_port_entry(name)).first;
```

I believe what this was *intending* to do was preserve the error cases and map the "nullptr unique_ptr" case to the "port does not exist" error message. So I think it's good to not drop the "child" error but I think this is incorrectly dropping a legitimate error too.

BillyONeal (dismissed):

src/vcpkg/portfileprovider.cpp
BillyONeal: I missed that you already discussed this in the description but the "is this whole thing now just" comment remains

bugale: @BillyONeal 
Yeah, you're right. Once the error is kept, both branches store the same thing, so the if/else does nothing. Pushed your version.
The nullptr case still gets the "does not exist" message, just one level up: `entry()` is only called from `load_control_file`, and that checks `!ent->get()` right after unwrapping. So a genuinely missing port reads the same as before, and a git or network failure now says what actually broke.

BillyONeal (approved):
Thanks!

BillyONeal:
(Oh, and sorry for the delay)

PR #2097, opened 2026-08-02
Use rename_or_delete when finalizing a downloaded .part file

### LINKED ISSUE #2096: Concurrent downloads race finalizing the .part file: rename(...): Access is denied
### Describe the bug

Concurrent vcpkg invocations that need the same asset fail with

```
error: rename(C:\Users\me\AppData\Local\vcpkg\downloads\PowerShell-7.6.1-win-x64.zip.5020.part,
              C:\Users\me\AppData\Local\vcpkg\downloads\PowerShell-7.6.1-win-x64.zip):
Access is denied.
```

followed by

```
A suitable version of powershell-core was not found (required v7.5.4).
```

The download itself succeeded and the hash matched; only the final rename into place failed.

### Cause

`download_file_asset_cached` downloads to a PID-suffixed part file and then renames it over the final path:

https://github.com/microsoft/vcpkg-tool/blob/60b5cfa92045b3abf92fc5ce81e0c8c8be3e73ac/src/vcpkg/base/downloads.cpp#L593-L600

https://github.com/microsoft/vcpkg-tool/blob/60b5cfa92045b3abf92fc5ce81e0c8c8be3e73ac/src/vcpkg/base/downloads.cpp#L642-L648

The PID suffix keeps the part files distinct, but the destination is shared and nothing serializes the rename. `Filesystem::rename` is `MoveFileExW(..., MOVEFILE_REPLACE_EXISTING)`, so a merely-existing destination is fine — the failure is `ERROR_ACCESS_DENIED`, which is what Windows returns when the destination is **open in another process**. That is precisely the interleaving here: process A renames its copy into place and immediately starts reading it (hashing it, or extracting it as a tool), and while that handle is open process B tries to replace the same file.

There is no lock over the downloads directory; the only file locks in the tree are the registries `.vcpkg-lock`, the installed database, and applocal.

### Why this is a plain oversight

The correct helper already exists and is used for exactly this pattern a few frames up the stack. `download_tool` extracts to a temp directory and then does a compare-and-swap:

https://github.com/microsoft/vcpkg-tool/blob/60b5cfa92045b3abf92fc5ce81e0c8c8be3e73ac/src/vcpkg/tools.cpp#L1019-L1027

`Filesystem::rename_or_delete` treats "the destination already exists" as "another process won the CAS", removes the loser's copy and reports it, and retries transient failures with backoff:

https://github.com/microsoft/vcpkg-tool/blob/60b5cfa92045b3abf92fc5ce81e0c8c8be3e73ac/src/vcpkg/base/files.cpp#L2206-L2236

So in `download_tool` the *extraction* is concurrency-safe and the *download* is not, even though the download is the easier of the two to make safe: the file is content-addressed, its SHA512 is verified on the line immediately before the rename, so a destination that already exists is provably bit-identical and losing the race is a no-op.

### Reproduction

Any two vcpkg processes started close together against a cold downloads cache. Concretely, six `vcpkg install` invocations for different triplets, launched together, against an empty `VCPKG_DOWNLOADS`:

```
iter 1 : 5/6 failed
    x86-windows-...   : rename("...\cmake-4.3.2-windows-x86_64.zip.69852.part", "...\cmake-4.3.2-windows-x86_64.zip"): Access is denied.
    arm64-windows-... : rename("...\cmake-4.3.2-windows-x86_64.zip.10416.part", "...\cmake-4.3.2-windows-x86_64.zip"): Access is denied.
    ...
iter 2 : 5/6 failed
iter 3 : 5/6 failed
TOTAL: 15 / 18 failed
```

It reproduces reliably because all six processes reach the same tool at the same moment.

This is not exotic: it is what happens whenever a build system fans out per-triplet or per-configuration installs, and it is the dominant CI failure mode for us. The reason it is not more widely reported is probably that a warm downloads cache hides it — `download_tool` hashes the existing archive first and skips the download entirely, so the race only exists on the first fetch:

https://github.com/microsoft/vcpkg-tool/blob/60b5cfa92045b3abf92fc5ce81e0c8c8be3e73ac/src/vcpkg/tools.cpp#L978-L1008

One tool makes it much worse than "first fetch only". The Visual Studio 2026 copy of `scripts/vcpkg-tools.json` lists powershell-core as `version: 7.5.4` while pointing `url` at the `v7.6.1` archive, and the extracted `pwsh.exe` reports 7.6.1. Under `--x-abi-tools-use-exact-versions` that version is then never accepted, so powershell-core is re-acquired on *every* invocation rather than once. (The data mismatch itself is a Visual Studio packaging problem — `microsoft/vcpkg` HEAD is self-consistent at 7.6.3 — but it means the rename above is hit repeatedly rather than only on a cold cache.)

### Suggested fix

Use `rename_or_delete` at the three sites in `downloads.cpp` that finalize a part file. PR to follow.

### Environment

- vcpkg-tool `2026-05-27-d5b6777d666efc1a7f491babfcdab37794c1ae3e` (bundled with Visual Studio 2026 18.7.4)
- Windows 11 / Windows Server 2022, x64
- Also present verbatim on `main` at `60b5cfa9`

bugale:
Fixes #2096.

Concurrent vcpkg processes download an asset to `<dest>.<pid>.part` and then rename it onto the shared `<dest>`. The part files are distinct, but the rename is unserialized, and on Windows it fails with `ERROR_ACCESS_DENIED` while another process holds the freshly-renamed destination open — which is exactly what that process does next, hashing it or extracting it as a tool.

`rename_or_delete` is the compare-and-swap that already handles this pattern for the extracted tool directory, a few frames up in `download_tool`:

```cpp
auto maybe_partial_path = extract_archive_to_temp_subdirectory(context, fs, *this, download_path, tool_dir_path);
if (auto partial_path = maybe_partial_path.get())
{
    if (!fs.rename_or_delete(context, *partial_path, tool_dir_path))
```

It is the right primitive for the download too, and in fact more obviously so: the destination is content-addressed and `check_downloaded_file_hash` runs on the line immediately before, so a destination that already exists is bit-identical and discarding the loser's copy is a no-op. It also retries transient failures with backoff, which the bare `rename` did not.

The `DiagnosticContext` overload returns `Optional<bool>`, whose `operator bool` is `has_value`, so `!result` distinguishes a genuine failure from "lost the CAS, which is fine" — the same discrimination `download_tool` relies on.

Three call sites finalize a part file and all three are changed: the direct download, and the two asset-cache-script paths (with and without an expected hash). The third has no hash to verify, so which copy survives is arbitrary either way — it was `MOVEFILE_REPLACE_EXISTING` (last writer wins) and is now first writer wins. Happy to drop that one if you would rather keep the change to the hash-verified paths.

Verified against the reproducer in the issue: six concurrent `vcpkg install` invocations for different triplets against a cold `VCPKG_DOWNLOADS` go from 15/18 failures to passing, once the destination is no longer contended.

### DIFF
--- a/src/vcpkg/base/downloads.cpp
+++ b/src/vcpkg/base/downloads.cpp
@@ -644,7 +644,11 @@ namespace vcpkg
             return DownloadPrognosis::OtherError;
         }
 
-        fs.rename(download_path_part_path, download_path, VCPKG_LINE_INFO);
+        if (!fs.rename_or_delete(context, download_path_part_path, download_path).has_value())
+        {
+            return DownloadPrognosis::OtherError;
+        }
+
         return DownloadPrognosis::Success;
     }
 
@@ -831,7 +835,11 @@ namespace vcpkg
                             *out_sha512 = std::move(hash_result.hash);
                         }
 
-                        fs.rename(download_path_part_path, download_path, VCPKG_LINE_INFO);
+                        if (!fs.rename_or_delete(context, download_path_part_path, download_path).has_value())
+                        {
+                            return DownloadPrognosis::OtherError;
+                        }
+
                         return DownloadPrognosis::Success;
                     }
 
@@ -865,7 +873,11 @@ namespace vcpkg
 
         if (fs.exists(download_path_part_path, VCPKG_LINE_INFO))
         {
-            fs.rename(download_path_part_path, download_path, VCPKG_LINE_INFO);
+            if (!fs.rename_or_delete(context, download_path_part_path, download_path).has_value())
+            {
+                return DownloadPrognosis::OtherError;
+            }
+
             return DownloadPrognosis::Success;
         }
 


src/vcpkg/base/downloads.cpp
```diff
@@ -644,7 +644,11 @@ namespace vcpkg
             return DownloadPrognosis::OtherError;
         }
 
-        fs.rename(download_path_part_path, download_path, VCPKG_LINE_INFO);
+        if (!fs.rename_or_delete(context, download_path_part_path, download_path))
+        {
+            return DownloadPrognosis::OtherError;
+        }
```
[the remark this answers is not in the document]
BillyONeal: I think this is in fact the intended outcome. If two processes race when there is no known SHA we are already doomed.

src/vcpkg/base/downloads.cpp
```diff
@@ -865,7 +873,11 @@ namespace vcpkg
 
         if (fs.exists(download_path_part_path, VCPKG_LINE_INFO))
```
BillyONeal, line 874: This exists check is a bit suspicious given what you're already fixing here.

src/vcpkg/base/downloads.cpp
```diff
@@ -865,7 +873,11 @@ namespace vcpkg
...
         {
-            fs.rename(download_path_part_path, download_path, VCPKG_LINE_INFO);
+            if (!fs.rename_or_delete(context, download_path_part_path, download_path))
```
BillyONeal: `rename_or_delete` "Returns whether the rename actually happened." so I'm not sure the target already existing should be an error as this does?

src/vcpkg/base/downloads.cpp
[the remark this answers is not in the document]
bugale: That was my thinking too. On the two hashed paths the destination is content-addressed and we check the hash on the line right above, so losing the race is harmless. With no SHA there's nothing to compare anyway, and the old code just killed the process, so this felt like the better choice.

src/vcpkg/base/downloads.cpp
bugale: It isn't treated as an error, but I can see why it might be confusing, so I've changed it.
The `DiagnosticContext` overload returns `Optional<bool>`, not `bool`. `Optional`'s `operator bool` is `has_value()`, so the `!` there means "the call failed", not "the rename didn't happen". Losing the CAS gives you a filled `Optional` holding `false`, which passes the check and falls through to `Success`.
That's way too subtle when the inner type is also a bool, so all three places are now `.has_value()`.

BTW about the doc comment you quoted, "if old_path and new_path are files, this function always returns true" isn't true once processes race, which is what #2096 is. The rename fails with `ERROR_ACCESS_DENIED`, not `file_exists`, so it doesn't match the error code checks. It gets caught by the `this->exists(new_path, local_ec)` at the end of that same condition, which deletes our `.part` and returns false.

src/vcpkg/base/downloads.cpp
bugale: I looked into it and I think it's fine, though not for an obvious reason.
It's not a cross-process race. The part path here is `fmt::format("{}.{}.part", ..., get_process_id())` a few lines up, so only this process can create or delete it.
It's also doing real work. `rename_or_delete` doesn't fail fast when the source is missing, it walks the retry loop sleeping 10ms, 100ms, 1s and 10s before giving up. So dropping the check gets you one of two bad outcomes: an ~11 second stall ending in a generic filesystem error instead of `msgAssetCacheScriptFailedToWriteFile`, or, if `download_path` happens to already exist, `rename_or_delete` reports a lost CAS and we return `Success` even though the script produced nothing.

The part I do think is off is the overload. `exists(..., VCPKG_LINE_INFO)` exits the process on a filesystem error, inside a function that reports everything else through `context`. There's no `DiagnosticContext` overload of `exists` though, only `ec` and `LineInfo`, so converting it means either swallowing the error or writing the report by hand.
Let me know if you want me to do it as part of this PR

BillyONeal: Oh, I see I got confused between the source and target.

BillyONeal (approved):
Thanks!

BillyONeal:
Sorry for the delay... took a bit of convincing of myself that this seemed reasonable.

PR #2112, opened 2026-09-02
Fix binary cache precheck result mapping

miyanyan:
Fixes `ReadOnlyBinaryCache::precheck()` assigning provider results to the wrong `CacheStatus` after the action list has been filtered.

`action_ptrs` and `cache_result` use indices local to the filtered list, while `statuses` uses indices from the original action list. The existing code wrote results to `statuses[i]`, ignoring the original indices stored in `indexes`.


BillyONeal (approved):
Yikes, thank you!

PR #2120, opened 2026-09-16
Quote POSIX redirection, tilde, and brace characters in append_shell_escaped

gmjasper:
The non-Windows quoting trigger list omitted `<`, `>`, `~`, `{`, and `}`, so an argument built from only those characters missed every trigger and was appended to the command line verbatim. On POSIX that line is handed to `/bin/sh -c`, so the shell acted on data that `Command::string_arg` is supposed to pass through literally: `a>b` redirected and created `b`, `a<b` consumed a redirection operand, `~` expanded to the home directory, and `{a,b}` brace expanded under a bash `/bin/sh`.

Fix this the same way 2181a1fe did for `*?[#`, by quoting the whole argument. The existing double-quoted encoding already escapes `\`, `"`, `` ` ``, and `$`, and inside POSIX double quotes these characters lose their special meaning, so no new per-character escaping is needed. Windows is unaffected: it builds its command line for CreateProcess rather than a shell.

Tested: full `vcpkg-test` suite passes on Linux (GCC 13.3) and Windows (MSVC 19.51); the extended `append_shell_escaped` case fails without the source change. I also have a runtime test that spawns child processes and asserts exact `argv` plus filesystem side effects rather than comparing formatted command strings.


gmjasper:
@microsoft-github-policy-service agree company="Microsoft"

BillyONeal (approved):

PR #2121, opened 2026-09-16
Add `apple` as a valid platform expression

cqundefine:
On the `vcpkg` side we already have `VCPKG_TARGET_IS_APPLE`, add a respective platform expression identifier.

The main reason I want that is I found at least some packages with a very wide `supports` clause which instead of listing the supported platforms lists the unsupported platforms like `!a & !b & ...`. There are a bunch of cases that support Windows, Linux (and often by extension Android) and macOS (which in a lot of cases by extension means all Apple platforms) so instead of listing `windows & linux & android & osx & ios & tvos & watchos & visionos` being able to write `windows & linux & android & apple` looks much nicer.


BillyONeal (approved):
Thanks! I also asked GPT 5.6 Sol if there were any other missing ones and it said no.

BillyONeal:
Also added https://github.com/MicrosoftDocs/vcpkg-docs/pull/612

PR #2124, opened 2026-09-17
Match xbox expressions to CMake definitions

BillyONeal:
While documenting the Apple platform expression added for #2121 in MicrosoftDocs/vcpkg-docs#612, the platform-expression reference exposed that the existing xbox identifier did not match `VCPKG_TARGET_IS_XBOX`.

The CMake definition recognizes Xbox only for the desktop Windows system name (empty or undefined) with a non-empty `VCPKG_XBOX_CONSOLE_TARGET`. Platform expressions previously checked only the console target, so UWP and unrelated target systems could incorrectly satisfy xbox.

Apply both conditions and cover missing, empty, desktop Windows, UWP, and non-Windows system-name cases.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


BillyONeal:
Note: https://github.com/microsoft/vcpkg/blob/a97c98d785fc0b1710bc711acf47767b756e6b7c/scripts/cmake/vcpkg_common_definitions.cmake#L3-L8

vicroms (approved):

PR #2125, opened 2026-09-17
Make binary cache ZIP compression level configurable

### LINKED ISSUE #53923: Make the binary-cache compression level configurable
## Description

vcpkg should provide a unified setting for controlling the compression level of binary-cache archives.

The setting should apply consistently on every platform:

Windows currently uses 7-Zip.
Linux and macOS currently use the zip command.
Any future archive backend should also respect the same setting.

Currently, vcpkg does not explicitly specify a compression level when creating binary-cache ZIP files. Consequently, the compression behavior depends on the default of the platform-specific archive tool.

On Windows, vcpkg currently invokes 7-Zip approximately as follows:

```
7z a <destination.zip> <package-directory>/*
```

Because no -mx option is specified, 7-Zip uses its default compression level, -mx=5.

On non-Windows platforms, vcpkg invokes zip approximately as follows:

```
zip --quiet -y -r <destination.zip> '*'
```

This also relies on the tool's default compression level.

For large packages, compression can dominate the entire binary-cache submission time. This is particularly inefficient when the binary cache is hosted on a fast LAN, where transferring a slightly larger archive is much cheaper than spending several minutes compressing it.

## Benchmark

Environment:

OS: Windows
CPU: AMD Ryzen 9 9950X3D v2
Package: dawn:x64-windows-static
Uncompressed package size: approximately 6.5 GB
vcpkg-tool version: 2026-07-27-98d7cb0cf1f4686a3e43aa5672b6230c1d56bce8

Results:

| Compression level |	Packaging time |	Resulting ZIP size |
| --- | --- | --- |
| 5 — current 7-Zip default |	3.5 minutes |	1.05 GB |
| 1 — fastest compression  |	20 seconds |	1.14 GB |
| 0 — no compression  |	6 seconds |	6.5 GB |

Compression level 1 is approximately 10.5 times faster, while increasing the resulting archive size by only approximately 90 MB (8.6%).

In this case, the default compression level saves relatively little storage and network traffic but adds approximately 190 seconds to the binary-cache submission.

## Proposed behavior

Add a platform-independent binary-cache compression-level setting, for example:

```
--binary-cache-compression-level=<0-9>
```

and/or:

```
VCPKG_BINARY_CACHE_COMPRESSION_LEVEL=<0-9>
```

The selected value should be passed to the platform-specific ZIP implementation.

For example:

## Windows

```
7z a -tzip -mx=<level> <destination.zip> <source>
```

## non-Windows

```
zip -<level> --quiet -y -r <destination.zip> '*'
```

Suggested semantics:

| Level |	Meaning |
| --- | --- |
| 0 |	Store without compression |
| 1 |	Fastest compression |
| 5 |	Normal compression |
| 9 |	Maximum compression |

The setting should affect every binary-cache provider that requires a ZIP archive

## Default value

If no compression level is configured, the current behavior is retained.

However, I think level 1 may be a more appropriate default for binary caching, and the benchmark above shows that substantially more CPU time can produce only a small reduction in archive size.

Regardless of the chosen default, users should be able to configure the trade-off without maintaining a patched version of vcpkg-tool.

## Motivation

The optimal compression level depends on the environment:

Fast LAN cache: level 1
Internet-hosted cache: a higher level may reduce transfer time
CPU-constrained CI worker: lower compression
Storage-constrained cache server: higher compression
Packages containing large .lib or .pdb files: compression level can significantly affect submission latency

Large libraries such as Dawn, LLVM, and Qt can produce multi-gigabyte package directories. For these packages, ZIP creation may take much longer than the actual upload.

A unified, cross-platform compression setting would make binary-cache performance predictable and allow users to choose the appropriate CPU, storage, and network trade-off for their environment.

xiaozhuai:
Adds `--binary-cache-compression-level=<0-9>` and `VCPKG_BINARY_CACHE_COMPRESSION_LEVEL` so users can trade archive size for binary-cache submission time. The command-line option takes precedence; leaving both unset preserves the archive tool's default compression level. Invalid values are rejected with a localized diagnostic.

The common ZIP creation path passes the configured level to 7-Zip (`-mx=<level>`) on Windows and `zip` (`-<level>`) elsewhere, covering every binary-cache provider that requests a ZIP archive. 7-Zip explicitly uses `-tzip` regardless of whether a level is configured. NuGet packaging is unaffected, and changing the setting does not change package ABIs or prevent restoration of existing archives.

Tests cover defaults, every configured level, argument validation and precedence, and one cache-only restoration using `--only-binarycaching` with a different compression setting and an exact payload comparison. The test harness isolates the new environment variable.

Also fixes the macOS CI failure in `ci.ps1`: PowerShell's asynchronous stdout/stderr merging interleaved regression diagnostics into the summary, breaking a contiguous-text assertion. That suite now captures each stream separately before combining them for assertions.

Documentation: https://github.com/MicrosoftDocs/vcpkg-docs/pull/613

Fixes https://github.com/microsoft/vcpkg/issues/53923

Validation after merging current upstream main:
- macOS Debug build; all 534 unit tests passed (8,005 assertions).
- Actual PowerShell execution of the binary-cache-compression and ci suites passed, using the repository prelude and a local runner with `/opt/vcpkg`.
- PowerShell message-usage validation passed; regenerated message map matches.
- `git diff --check` passed.
- Prior macOS integration checks covered archive integrity, restored payloads, and eight invalid CLI/environment values.

Windows and Linux execution for the latest revision is left to CI.


azure-pipelines/end-to-end-tests-dir/binary-cache-compression.ps1
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        Remove-Item -Recurse -Force $installRoot
+        Remove-Item -Recurse -Force $buildtreesRoot
+        # A different compression setting must still restore the same ABI from the cache.
+        $env:VCPKG_BINARY_CACHE_COMPRESSION_LEVEL = '9'
+        $output = Run-VcpkgAndCaptureOutput ($commonArgs + @('install', 'binary-cache-compression', "--binarysource=clear;files,$ArchiveRoot,read"))
+        Throw-IfFailed
+        Require-FileExists "$installRoot/$Triplet/share/binary-cache-compression/copyright"
+        Throw-IfNonContains -Actual $output -Expected "Restored 1 package(s)"
```
BillyONeal: Can you use `--only-binarycaching` rather than asserting on bits that aren't actually part of the contract here? (e.g. if we change the "Restored" message that is in contract but would break this test)

azure-pipelines/end-to-end-tests-dir/binary-cache-compression.ps1
```diff
@@ -0,0 +1,47 @@
...
+            } elseif ($null -ne $level -and $entry.CompressedLength -ge $entry.Length) {
+                throw 'Expected compression of repetitive package content'
+            }
+        } finally {
+            $archive.Dispose()
+        }
+
+        Remove-Item -Recurse -Force $installRoot
+        Remove-Item -Recurse -Force $buildtreesRoot
+        # A different compression setting must still restore the same ABI from the cache.
```
BillyONeal: I don't think this needs to be checked 10 times.

docs/binary-cache-compression.md
```diff
@@ -0,0 +1,25 @@
+# Binary cache compression
```
BillyONeal: This needs to go to the vcpkg-docs repo not here.

BillyONeal (changes requested):

BillyONeal:
Thanks for the new feature! All the product code changes look correct to me.

xiaozhuai:
https://github.com/MicrosoftDocs/vcpkg-docs/pull/613

MichalPetryka:
Would it make sense to migrate to some more modern compression format like brotli or zstd before exposing levels as a documented option?

BillyONeal:
> Would it make sense to migrate to some more modern compression format like brotli or zstd before exposing levels as a documented option?

That will be substantially more complicated because it has backcompat impact. Whereas this is just passing through an option so it's effectively "free".

azure-pipelines/end-to-end-tests-dir/ci.ps1
```diff
@@ -1,7 +1,22 @@
 . $PSScriptRoot/../end-to-end-tests-prelude.ps1
 
+# Keep each stream intact: regression diagnostics on stderr can otherwise split
+# the summary on stdout when PowerShell merges the two streams asynchronously.
+function Run-CiAndCaptureBoth {
```
BillyONeal: I was going to ask why you touched this, then I remembered about you getting random test failures here. I... don't like this solution but I also don't want to block your PR over it so :shrug:

I do think this suggests there may be a bug where we should be issuing flush calls.

azure-pipelines/e2e-ports/binary-cache-compression/portfile.cmake
```diff
@@ -0,0 +1,3 @@
+set(VCPKG_POLICY_EMPTY_INCLUDE_FOLDER enabled)
```
BillyONeal: Can we just reuse the `vcpkg-header-only` test port for this or do we need a separate new port?

azure-pipelines/end-to-end-tests-dir/ci.ps1
BillyONeal: If this is ready to land before https://github.com/microsoft/vcpkg-tool/pull/2127 I can just revert this part there.

xiaozhuai: > I remembered about you getting random test failures here.

Yes, this is the reason.
Since I'm not sure what happened, I only noticed that stdout and stderr were mixed together.
To minimize the scope of the changes, I've implemented this local workaround.
If this issue has already been fixed, I'd be happy to revert it.

BillyONeal (approved):
Thank you!

PR #2126, opened 2026-09-18
Add a host-info command

rkitover:
Diagnosing a vcpkg problem from a bug report means asking the reporter what their machine is and how vcpkg is configured on it, a question that takes several rounds to answer and is easy to answer incompletely. Print it in one place instead.

The output takes the shape of the depend-info list format: a name, a colon, and a comma separated list of values. It covers the operating system, the invoked executable, the vcpkg root, the host triplet, the CMake variables this tool resolves for the toolchain and for port builds, and the environment variables that configure vcpkg. Each is named as $ENV{VAR}, $TOOLCHAIN{VAR} or $PORT{VAR}, so it is clear which of the three a name belongs to.

VCPKG_HOST_TRIPLET is not among the toolchain variables. The toolchain only reads it and never assigns it, so it holds a value only when the user supplied one; the host triplet actually in effect is reported on its own line. X_VCPKG_RECURSIVE_DATA is likewise left out of the environment: vcpkg sets it for its own child processes to pass configuration down and to stop a second level of recursion, so it says nothing about how the user configured anything.

What is reported about the operating system differs by platform: the version, edition, release and build on Windows; the kernel and the C library on Linux; the product version and build on macOS. Elsewhere, the name alone.

Naming the C library takes two routes. glibc answers gnu_get_libc_version(). musl defines no version macro, on purpose, and offers no equivalent, so its dynamic loader is asked instead: run with no arguments it reports its own version, and its mere presence says the library is musl even if the version cannot be read back.

Two Windows details are worth calling out. ProductName was never updated for Windows 11 and still reads "Windows 10 ...", so the edition is corrected using the build number as Microsoft's own tooling does, except on Windows Server, which is versioned by release year and names itself correctly. The version itself comes from the version resource of kernel32.dll rather than GetVersionEx, which reports only as much as this process is manifested for.

Variables that vcpkg splits into several values -- the overlay paths, the feature flags, the passed through environment variables -- are printed as the list vcpkg acts on, in the order given. Variables that are not set are left out, so what appears is what is in effect.

Only values this tool resolves on its own are printed for the CMake variables. The toolchain file lives in the registry rather than here, so its own variables are not known, and the variables a port build receives are only meaningful once a port is being built. The platform toolset and the git path are omitted deliberately: resolving them selects a Visual Studio instance and can acquire git, which a command that only reports should not do.

The command is documented at the learn.microsoft.com URL the other public commands point at, which `vcpkg help host-info` prints. That page does not exist yet and would need writing alongside this.

Tested on Windows Dev Insider Build, macOS latest, Linux Ubuntu and Linux Alpine.

Assisted-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


rkitover:
Here is a screenshot:

svcpkg host-info screenshot

